PR #14, opened 2019-09-30
Load window size/position

DamianMadden:
Applies loaded window size/position to the window, removed limitation on dock loaded state because loading defaults is now on the login screen

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -199,6 +199,9 @@ void vcMain_LoadSettings(vcState *pProgramState, bool forceDefaults)
     case 1: ImGui::StyleColorsDark(); break;
     case 2: ImGui::StyleColorsLight(); break;
     }
+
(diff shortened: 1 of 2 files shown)


src/main.cpp
```diff
@@ -199,6 +199,9 @@ void vcMain_LoadSettings(vcState *pProgramState, bool forceDefaults)
     case 1: ImGui::StyleColorsDark(); break;
     case 2: ImGui::StyleColorsLight(); break;
     }
+
+    SDL_SetWindowSize(pProgramState->pWindow, pProgramState->settings.window.width, pProgramState->settings.window.height);
```
OsiJr: I can't remember the exact rules for this- there was problems doing it this way originally. From memory- you want it to happen at start-up but not on force reset?

DamianMadden: This resolves AB#97

samsinsane: There's a good chance that this has broken the web client, I don't recall the exact rules when it came to the `SDL_SetWindowSize` function but there's a situation that will cause it to not resize with the browser window.

PR #20, opened 2019-10-01
System for more detailed project change errors

DamianMadden:
Resolves [AB#123](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/123)


src/vcState.h
```diff
@@ -115,6 +115,8 @@ struct vcState
   vcLoginStatus loginStatus;
   vdkError logoutReason;
 
+  vdkError lastError;
```
OsiJr: This isn't a good system; having a single value for an error isn't enough- you should leverage the file error system or modify that to suit instead.

PR #21, opened 2019-10-01
Changed to isWindowHovered, deleted some excess code

DamianMadden:
Resolves [AB#112](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/112)

### DIFF
--- a/src/vcCamera.cpp
+++ b/src/vcCamera.cpp
@@ -600,9 +600,9 @@ void vcCamera_HandleSceneInput(vcState *pProgramState, udDouble3 oscMove, udFloa
     memset(isBtnClicked, 0, sizeof(isBtnClicked));
     memset(isBtnDoubleClicked, 0, sizeof(isBtnDoubleClicked));
     memset(isBtnHeld, 0, sizeof(isBtnHeld));
-    //memset(isBtnReleased, 0, sizeof(isBtnReleased));
     mouseDelta = ImVec2();
     mouseWheel = 0.0f;
+    // Leaving isBtnReleased unchanged as there should be no reason to ignore a mouse release while the window has mouse focus
   }
 
   // Controller Input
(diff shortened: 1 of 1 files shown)


OsiJr:
@CarlCFX could you please test this functionality?

CarlCFX:
This fixes the instance of the bug (commented on in DevOps). No other issues relating to this change have been observed.

CarlCFX:
Well spotted Pete. Also occurs if you try to release the Pan over the UI within the Scene window too.

src/vcCamera.cpp
```diff
@@ -600,7 +600,7 @@ void vcCamera_HandleSceneInput(vcState *pProgramState, udDouble3 oscMove, udFloa
     memset(isBtnClicked, 0, sizeof(isBtnClicked));
     memset(isBtnDoubleClicked, 0, sizeof(isBtnDoubleClicked));
     memset(isBtnHeld, 0, sizeof(isBtnHeld));
-    //memset(isBtnReleased, 0, sizeof(isBtnReleased)); // 09/04/19 - commented out to fix pan-release bug, check back and review once more fully tested
+    //memset(isBtnReleased, 0, sizeof(isBtnReleased));
```
samsinsane: Can you create an issue in Azure for this unless it already exists? I recommend referencing this code block rather than this specific line that's commented out as it may not exist when the next person comes along to fix the pan-release problem.

DamianMadden: Do we need an issue though? I mean this is pretty much how I'd go about fixing it regardless, when would we ever want to ignore a mouse button release event while the window has mouse focus?

samsinsane: If this is the correct solution, then replace it with a meaningful comment about how the `isBtnReleased` variable shouldn't be zeroed.

src/vcCamera.cpp
```diff
@@ -817,25 +817,6 @@ void vcCamera_HandleSceneInput(vcState *pProgramState, udDouble3 oscMove, udFloa
...
-  {
-    if (io.MouseDown[0] && !io.MouseDown[1] && !io.MouseDown[2])
-      pProgramState->cameraInput.currentPivotMode = pProgramState->settings.camera.cameraMouseBindings[0];
-
-    if (!io.MouseDown[0] && io.MouseDown[1] && !io.MouseDown[2])
-      pProgramState->cameraInput.currentPivotMode = pProgramState->settings.camera.cameraMouseBindings[1];
-
-    if (!io.MouseDown[0] && !io.MouseDown[1] && io.MouseDown[2])
-      pProgramState->cameraInput.currentPivotMode = pProgramState->settings.camera.cameraMouseBindings[2];
-  }
```
OsiJr: You should have taken the variable out of the header as well.

PR #23, opened 2019-10-02
Improvements to internal handling of texture formats

peterAdams1014:
Removed all assumptions about texture storage in texture functions
Error checking in vcFramebuffer and vcTexture classes is now more thorough and consistent
[AB#327](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/327)

### DIFF
--- a/src/gl/metal/vcTexture.mm
+++ b/src/gl/metal/vcTexture.mm
@@ -14,75 +14,99 @@
 uint32_t g_textureIndex = 0;
 uint32_t g_blitTargetID = 0;
 
+void vcTexture_GetFormatAndPixelSize(const vcTextureFormat format, int *pPixelSize = nullptr, MTLPixelFormat *pPixelFormat = nullptr, MTLStorageMode *pStorageMode = nullptr, MTLTextureUsage *pUsage = nullptr)
+{
+  // defaults
+  MTLPixelFormat pixelFormat = MTLPixelFormatRGBA8Unorm;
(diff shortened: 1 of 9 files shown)


src/gl/metal/vcTexture.mm
```diff
@@ -14,75 +14,99 @@
...
+  case vcTextureFormat_BGRA8:
+    pixelFormat = MTLPixelFormatBGRA8Unorm;
+    usage = MTLTextureUsageShaderRead;
+    pixelSize = 4;
+#if UDPLATFORM_IOS || UDPLATFORM_IOS_SIMULATOR
+    storageMode = MTLStorageModeShared;
+#elif UDPLATFORM_OSX
+    storageMode = MTLStorageModeManaged;
+#endif
+    break;
```
DamianMadden: Storage mode for these types cannot be reliably predicted from vcTextureFormat, contrary to something I said at one point the render targets are all private, and any cpu reads we do on them atm are actually blits, are we ever calling this function to check render targets?

peterAdams1014: I'm not sure what you mean here. This PR has simply re-organised code to remove duplication. No functionality has been added or removed. 

Are you suggesting the original code has problems? Or have I made a mistake in consolidating this functionality?

DamianMadden: Oh nvm, I was considering the function entirely out of context

PR #25, opened 2019-10-02
Translate menu window name

DamianMadden:
Probably the only way to do this, ImGui was filling a backup name for it (in english) since the window had no name, also no built in way to give it a name
Resolves [AB#2](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/2)

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2065,7 +2065,14 @@ void vcRenderWindow(vcState *pProgramState)
     int margin = vcMainMenuGui(pProgramState);
 
     if (pProgramState->settings.docksLoaded != vcSettings::vcDockLoaded::vcDL_True)
+    {
+      ImGuiWindow *pWindow = ImGui::FindWindowByName("##MainMenuBar");
+      delete pWindow->Name;
+      pWindow->Name = new char[80];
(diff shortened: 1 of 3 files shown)


DamianMadden:
> This is a very dirty fix and relies on ImGui not changing as well. ImGui has no function to rename a window?

Nope, unless it's very well hidden

src/main.cpp
```diff
@@ -2065,7 +2065,14 @@ void vcRenderWindow(vcState *pProgramState)
     int margin = vcMainMenuGui(pProgramState);
 
     if (pProgramState->settings.docksLoaded != vcSettings::vcDockLoaded::vcDL_True)
+    {
+      ImGuiWindow *pWindow = ImGui::FindWindowByName("##MainMenuBar");
+      delete pWindow->Name;
+      pWindow->Name = new char[80];
+      vcStringFormat(pWindow->Name, 80, "{0}##MainMenuBar", vcString::Get("menuName"));
```
OsiJr: This won't be triggered when you change languages?

samsinsane:
We're doing development for Client in GitHub now, so we could just bug this in the ImGui repo, tag it in the Azure task and move on? It seems like an incredibly niche situation to invest effort in, it's also not the only string in ImGui that isn't translated.

PR #27, opened 2019-10-02
Added more error checking on vcGL operations

peterAdams1014:
Minor modifications with some functions to be more consistent with error checks.
[AB#341](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/341)

### DIFF
--- a/src/gl/directx11/vcShader.cpp
+++ b/src/gl/directx11/vcShader.cpp
@@ -196,14 +196,14 @@ void vcShader_DestroyShader(vcShader **ppShader)
 
 bool vcShader_Bind(vcShader *pShader)
 {
-  if (pShader == nullptr)
-    return false;
+  if (pShader != nullptr)
+  {
+    g_pd3dDeviceContext->IASetInputLayout(pShader->pLayout);
 
-  g_pd3dDeviceContext->IASetInputLayout(pShader->pLayout);
-
-  g_pd3dDeviceContext->VSSetShader(pShader->pVertexShader, NULL, 0);
-  g_pd3dDeviceContext->PSSetShader(pShader->pPixelShader, NULL, 0);
-  g_pd3dDeviceContext->GSSetShader(pShader->pGeometryShader, NULL, 0);
+    g_pd3dDeviceContext->VSSetShader(pShader->pVertexShader, NULL, 0);
+    g_pd3dDeviceContext->PSSetShader(pShader->pPixelShader, NULL, 0);
+    g_pd3dDeviceContext->GSSetShader(pShader->pGeometryShader, NULL, 0);
+  }
 
   return true;
 }
--- a/src/rendering/vcCompass.cpp
+++ b/src/rendering/vcCompass.cpp
@@ -59,7 +59,7 @@ udResult vcCompass_Destroy(vcAnchor **ppCompass)
   return udR_Success;
 }
 
-bool vcCompass_Render(vcAnchor *pCompass, vcAnchorStyle anchorStyle, const udDouble4x4 &worldViewProj, const udDouble4 &colour /*= udDouble4::create(1.0, 1.0, 1.0, 1.0)*/)
+bool vcCompass_Render(vcAnchor *pCompass, vcAnchorStyle anchorStyle, const udDouble4x4 &worldViewProj, const udDouble4 &colour /*= udDouble4::create(1.0, 0.8431, 0.0, 1.0)*/)
 {
   if (pCompass == nullptr || anchorStyle >= vcAS_Count || vcASToMeshType[anchorStyle] == vcInternalMeshType_Count)
     return false;
--- a/src/rendering/vcCompass.cpp
+++ b/src/rendering/vcCompass.cpp
@@ -59,23 +59,18 @@ udResult vcCompass_Destroy(vcAnchor **ppCompass)
   return udR_Success;
 }
 
-udResult vcCompass_Render(vcAnchor *pCompass, vcAnchorStyle anchorStyle, const udDouble4x4 &worldViewProj, const udDouble4 &colour /*= udDouble4::create(1.0, 1.0, 1.0, 1.0)*/)
+bool vcCompass_Render(vcAnchor *pCompass, vcAnchorStyle anchorStyle, const udDouble4x4 &worldViewProj, const udDouble4 &colour /*= udDouble4::create(1.0, 1.0, 1.0, 1.0)*/)
 {
   if (pCompass == nullptr || anchorStyle >= vcAS_Count || vcASToMeshType[anchorStyle] == vcInternalMeshType_Count)
-    return udR_InvalidParameter_;
-
-  udResult result = udR_Failure_;
+    return false;
 
   pCompass->shaderBuffer.u_colour = udFloat4::create(colour);
   pCompass->shaderBuffer.u_viewProjectionMatrix = udFloat4x4::create(worldViewProj);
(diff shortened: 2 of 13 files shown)


src/gl/directx11/vcShader.cpp
```diff
@@ -196,14 +196,14 @@ void vcShader_DestroyShader(vcShader **ppShader)
...
 
-  g_pd3dDeviceContext->VSSetShader(pShader->pVertexShader, NULL, 0);
-  g_pd3dDeviceContext->PSSetShader(pShader->pPixelShader, NULL, 0);
-  g_pd3dDeviceContext->GSSetShader(pShader->pGeometryShader, NULL, 0);
+    g_pd3dDeviceContext->VSSetShader(pShader->pVertexShader, NULL, 0);
+    g_pd3dDeviceContext->PSSetShader(pShader->pPixelShader, NULL, 0);
+    g_pd3dDeviceContext->GSSetShader(pShader->pGeometryShader, NULL, 0);
+  }
 
   return true;
```
samsinsane: If this function always returns `true`, why have a return at all?

src/rendering/vcCompass.cpp
```diff
@@ -59,23 +59,18 @@ udResult vcCompass_Destroy(vcAnchor **ppCompass)
   return udR_Success;
 }
 
-udResult vcCompass_Render(vcAnchor *pCompass, vcAnchorStyle anchorStyle, const udDouble4x4 &worldViewProj, const udDouble4 &colour /*= udDouble4::create(1.0, 1.0, 1.0, 1.0)*/)
+bool vcCompass_Render(vcAnchor *pCompass, vcAnchorStyle anchorStyle, const udDouble4x4 &worldViewProj, const udDouble4 &colour /*= udDouble4::create(1.0, 1.0, 1.0, 1.0)*/)
```
samsinsane: This seems like a strange choice? Also, `udDouble4::one()` could be used for the default parameter?

src/rendering/vcFenceRenderer.cpp
```diff
@@ -60,44 +60,60 @@ static vcTexture *gSolidTexture = nullptr;
...
+
+  UD_ERROR_IF(!vcTexture_CreateFromFilename(&gArrowTexture, "asset://assets/textures/fenceArrow.png", nullptr, nullptr, vcTFM_Linear, true), udR_InternalError);
+  UD_ERROR_IF(!vcTexture_CreateFromFilename(&gGlowTexture, "asset://assets/textures/fenceGlow.png", nullptr, nullptr, vcTFM_Linear, true), udR_InternalError);
+  UD_ERROR_IF(!vcTexture_CreateFromFilename(&gDiagonalTexture, "asset://assets/textures/fenceDiagonal.png", nullptr, nullptr, vcTFM_Linear, true), udR_InternalError);
+
+  UD_ERROR_CHECK(vcTexture_Create(&gSolidTexture, 1, 1, whitePixel));
+
+  result = udR_Success;
+epilogue:
+
```
samsinsane: You've got the spacing around the wrong way here. Usually we have the empty line _before_ the epilogue.

src/rendering/vcFenceRenderer.cpp
```diff
@@ -60,44 +60,60 @@ static vcTexture *gSolidTexture = nullptr;
...
+  if (result != udR_Success)
+    vcFenceRenderer_Destroy();
+  return result;
```
samsinsane: Should be an empty line before this.

src/rendering/vcFenceRenderer.cpp
```diff
@@ -60,44 +60,60 @@ static vcTexture *gSolidTexture = nullptr;
...
+
+  vcTexture_Destroy(&gArrowTexture);
+  vcTexture_Destroy(&gGlowTexture);
+  vcTexture_Destroy(&gSolidTexture);
+  vcTexture_Destroy(&gDiagonalTexture);
+
+  result = udR_Success;
+epilogue:
+
+  return result;
```
samsinsane: This only returns success?

src/rendering/vcFenceRenderer.cpp
```diff
@@ -109,17 +125,21 @@ udResult vcFenceRenderer_Create(vcFenceRenderer **ppFenceRenderer)
...
+  UD_ERROR_CHECK(vcFenceRenderer_Init());
 
-  vcFenceRenderer_Init();
   *ppFenceRenderer = pFenceRenderer;
   pFenceRenderer = nullptr;
-
+  result = udR_Success;
 epilogue:
-  udFree(pFenceRenderer);
+
```
samsinsane: Empty line goes before the epilogue.

src/rendering/vcFenceRenderer.cpp
```diff
@@ -60,44 +60,60 @@ static vcTexture *gSolidTexture = nullptr;
...
+  UD_ERROR_IF(gRefCount != 0, udR_Success);
+
+  vcTexture_Destroy(&gArrowTexture);
+  vcTexture_Destroy(&gGlowTexture);
+  vcTexture_Destroy(&gSolidTexture);
+  vcTexture_Destroy(&gDiagonalTexture);
+
+  result = udR_Success;
+epilogue:
+
```
samsinsane: Empty line before the epilogue.

src/rendering/vcSceneLayerRenderer.cpp
```diff
@@ -32,8 +32,7 @@ udResult vcSceneLayerRenderer_Create(vcSceneLayerRenderer **ppSceneLayerRenderer
 epilogue:
   if (pSceneLayerRenderer != nullptr)
   {
```
samsinsane: Braces can go too?

src/rendering/vcSceneLayerRenderer.cpp
```diff
@@ -32,8 +32,7 @@ udResult vcSceneLayerRenderer_Create(vcSceneLayerRenderer **ppSceneLayerRenderer
...
-    vcSceneLayer_Destroy(&pSceneLayerRenderer->pSceneLayer);
-    udFree(pSceneLayerRenderer);
```
samsinsane: Does this happen somewhere else now? I feel like I added this because it leaks in certain situations?

src/gl/directx11/vcShader.cpp
peterAdams1014: Some graphics APIs can detect an erroneous state (openGL). Having this not 'fail' on `nullptr` is correct for our usage (we use vcShader_Bind(nullptr) for unbinding, however I'm unsure what to do here in d3d yet.)

src/rendering/vcCompass.cpp
peterAdams1014: More consistent with our other rendering operations to return `bool` rather than `udResult`.
Also the comment was wrong, I'll fix it.

src/rendering/vcFenceRenderer.cpp
peterAdams1014: Arg! Fixing.

src/rendering/vcFenceRenderer.cpp
peterAdams1014: Yes. Intentional so future additions to the function won't need error handling code added back in.

src/rendering/vcSceneLayerRenderer.cpp
peterAdams1014: No I don't believe so. I just changed this because it's more consistent with our error handling. I will revert it however, not worth the risk.

PR #28, opened 2019-10-02
vcDBF - DBase handling

DamianMadden:


src/vcDBF.h
```diff
@@ -0,0 +1,64 @@
...
+udResult vcDBF_FindFieldIndex(vcDBF *pDBF, const char *pFieldName, uint16_t *pIndex);
+udResult vcDBF_GetFieldIndex(vcDBF *pDBF, const char *pFieldName, uint16_t *pIndex);
+uint16_t vcDBF_GetFieldCount(vcDBF *pDBF);
+uint32_t vcDBF_GetRecordCount(vcDBF *pDBF);
+
+// These two can only be performed if there's no data in pDBF
+udResult vcDBF_AddField(vcDBF *pDBF, char *pFieldName, char fieldType, uint8_t fieldLen);
+udResult vcDBF_RemoveField(vcDBF *pDBF, uint16_t fieldIndex);
+
+// Record indexes here start from 0 not 1, for SHP files need to pass in recNumber - 1
```
samsinsane: If DBase files usually use an index starting from 1 it might be better to follow that system? Also, I appreciate that you're also writing the SHP file reader but this comment shouldn't contain anything to do with that. It implies that every usage of this system should also update this comment and document what index it starts at.

src/vcDBF.h
```diff
@@ -0,0 +1,64 @@
...
+udResult vcDBF_RecordReadFieldDouble(vcDBF_Record *pRecord, double *pValue, uint16_t fieldIndex);
+udResult vcDBF_RecordReadFieldInt(vcDBF_Record *pRecord, int32_t *pValue, uint16_t fieldIndex);
+udResult vcDBF_RecordReadFieldString(vcDBF_Record *pRecord, const char **ppValue, uint16_t fieldIndex);
+udResult vcDBF_RecordReadFieldMemo(vcDBF *pDBF, vcDBF_Record *pRecord, const char **ppValue, uint16_t fieldIndex);
+
+udResult vcDBF_RecordWriteFieldBool(vcDBF_Record *pRecord, bool value, uint16_t fieldIndex);
+udResult vcDBF_RecordWriteFieldDouble(vcDBF_Record *pRecord, double value, uint16_t fieldIndex);
+udResult vcDBF_RecordWriteFieldInt(vcDBF_Record *pRecord, int32_t value, uint16_t fieldIndex);
+udResult vcDBF_RecordWriteFieldString(vcDBF_Record *pRecord, const char *pValue, uint16_t fieldIndex);
+udResult vcDBF_RecordWriteFieldMemo(vcDBF_Record *pRecord, vcDBF *pDBF, const char *pValue, uint16_t fieldIndex);
```
samsinsane: I think I told you to put the parameters in this order and I apologise if I did; I think it makes more sense to have the `fieldIndex` in the middle. "This record, this field, this value" instead of "This record, this value, this field", thoughts?

src/vcDBF.cpp
```diff
@@ -0,0 +1,1159 @@
...
+  uint32_t i;
+
+  UD_ERROR_CHECK(udFile_Load(pFilename, &pData, &fileLength));
+  
+  pDBF->memoData.newIndex = (uint32_t)*pData;
+  vcDBF_FlipEndian(&pDBF->memoData.newIndex);
+
+  pDBF->memoData.memoBlockSize = *((uint16_t *)pData + 4);
+
+  pDBF->memoData.firstIndex = (uint32_t)(512 / pDBF->memoData.memoBlockSize + .5); // Round up
```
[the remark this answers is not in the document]
DamianMadden: but I already need to cast to uint32_t anyway

src/vcDBF.cpp
```diff
@@ -0,0 +1,1159 @@
...
+
+  uint8_t blank[10] = {};
+  uint8_t strLen = 0;
+  for (; pDesc->fieldName[strLen] != '\0' && strLen < 11; ++strLen);
+
+  // All these fields are single byte numbers or unused, no need to flip endian
+  UD_ERROR_CHECK(udFile_Write(pFile, &pDesc->fieldName, strLen));
+
+  if (strLen < 11)
+    UD_ERROR_CHECK(udFile_Write(pFile, blank, 11 - strLen));
```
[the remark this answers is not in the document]
DamianMadden: Yes, the specification indicates that the rest of this field is padded with nulls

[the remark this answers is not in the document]
samsinsane: I thought the specification indicated it was padded with spaces, if that's not the case then you can resolve this comment as it was a misunderstanding.

src/vcDBF.h
```diff
@@ -0,0 +1,80 @@
...
+  {
+    char *pString;
+    double floaty;
+    bool logical;
+    int integer;
+  } *pFields;
+};
+
+template <typename T>
+void vcDBF_FlipEndian(T *pData)
```
samsinsane: This doesn't need to be in the header, it's only used in the cpp.

DamianMadden: Yes, but this functionality is needed elsewhere, specifically for shapefiles

PR #30, opened 2019-10-04
Make SDL detect mouse events on gaining focus

DamianMadden:
Resolves [AB#352](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/352)

### DIFF
--- a/src/imgui_ex/imgui_impl_sdl.cpp
+++ b/src/imgui_ex/imgui_impl_sdl.cpp
@@ -328,7 +328,11 @@ static bool    ImGui_ImplSDL2_Init(SDL_Window* window)
     if ((io.ConfigFlags & ImGuiConfigFlags_ViewportsEnable) && (io.BackendFlags & ImGuiBackendFlags_PlatformHasViewports))
       ImGui_ImplSDL2_InitPlatformInterface(window);
 
+    // Block here copied from InitPlatformInterface below
+    // SDL2 by default doesn't pass mouse clicks to the application when the click focused a window. This is getting in the way of our interactions and we disable that behavior.
+#if SDL_HAS_MOUSE_FOCUS_CLICKTHROUGH
     SDL_SetHint(SDL_HINT_MOUSE_FOCUS_CLICKTHROUGH, "1");
+#endif
 
     return true;
 }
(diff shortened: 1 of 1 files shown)


src/imgui_ex/imgui_impl_sdl.cpp
```diff
@@ -328,6 +328,12 @@ static bool    ImGui_ImplSDL2_Init(SDL_Window* window)
     if ((io.ConfigFlags & ImGuiConfigFlags_ViewportsEnable) && (io.BackendFlags & ImGuiBackendFlags_PlatformHasViewports))
       ImGui_ImplSDL2_InitPlatformInterface(window);
 
+    // Block here copied from InitPlatformInterface below
+    // SDL2 by default doesn't pass mouse clicks to the application when the click focused a window. This is getting in the way of our interactions and we disable that behavior.
+#if SDL_HAS_MOUSE_FOCUS_CLICKTHROUGH
+    SDL_SetHint(SDL_HINT_MOUSE_FOCUS_CLICKTHROUGH, "1");
```
samsinsane: What's the plan for platforms that don't support this? Such as Ubuntu 16.04?

DamianMadden: To do nothing, I considered making some default action take over when the window gains focus until mouse is released, but it fails ungracefully if the user alt-tabs in, SDL doesn't detect the mouse being down at all without this hint, so there's probably not much that can be done

samsinsane: > To do nothing

I'm not sure this is acceptable, Ubuntu 16.04 is a supported platform. I believe we've had customers ask about Ubuntu 14.04, so I would imagine that we also have customers using 16.04 too.

@OsiJr What's the plan for issues like this?

OsiJr: This functionality used to work on all platforms- I'd be more interested in finding out why its _now_ broken when we never had this Hint enabled before. Are we now ignoring events elsewhere if the window isn't focused?

DamianMadden: If what changed was something between SDL versions then I imagine it may already work as expected on ubuntu. What events do you mean?

PR #31, opened 2019-10-04
SLPK Optimisations

DamianMadden:


src/vcSceneLayerHelper.h
```diff
@@ -8,93 +8,111 @@ template <typename T>
 size_t vcSceneLayerHelper_ReadSceneLayerType(T *pValue, char *pData, const char *pType)
 {
   size_t size = 0;
-  if (udStrEqual(pType, "Int64"))
+  if (*pType == 'I')
```
OsiJr: This is a somewhat misguided optimisation- this function shouldn't be hit that often for it to make any measurable difference and the original implementation was easier to scroll through and see the strings aligned visually it also makes it harder to reorder the list so the most common types are at the top if we do need to optimize this in future (which is more likely to make a bigger difference to how fast this runs on most datasets.

The (currently empty) block for a totally unexpected type should probably be put in each branch at least.

PR #32, opened 2019-10-04
Set read position using byte offset from slpk files

DamianMadden:
Mostly resolves [AB#360](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/360) (?) the verts and normals are correct now at least, but the objects in the scene still phase out of existence depending on where the camera is positioned


DamianMadden:
Also tested with other slpk's and I seem not to have broken anything

PR #33, opened 2019-10-07
Updated VDK to Updated 0.4.0 API

OsiJr:


buildscripts/build.sh
```diff
@@ -14,7 +14,7 @@ else
 fi
 
 export DEPLOYDIR="$DEV/Builds/vault/client/Pipeline_$CI_PIPELINE_ID"
-export VAULTSDK_HOME="$DEV/Builds/vault/linkedvdk/Pipeline_43776"
+export VAULTSDK_HOME="$DEV/Builds/vault/linkedvdk/Pipeline_44784"
```
samsinsane: Probably don't need to update this file, it's not used in the build process anymore. However, keeping it in sync makes sense while it's still in the repo.

PR #36, opened 2019-10-08
Shader uniform name changes for consistency

peterAdams1014:
Removed incorrect ubo in polygon shader


src/gl/metal/Shaders.metal
```diff
@@ -539,27 +539,23 @@ struct PNUVSOutput
 
 struct PNUVSUniforms1
```
samsinsane: Should this 1 be changed to 3? Based on the previous code, this number indicates the number of members?

PR #43, opened 2019-10-10
Don't try to login without any server + error message

DamianMadden:
Resolves [AB#375](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/375)


samsinsane:
There's something wrong with this branch? There's 12 commits, but only 4 unique commit messages?

PR #56, opened 2019-10-14
Fixed Map-mode zoom out issue.

icyvapor:
https://dev.azure.com/euclideon/vaultclient/_workitems/edit/102

Tried lerping the projection matrix, but most of the visual change happened during the last leg of lerping. So instead, switch to ortho right after the camera starts rotating downward. 

First tried a threshold of 0.5, that's still a bit visually jarring. Settled to 0.15 after several trials.

### DIFF
--- a/src/vcCamera.cpp
+++ b/src/vcCamera.cpp
@@ -109,7 +109,7 @@ void vcCamera_BeginCameraPivotModeMouseBinding(vcState *pProgramState, int bindi
   };
 }
 
-void vcCamera_UpdateMatrices(vcCamera *pCamera, const vcCameraSettings &settings, const udFloat2 &windowSize, const udFloat2 *pMousePos = nullptr)
+void vcCamera_UpdateMatrices(vcCamera *pCamera, const vcCameraSettings &settings, vcCameraInput* pCamInput, const udFloat2 &windowSize, const udFloat2 *pMousePos = nullptr)
 {
   // Update matrices
   double fov = settings.fieldOfView;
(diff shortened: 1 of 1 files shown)


src/vcCamera.cpp
```diff
@@ -126,22 +126,39 @@ void vcCamera_UpdateMatrices(vcCamera *pCamera, const vcCameraSettings &settings
...
   default:
-    pCamera->matrices.projectionUD = udDouble4x4::perspectiveZO(fov, aspect, zNear, zFar);
+    pCamera->matrices.projectionUD = projectionPerspUD;
 #if GRAPHICS_API_OPENGL
-    pCamera->matrices.projection = udDouble4x4::perspectiveNO(fov, aspect, zNear, zFar);
+    pCamera->matrices.projection = projectionPersp;
 #endif
   }
 
+
```
samsinsane: We generally limit spacing between blocks to one blank line, but I don't think it's worth holding this fix up over.

src/vcCamera.cpp
```diff
@@ -126,22 +126,39 @@ void vcCamera_UpdateMatrices(vcCamera *pCamera, const vcCameraSettings &settings
...
-  case vcCM_FreeRoam: // fall through
+  case vcCM_FreeRoam:
+    if (pCamInput->transitioningToMapMode && pCamInput->progress > 0.15)
+    {
+      //Switch to ortho projection soon after camera starts rotating downward, hence the progress threshold 0.15.
+      pCamera->matrices.projectionUD = projectionOrthoUD;
+#if GRAPHICS_API_OPENGL
+      pCamera->matrices.projection = projectionOrtho;
+#endif
+      break;
```
OsiJr: This break is causing issues- if you intended this to always break you'll need to put it outside the brace below

OsiJr:
Can you rebase this onto latest so we can rule out why those changes are here?

src/vcCamera.cpp
icyvapor: It's intentionally only break here when the `if` condition holds.  Otherwise this case becomes 'fall through' as how it was before. 
What issues is it causing?

src/vcCamera.cpp
icyvapor: thanks, gotta modify the code anyhow, will remove.

src/vcCamera.cpp
OsiJr: Yeah, we have strict compilation settings on so we don't get unexpected results later.

PR #59, opened 2019-10-14
Moved get/got geo code out of HandleImGui

DamianMadden:
Works for submodels in folders now

Resolves [AB#356](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/356)

### DIFF
--- a/src/scene/vcFolder.cpp
+++ b/src/scene/vcFolder.cpp
@@ -61,7 +61,7 @@ void vcFolder::AddToScene(vcState *pProgramState, vcRenderData *pRenderData)
   {
     if (pNode->pUserData != nullptr)
     {
-      vcSceneItem *pSceneItem = (vcSceneItem*)pNode->pUserData;
+      vcSceneItem *pSceneItem = (vcSceneItem *)pNode->pUserData;
 
       pSceneItem->AddToScene(pProgramState, pRenderData);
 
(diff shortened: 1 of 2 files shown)


src/vcState.h
```diff
@@ -173,7 +174,7 @@ struct vcState
   vcDocks changeActiveDock;
 
   bool getGeo;
-  udGeoZone *pGotGeo;
+  vcModel *pGotGeo;
```
samsinsane: I'm not sure I understand why this was changed?

src/scene/vcFolder.cpp
```diff
@@ -93,8 +93,24 @@ void vcFolder::AddToScene(vcState *pProgramState, vcRenderData *pRenderData)
...
+      vcModel *pModel = (vcModel *)pNode->pUserData;
+      if (pProgramState->pGotGeo == nullptr)
+      {
+        pProgramState->getGeo = false;
+        pProgramState->pGotGeo = pModel;
+        vcProject_UseProjectionFromItem(pProgramState, pModel);
+      }
+    }
+
+    if (pProgramState->pGotGeo != nullptr)
```
samsinsane: Why is this after the above block now? It used to be before?

src/vcState.h
OsiJr: Changing this type is wrong. Lots of other things are sources for this and I already have changes that change this to support more types.

src/scene/vcFolder.cpp
DamianMadden: I was mistakenly thinking we needed to changeprojection on the source model, which previously might have happened after cycling through every scene item and finding the source model again before unsetting gotGeo

src/scene/vcFolder.cpp
```diff
@@ -93,8 +93,25 @@ void vcFolder::AddToScene(vcState *pProgramState, vcRenderData *pRenderData)
 
     HandleNodeSelection(pProgramState, pNode);
 
+    if (pProgramState->pGotGeo != nullptr)
+    {
+      ((vcSceneItem *)(pNode->pUserData))->ChangeProjection(*pProgramState->pGotGeo);
+    }
+    else if (pProgramState->pGotGeo == nullptr && pProgramState->getGeo && pNode->itemtype == vdkPNT_PointCloud && ((vcModel *)(pNode->pUserData))->m_pPreferredProjection != nullptr)
```
samsinsane: The `pProgramState->pGotGeo == nullptr` is unnecessary due to the `if` statement.

PR #64, opened 2019-10-15
Disable Orbit with controller in mapmode

DamianMadden:
Resolves [AB#324](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/324)

### DIFF
--- a/src/vcCamera.cpp
+++ b/src/vcCamera.cpp
@@ -655,7 +655,7 @@ void vcCamera_HandleSceneInput(vcState *pProgramState, udDouble3 oscMove, udFloa
       {
         pProgramState->isUsingAnchorPoint = true;
         pProgramState->worldAnchorPoint = pProgramState->worldMousePosCartesian;
-        pProgramState->cameraInput.inputState = vcCIS_Orbiting;
+        pProgramState->cameraInput.inputState = pProgramState->settings.camera.cameraMode == vcCM_OrthoMap ? vcCIS_Panning : vcCIS_Orbiting;
         vcCamera_StopSmoothing(&pProgramState->cameraInput);
       }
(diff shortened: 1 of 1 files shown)


src/vcCamera.cpp
```diff
@@ -655,7 +655,7 @@ void vcCamera_HandleSceneInput(vcState *pProgramState, udDouble3 oscMove, udFloa
       {
         pProgramState->isUsingAnchorPoint = true;
         pProgramState->worldAnchorPoint = pProgramState->worldMousePosCartesian;
-        pProgramState->cameraInput.inputState = vcCIS_Orbiting;
+        pProgramState->cameraInput.inputState = pProgramState->settings.camera.cameraMode == vcCM_OrthoMap ? vcCIS_Panning : vcCIS_Orbiting;
```
samsinsane: I wonder if perhaps the trigger checks should be:
```c++
if (isLeftTriggerHeld || (isRightTriggerHeld && pProgramState->settings.camera.cameraMode == vcCM_OrthoMap))
```
and
```c++
if (isRightTriggerHeld && pProgramState->settings.camera.cameraMode != vcCM_OrthoMap)
```
I'm not super familiar with this module, so this might just make it unnecessarily difficult to process. Thoughts?

DamianMadden: From the guide: - `Left Trigger` - Increase Move Speed (sprint), so that shouldn't pan I guess, but I think for either map mode we do want to do everything there except pan instead of orbit

samsinsane: I assumed the triggers mapped to the mouse buttons, which doesn't really make sense in hindsight. Sorry about that!

PR #82, opened 2019-10-21
Fix scene explorer selection to work with scene tab closed

DamianMadden:
Resolves [AB#336](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/336)

### DIFF
--- a/src/scene/vcFolder.cpp
+++ b/src/scene/vcFolder.cpp
@@ -231,7 +231,20 @@ void vcFolder::HandleImGui(vcState *pProgramState, size_t *pItemID)
       bool sceneExplorerItemClicked = ((ImGui::IsMouseReleased(0) && ImGui::IsItemHovered() && !ImGui::IsItemActive()) || (!pSceneItem->m_selected && ImGui::IsItemActive()));
       if (sceneExplorerItemClicked)
       {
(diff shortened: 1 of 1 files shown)


src/scene/vcFolder.cpp
```diff
@@ -231,7 +231,20 @@ void vcFolder::HandleImGui(vcState *pProgramState, size_t *pItemID)
       bool sceneExplorerItemClicked = ((ImGui::IsMouseReleased(0) && ImGui::IsItemHovered() && !ImGui::IsItemActive()) || (!pSceneItem->m_selected && ImGui::IsItemActive()));
       if (sceneExplorerItemClicked)
       {
-        udStrcpy(pProgramState->sceneExplorer.selectUUIDWhenPossible, pNode->UUID);
```
samsinsane: What ramifications are there to not using this system? Why was it used in the first place?

DamianMadden: I think it just seemed like a nice way to consolidate the code into a single place, which it would be if there were a single place at this point where every node in the tree were iterated through every frame. The only ramification I see is more code, I think initially that system was implemented for the case where you wanted to select a new scene item that wasn't yet instantiated

PR #110, opened 2019-10-28
Moved UDP parser to parsers

OsiJr:
This is in preparation for more parsers that are required- 12DXML for example.


OsiJr:
> Should it be in `parsers/legacy/vcUDP.cpp`?

I made the call to not do that in order to not set a precedent for other formats that are no longer recommended or supported by their creators or a committee but are in ultra wide-spread usage still (like OBJ) due to potential arguments that will arise purely from where they live in our repo "we still use format X so it can't be legacy"

PR #114, opened 2019-10-28
Polygon model implementation of image selection/highlighting

DamianMadden:


DamianMadden:
@peterAdams1014 Could use some help getting the highlight shader to run on this, atm the selection itself works but just the highlight is missing

OsiJr:
I was thinking the non-billboard meshes at this point can just go through the normal polygon pipeline anyway as everything they need exists there now (it didn't originally) and the @peterAdams1014 just needs to come up with a solution for the billboards. Fortunately billboards can be picked entirely on the CPU anyway (we know the depth and size in screen pixels) which avoids needing to do the post pass on those.

peterAdams1014:
There are too many unrelated changes in this PR.

I can see how you arrived at this solution - and while it may work correctly - I think it could be improved. I've been working on picking recently and I had some different thoughts on how to implement this task. Unfortunately I think this is an example of a failure to communicate from our team.

src/main.cpp
```diff
@@ -441,6 +441,7 @@ void vcMain_MainLoop(vcState *pProgramState)
             if (clicked.pParent != nullptr && clicked.pItem->itemtype == vdkPNT_Media)
             {
               vdkProjectNode_SetURI(pProgramState->activeProject.pProject, clicked.pItem, pNextLoad);
+              ((vcSceneItem *)clicked.pItem->pUserData)->m_loadStatus = vcSLS_Unloaded;
```
OsiJr: I don't think this is required? The OnNodeUpdate system should trigger for this change.

src/rendering/vcImageRenderer.cpp
```diff
@@ -90,10 +87,6 @@ bool vcImageRenderer_Render(vcImageRenderInfo *pImageInfo, const udDouble4x4 &vi
   udDouble4x4 mvp = viewProjectionMatrix * udDouble4x4::translation(pImageInfo->position) * udDouble4x4::rotationYPR(pImageInfo->ypr) * udDouble4x4::scaleUniform(pImageInfo->scale);
   if (pImageInfo->type == vcIT_OrientedPhoto)
     mvp = mvp * udDouble4x4::scaleNonUniform(worldScale, worldScale, worldScale * aspect);
```
OsiJr: Oriented Photo can also be removed

src/rendering/vcInternalModels.h
```diff
@@ -29,6 +29,8 @@ enum vcInternalModelType
   vcInternalModelType_Sphere,
   vcInternalModelType_Cylinder,
 
+  vcInternalModelType_Panorama,
```
OsiJr: `Tube` or `Pipe` would be more consistent with the other names here

src/scene/vcMedia.cpp
```diff
@@ -87,13 +91,13 @@ void vcMedia::OnNodeUpdate(vcState *pProgramState)
...
-
-    if (m_image.pTexture != nullptr)
-      vcTexture_Destroy(&m_image.pTexture);
+    vcTexture_Destroy(&m_image.pTexture);
   }
-
-  if (m_loadStatus == vcSLS_Loaded && ((m_pNode->pURI != nullptr && !udStrEqual(m_pLoadedURI, m_pNode->pURI)) || (m_reloadTimeSecs != 0.0 && m_loadLoadTimeSec + m_reloadTimeSecs < udGetEpochSecsUTCf())))
+  else if ((m_loadStatus == vcSLS_Loaded && !udStrEqual(m_pLoadedURI, m_pNode->pURI)) ||
+    (m_reloadTimeSecs != 0.0 && m_loadLoadTimeSec + m_reloadTimeSecs < udGetEpochSecsUTCf()) ||
+    (m_loadStatus == vcSLS_Unloaded))
```
OsiJr: New lines aren't required here, if you want to split it out, bool variables are preferred

src/scene/vcMedia.cpp
```diff
@@ -87,13 +91,13 @@ void vcMedia::OnNodeUpdate(vcState *pProgramState)
...
   {
+    vcTexture_Destroy(&m_image.pTexture);
```
OsiJr: Is this required? Uploading new pixels isn't sufficient rather than recreating the texture if the size hasn't changed _after_ the reload occurs?

src/scene/vcMedia.cpp
```diff
@@ -144,28 +171,41 @@ void vcMedia::AddToScene(vcState *pProgramState, vcRenderData *pRenderData)
...
     return;
 
-  if (m_image.pTexture != nullptr)
+  if (m_image.pTexture != nullptr && m_loadStatus == vcSLS_Loaded)
   {
-    // For now brute force sorting (n^2)
-    double distToCameraSqr = udMagSq3(m_image.position - pProgramState->pCamera->position);
-    size_t i = 0;
-    for (; i < pRenderData->images.length; ++i)
+    if (m_image.type == vcIT_StandardPhoto || m_image.type == vcIT_OrientedPhoto)
```
OsiJr: Oriented Photo can go through the poly pipe as well

src/scene/vcMedia.h
```diff
@@ -39,6 +40,7 @@ class vcMedia : public vcSceneItem
 
 public:
   const char *m_pLoadedURI; // The URI of the media we have (or are loading) currently
+  vcPolygonModel *m_pModel;
```
OsiJr: This doesn't need to be cached? No other system caches these.

src/main.cpp
DamianMadden: Crashes without this

src/scene/vcMedia.cpp
DamianMadden: That'd be the same poly model as a billboard right? I'll add them together in future PR

OsiJr: Billboard is different, Oriented is just a normal mesh, Billboard needs to be expanded in screen space.

src/main.cpp
OsiJr: That's a very concerning crash as this will crash when in collaborative mode as well.

src/scene/vcMedia.cpp
DamianMadden: same model different scale though right?

OsiJr: Shouldn't be. It might be currently but that would be wrong.

The oriented model should be a 0-1 square and the billboard should be all positions at {0,0,0} and corners to say which way to expand.

src/scene/vcMedia.cpp
```diff
@@ -144,35 +143,106 @@ void vcMedia::AddToScene(vcState *pProgramState, vcRenderData *pRenderData)
...
+        {
+          pPoly->pModel = gInternalModels[vcInternalModelType_Tube];
+          pPoly->worldMat *= udDouble4x4::scaleNonUniform(worldScale, worldScale, worldScale * aspect * UD_PI);
+          pPoly->insideOut = true;
+        }
+        else if (m_image.type == vcIT_OrientedPhoto)
+        {
+          pPoly->pModel = gInternalModels[vcInternalModelType_Quad];
+          pPoly->worldMat *= udDouble4x4::scaleNonUniform(worldScale, worldScale, worldScale * aspect) * udDouble4x4::rotationZ(UD_PI);
+          pPoly->insideOut = true;
```
OsiJr: I don't think this one should be true?

OsiJr: @CarlCFX I'm going to merge this- can you check if they are shown on the wrong side by default?

PR #115, opened 2019-10-28
Reorganised where projects are saved and added defaultCRS information

OsiJr:
Note that for this to work fully it requires VDK0.4.1 due to a bug this exposed in root node metadata

### DIFF
--- a/src/vcProject.cpp
+++ b/src/vcProject.cpp
@@ -60,7 +60,11 @@ bool vcProject_InitFromURI(vcState *pProgramState, const char *pFilename)
       temp.SetFilenameWithExt("");
       pProgramState->activeProject.pRelativeBase = udStrdup(temp.GetPath());
 
-      pProgramState->getGeo = true;
+      int32_t recommendedSRID = -1;
+      if (vdkProjectNode_GetMetadataInt(pProgramState->activeProject.pRoot, "defaultcrs", &recommendedSRID, -1) == vE_Success && recommendedSRID >= 0 && udGeoZone_SetFromSRID(&zone, recommendedSRID) == udR_Success)
+        vcGIS_ChangeSpace(&pProgramState->gis, zone);
+      else
+        pProgramState->getGeo = true;
     }
     else
     {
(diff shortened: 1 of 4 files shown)


src/vcProject.cpp
```diff
@@ -129,6 +133,50 @@ void vcProject_Deinit(vcState *pProgramData, vcProject *pProject)
...
+      exportFilename.SetFromFullPath("%s/untitled_project.json", pPath);
+    else if (exportFilename.HasFilename())
+      exportFilename.SetExtension(".json");
+    else
+      exportFilename.SetFilenameWithExt("untitled_project.json");
+
+    // Check if file path exists before writing to disk, and if so, the user will be presented with the option to overwrite or cancel
+    if (allowOverride || vcModals_OverwriteExistingFile(exportFilename.GetPath()))
+    {
+      vcState::ErrorItem projectError;
```
peterAdams1014: Potential uninitialized memory

src/vcProject.cpp
```diff
@@ -129,6 +133,50 @@ void vcProject_Deinit(vcState *pProgramData, vcProject *pProject)
...
+        projectError.resultCode = udR_Success;
+      else
+        projectError.resultCode = udR_WriteFailure;
+
+      pProgramState->errorItems.PushBack(projectError);
+
+      vcModals_OpenModal(pProgramState, vcMT_ProjectChange);
+    }
+
+    if (pDir != nullptr)
```
peterAdams1014: Unnecessary `nullptr` check.

src/vcProject.cpp
```diff
@@ -129,6 +133,50 @@ void vcProject_Deinit(vcState *pProgramData, vcProject *pProject)
...
+      udFree(pDir);
```
DamianMadden: nullptr check unnecessary, also udCloseDir?

src/vcProject.cpp
OsiJr: This is currently just moved code- do you want me to also make fixes (which complicates this MR)?

PR #123, opened 2019-10-29
Added ability to save and load polygon models

OsiJr:
[AB#442](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/442)

### DIFF
--- a/src/scene/vcPolyModelNode.cpp
+++ b/src/scene/vcPolyModelNode.cpp
@@ -26,12 +26,31 @@ vcPolyModelNode::vcPolyModelNode(vdkProject *pProject, vdkProjectNode *pNode, vc
   OnNodeUpdate(pProgramState);
 }
 
-void vcPolyModelNode::OnNodeUpdate(vcState * /*pProgramState*/)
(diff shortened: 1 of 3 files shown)


src/scene/vcPolyModelNode.cpp
```diff
@@ -26,12 +26,31 @@ vcPolyModelNode::vcPolyModelNode(vdkProject *pProject, vdkProjectNode *pNode, vc
...
+    m_invert = !udStrEqual(pTemp, "back");
 
-  m_invert = !udStrEqual(pTemp, "back");
+  if (m_pNode->geomCount != 0)
+  {
+    udDouble3 *pPosition = nullptr;
+    udDouble3 scale;
+    udDouble3 euler;
+
+    vcProject_FetchNodeGeometryAsCartesian(pProgramState->activeProject.pProject, m_pNode, pProgramState->gis.zone, &pPosition, nullptr);
```
samsinsane: Is there a possibility that `pPosition` could still be `nullptr` after this call?

OsiJr: Only if there were no nodes which is checked above

PR #128, opened 2019-10-30
More submitted Chinese changes from Hongzhi Liu

OsiJr:


builds/assets/lang/zhCN.json
```diff
@@ -429,6 +451,7 @@
   "scenePOIThumbnailSizeLarge": "大",
   "scenePOILabelImageType": "图像类型",
   "scenePOILabelImageTypeStandard": "标准",
+  "scenePOILabelImageTypeOriented": "Oriented Image",
```
samsinsane: I just want to confirm that this is correct?

OsiJr: I doubt it's correct but it is what was submitted :/

PR #133, opened 2019-10-31
Enable map mode shortcut for mac

DamianMadden:
Resolves [AB#103](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/103)


DamianMadden:
> ImGui already does the mapping of CMD to Ctrl - there are things that specifically use Ctrl which is why AB#215 exists, but I don't think we should be mapping CMD to Ctrl like this. Additionally, when holding Ctrl on macOS or Linux, pressing M doesn't trigger any events through SDL. As I've previously stated, using Ctrl+M for map mode was a bad idea and this should be reconsidered instead of trying to hack around the OS functionality.

You mean when holding cmd? because pressing M with ctrl on mac does trigger events presently.

PR #136, opened 2019-10-31
Fix orthographic mode cutting off model

DamianMadden:
Resolves [AB#363](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/363)


DamianMadden:
Resolves AB#151

src/main.cpp
```diff
@@ -781,6 +781,7 @@ int main(int argc, char **args)
   programState.settings.camera.nearPlane = 0.5f;
   programState.settings.camera.farPlane = 10000.f;
   programState.settings.camera.fieldOfView = UD_PIf * 5.f / 18.f; // 50 degrees
+  programState.cameraInput.progMultiplier = 1.0;
```
OsiJr: I assume this is `progress`? Might also be `program` though?

src/vcCamera.cpp
```diff
@@ -543,6 +547,8 @@ void vcCamera_SwapMapMode(vcState *pProgramState)
     // defer actually swapping projection mode
     pProgramState->cameraInput.transitioningToMapMode = true;
     lookAtPosition += udDouble3::create(0, 0, -1); // up
+
+    cameraHeight = pProgramState->settings.camera.farPlane;
```
OsiJr: This will mean anything below 0 won't be visible

OsiJr: When @peterAdams1014 added map mode it worked well with auto-selecting a height. Might be worthwhile going through the tree and seeing what he did originally.

DamianMadden: It was farplane / 2

DamianMadden: I think at some point I changed it to fix a bug thinking that whatever height it was at would suffice, incidentally these changes fix the original bug too, that if you swapped map mode right near the end of the transition it would recalculate orthographic size

PR #150, opened 2019-11-06
Fix crash and memory leak

DamianMadden:
Resolves [AB#516](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/516)

Would crash trying to stbi_image_free texture data shared by multiple materials, also would leak material/texture memory if conversion were cancelled

### DIFF
--- a/src/vcFBX.cpp
+++ b/src/vcFBX.cpp
@@ -260,6 +260,38 @@ FbxAMatrix vcFBX_GetGeometryTransformation(FbxNode *inNode)
   return FbxAMatrix(lT, lR, lS);
 }
 
+void vcFBX_CleanMaterials(udChunkedArray<vcFBXMaterial> *pMats)
+{
+  // Clean up
+  for (uint32_t i = 0; i < pMats->length; ++i)
+  {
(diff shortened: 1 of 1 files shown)


src/vcFBX.cpp
```diff
@@ -260,6 +260,38 @@ FbxAMatrix vcFBX_GetGeometryTransformation(FbxNode *inNode)
...
+  {
+    for (uint32_t j = 0; j < (*pMats)[i].textures.length; ++j)
+    {
+      if ((*pMats)[i].textures[j].pPixels)
+      {
+        stbi_image_free((*pMats)[i].textures[j].pPixels);
+
+        // Clean up all references to this allocation
+        void *pPtr = (*pMats)[i].textures[j].pPixels;
+        for (uint32_t k = 0; k < pMats->length; ++k)
```
samsinsane: This should start at `i`, and `l` should start at `j + 1`. The search only needs to go forwards since you've already destroyed the materials and textures up to `i` and `j`.

PR #155, opened 2019-11-07
Fix crash

DamianMadden:
Resolves [AB#516](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/516)

No idea why but assigning nullptr to the very same address here instead of winding back j by 1 still crashes

### DIFF
--- a/src/vcFBX.cpp
+++ b/src/vcFBX.cpp
@@ -275,7 +275,7 @@ void vcFBX_CleanMaterials(udChunkedArray<vcFBXMaterial> *pMats)
         void *pPtr = (*pMats)[i].textures[j].pPixels;
         for (uint32_t k = i; k < pMats->length; ++k)
         {
-          for (uint32_t l = j + 1; l < (*pMats)[k].textures.length; ++l)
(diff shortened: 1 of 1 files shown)


src/vcFBX.cpp
```diff
@@ -275,7 +275,7 @@ void vcFBX_CleanMaterials(udChunkedArray<vcFBXMaterial> *pMats)
         void *pPtr = (*pMats)[i].textures[j].pPixels;
         for (uint32_t k = i; k < pMats->length; ++k)
         {
-          for (uint32_t l = j + 1; l < (*pMats)[k].textures.length; ++l)
+          for (uint32_t l = j; l < (*pMats)[k].textures.length; ++l)
```
samsinsane: To confirm, setting `(*pMats)[i].textures[j].pPixels` to `nullptr` after assigning `pPtr` doesn't work?

DamianMadden: Yeah, I'm as shocked as you, tried like twice

PR #158, opened 2019-11-08
Tile Renderer optimisations

DamianMadden:
Change tile renderer to curate a sorted list of tiles by priority, using priority insert & insertion sort as this was determined to be best in the near sorted case. Rewrite point to rectangle distance function to optimise, see profiler output before and after (note though this was in debug as release doesn't show symbols, could probably change that in the build settings though):

Function Name	Total CPU [unit, %]	Self CPU [unit, %]	Module
| - vcQuadTree_PointToRectDistance	650 (2.21%)	103 (0.35%)	vaultClient_d3d11.exe
| - vcQuadTree_PointToRectDistance	36 (0.19%)	10 (0.05%)	vaultClient_d3d11.exe

Resolves [AB#252](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/252)

### DIFF
--- a/src/rendering/vcTileRenderer.cpp
+++ b/src/rendering/vcTileRenderer.cpp
@@ -58,7 +58,7 @@ struct vcTileRenderer
   struct vcTileCache
   {
     volatile bool keepLoading;
-    udThread *pThreads[8];
+    udThread *pThreads[10];
     udSemaphore *pSemaphore;
     udMutex *pMutex;
     udChunkedArray<vcTileLoadPacked> tileLoadList;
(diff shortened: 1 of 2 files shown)


src/rendering/vcTileRenderer.cpp
```diff
@@ -128,173 +149,157 @@ bool vcTileRenderer_ShouldLoadNode(vcQuadTreeNode *pNode)
   return pNode->renderInfo.tryLoad && pNode->touched && (pNode->renderInfo.loadStatus == vcNodeRenderInfo::vcTLS_InQueue);
 }
 
-void vcTileRenderer_LoadThread(void *pThreadData)
+void vcTileRenderer_PriorityInsert(udChunkedArray<vcTileLoadPacked> *pInput, vcTileRenderer *pRenderer, vcQuadTreeNode *pNode)
```
OsiJr: This function does a lot of stuff that looks like it should be locked?

DamianMadden: This function is used in place of code that previously directly inserted, so in every instance where it's called it should already be locked

peterAdams1014: It's usage is thread safe.

peterAdams1014: This function doesn't appear to work correctly.

PR #159, opened 2019-11-08
Invert function of LOD modifier

DamianMadden:
Resolves [AB#348](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/348)

Also missed some things originally

### DIFF
--- a/src/scene/vcLiveFeed.cpp
+++ b/src/scene/vcLiveFeed.cpp
@@ -295,8 +295,7 @@ vcLiveFeed::vcLiveFeed(vdkProject *pProject, vdkProjectNode *pNode, vcState *pPr
   m_updateFrequency(15.0),
   m_decayFrequency(300.0),
   m_maxDisplayDistance(50000.0),
-  m_pMutex(udCreateMutex()),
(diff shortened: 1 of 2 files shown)


src/scene/vcLiveFeed.cpp
```diff
@@ -306,6 +305,8 @@ vcLiveFeed::vcLiveFeed(vdkProject *pProject, vdkProjectNode *pNode, vcState *pPr
   m_oldestFeedUpdate = m_newestFeedUpdate;
   m_fetchNow = true;
 
+  m_labelLODModifier = 1.0;
```
OsiJr: This was moved out of the initialiser list?

DamianMadden: one of the build machines had a problem with it

OsiJr: There is a rule in C++ that the list here has to be in the same order as they are defined in the header; you probably needed to move it above the mutex.

OsiJr: Most compilers don't care though

PR #161, opened 2019-11-11
Removed cameraPositionInLongLat

OsiJr:


src/main.cpp
```diff
@@ -2305,8 +2299,10 @@ void vcRenderWindow(vcState *pProgramState)
           vdkProjectNode *pNode;
           if (vdkProjectNode_Create(pProgramState->activeProject.pProject, &pNode, pProgramState->activeProject.pRoot, "Camera", vcString::Get("viewpointDefaultName"), nullptr, nullptr) == vE_Success)
           {
+            udDouble3 cameraPositionInLongLat = udGeoZone_CartesianToLatLong(pProgramState->gis.zone, pProgramState->pCamera->position, true);
+
             if (pProgramState->gis.isProjected)
-              vdkProjectNode_SetGeometry(pProgramState->activeProject.pProject, pNode, vdkPGT_Point, 1, (double*)&pProgramState->pCamera->positionInLongLat);
+              vdkProjectNode_SetGeometry(pProgramState->activeProject.pProject, pNode, vdkPGT_Point, 1, (double*)&cameraPositionInLongLat);
```
samsinsane: `&cameraPositionInLongLat.x` will avoid the need to type cast unnecessarily.

OsiJr: I'll address this in a coming PR. I've made the change locally already so it doesn't get missed.

PR #162, opened 2019-11-11
Changed camera movemode to lockAltitude

OsiJr:
Relies on #161

### DIFF
--- a/src/vcCamera.cpp
+++ b/src/vcCamera.cpp
@@ -131,7 +131,6 @@ void vcCamera_UpdateMatrices(vcCamera *pCamera, const vcCameraSettings &settings
 
   pCamera->matrices.camera = vcCamera_GetMatrix(pCamera);
 
-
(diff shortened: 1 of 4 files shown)


src/vcCamera.cpp
```diff
@@ -238,11 +238,11 @@ void vcCamera_Apply(vcState *pProgramState, vcCamera *pCamera, vcCameraSettings
     // Translation
     if (pCamSettings->cameraMode == vcCM_FreeRoam)
     {
-      if (pCamSettings->moveMode == vcCMM_Plane)
+      if (!pCamSettings->lockAltitude)
```
samsinsane: Swap cases around to avoid negative test? I appreciate it's the simple case but `if not condition else` is a weird pattern.

OsiJr: Fixed in a future PR. Already made the change locally so it isn't missed.

PR #165, opened 2019-11-11
Compare models now copies all attributes

samsinsane:


src/vcBPA.cpp
```diff
@@ -921,18 +920,35 @@ vdkError vcBPA_ConvertReadPoints(vdkConvertCustomItem *pConvertInput, vdkPointBu
...
-    *pColour = *pOldColour;
+    for (uint32_t j = 0; j < pData->activeItem.pGrid->pBuffer->attributes.count; ++j)
+    {
+      vdkAttributeDescriptor &oldAttrDesc = pData->activeItem.pGrid->pBuffer->attributes.pDescriptors[j];
+      uint32_t attributeSize = (oldAttrDesc.typeInfo & (vdkAttributeTypeInfo_SizeMask << vdkAttributeTypeInfo_SizeShift));
+
+      // Get attribute old offset and pointer
+      uint32_t attrOldOffset = 0;
+      if (vdkAttributeSet_GetOffsetOfNamedAttribute(&pData->activeItem.pGrid->pBuffer->attributes, oldAttrDesc.name, &attrOldOffset) != vE_Success)
+        continue;
```
DamianMadden: Should this take issue if it fails? assert or something?

samsinsane: For what it's worth, we use the attribute information of the "new model" (old variables in this function) to populate the comparison model (new variables in this function), so at no point should an attribute not exist - @OsiJr will need to correct me here if I've done something wrong. I chose to prioritise the generation and output of the displacement data over anything else, as that's the entire point of the feature. It's inconvenient for the model to not contain the other data, but it's completely useless when it doesn't generate anything because of an oversight on our side.

src/vcBPA.cpp
```diff
@@ -78,13 +78,12 @@ struct vcBPAGrid
     return grid;
   }
 
-  void Init(bool initPointBuffer = true)
+  void Init(vdkStandardAttributeContent standardContent)
```
OsiJr: I'll have a chat with you offline but I think this should be a `vdkStandardAttributeSet`

OsiJr: After offline chat, this is correct for the current version of VDK0.5.0

PR #173, opened 2019-11-21
Multi scene

zengweicheng666:
The realization function is multi-scene rendering, can be used for comparison between scenes.


src/main.cpp
```diff
@@ -390,10 +390,10 @@ void vcMain_MainLoop(vcState *pProgramState)
         else // We need to add it to the scene (hopefully)
         {
           vdkProjectNode *pNode = nullptr;
-
+		  vcRender_SwitchContext(pProgramState, pProgramState->eLoadModelFormCurIndex);
```
OsiJr: Wrong tabs here

src/vcState.h
```diff
@@ -115,10 +122,12 @@ struct vcState
   bool forceLogout;
   double lastServerAttempt;
   double lastServerResponse;
-  vdkContext *pVDKContext;
-  vcRenderContext *pRenderContext;
+  vcContextIndex eCurrentContextIndex;
+  vdkContext *pVDKContext[vcContext_Count];
```
OsiJr: There should only be 1 pVDKContext; This will use multiple licenses for no reason?

src/vcState.h
```diff
@@ -115,10 +122,12 @@ struct vcState
...
+  vcRenderContext *pRenderContext[vcContext_Count];
```
OsiJr: There doesn't need to be multiple render contexts either, the internals can be shared without issue

src/vcState.h
```diff
@@ -62,6 +62,13 @@ enum vcErrorSource
   vcES_ProjectChange
 };
 
+enum vcContextIndex
+{
+	vcContext_Scene,
+	vcContext_History,
+	vcContext_Count
+};
```
OsiJr: Making these temporal with onion skinning would give a better user experience

src/vcState.h
zengweicheng666: I'm making a few changes to support multiple scenarios using one render context.

zengweicheng666: I'm making a few changes to support multiple scene using one render context.

DamianMadden:
So, the idea behind this is to have 2 different scenes rendering, each with a pointcloud of the same object at different times, but receiving the same inputs, to place side by side on screen or tab between and compare?

src/main.cpp
```diff
@@ -2417,6 +2685,33 @@ void vcRenderWindow(vcState *pProgramState)
...
+
+			ImGui::End();
+		}
+		else
+		{
+			ImGui::SetNextWindowSize(size);
+			ImGui::PushStyleVar(ImGuiStyleVar_WindowPadding, ImVec2(2, 2));
+			ImGui::SetNextWindowPos(ImVec2(0, 0));
+
+			bool sceneWindow = ImGui::Begin(udTempStr("%s###scenePresentation", vcString::Get("historyTitle")), &pProgramState->settings.window.windowsOpen[vcDocks_History]);
```
DamianMadden: This will probably collide with the existing scenePresentation window if they can ever exist simultaneously, the part after ### is used as a unique ID if it's present...

DamianMadden:
It looks like huge portions of this copy pasted code could be just refactored to take in some information about which 'scene' it pertains to... But I'm thinking you should somewhere just use a numerical reference to conditionally add items to the render data or conditionally follow through with the render call for those items, with some sentinel indicating to render in all instances, but it may only be the case that you need to differently handle scene items, and the use case I'm imagining wouldn't see any scene item used in more than one instance

OsiJr:
> It looks like huge portions of this copy pasted code could be just refactored to take in some information about which 'scene' it pertains to... But I'm thinking you should somewhere just use a numerical reference to conditionally add items to the render data or conditionally follow through with the render call for those items, with some sentinel indicating to render in all instances, but it may only be the case that you need to differently handle scene items, and the use case I'm imagining wouldn't see any scene item used in more than one instance

@DamianMadden, this PR is being handled offline.

OsiJr:
Closing this PR due to decay. It should be rebased and refreshed before being reopened.

PR #174, opened 2019-11-21
Screenshots

DamianMadden:
F8 Hotkey, menu buttons to capture/bring up settings, saves into C:\Users\<user>\AppData\Roaming\euclideon\client for now, until we implement a way to determine a path platform independently like My Pictures / Camera Roll on phones, etc.


src/vcModals.cpp
```diff
@@ -746,29 +749,154 @@ void vcModals_DrawUnsupportedFiles(vcState *pProgramState)
...
+  }
+  else if (pProgramState->image.pImage == nullptr)
+  {
+    return true;
+  }
+
+  if (pFilename == "")
+    udStrcpy(pFilename, bufLen, vcString::Get("screenshotDefaultName"));
+
+  pProgramState->screenshot.pPixels = udAllocType(uint8_t, pProgramState->sceneResolution.x * pProgramState->sceneResolution.y * 4, udAF_Zero);
```
peterAdams1014: Will this leak if line 784 is hit?

src/vcState.h
```diff
@@ -174,6 +174,17 @@ struct vcState
   vcTranslationInfo languageInfo;
   bool showUI;
   vcDocks changeActiveDock;
+
+  struct screenshot
+  {
+    bool taking;
+    bool renderLabels;
+    bool resize;
+    int format;
```
peterAdams1014: Turn `format` and `res` into enums.

src/vcModals.cpp
peterAdams1014: Also no check for allocation failure?

src/main.cpp
```diff
@@ -2247,6 +2277,12 @@ void vcRenderWindow(vcState *pProgramState)
     vcCamera_SwapMapMode(pProgramState);
 #endif
 
+  if (ImGui::IsKeyPressed(SDL_SCANCODE_F8))
```
OsiJr: PrintScreen?

src/main.cpp
```diff
@@ -2247,6 +2277,12 @@ void vcRenderWindow(vcState *pProgramState)
...
+  {
+    pProgramState->screenshot.taking = true;
+    vcModals_TakeScreenshot(pProgramState, pProgramState->screenshot.outputName, vcMaxPathLength);
```
OsiJr: It would be better to move the configuration to settings so that pressing the button takes it immediately (or as immediately as possible) without the modal popping up.

This will prevent things moving that the user wanted to capture.

An additional setting for "Show Screenshots" to open the display modal optionally would be useful.

src/vcModals.cpp
```diff
@@ -746,6 +749,131 @@ void vcModals_DrawUnsupportedFiles(vcState *pProgramState)
...
+
+  vcTexture_BeginReadPixels(pProgramState->image.pImage, 0, 0, pProgramState->sceneResolution.x, pProgramState->sceneResolution.y, pProgramState->screenshot.pPixels, pProgramState->pRenderContext->pFramebuffer[1]);
```
OsiJr: It would be useful to `vcTexture_SaveToFile` to have this functionality rather than doing it in a modal.

src/vcModals.h
```diff
@@ -22,6 +24,7 @@ enum vcModalTypes
   vcMT_ConvertAdd,
   vcMT_ConvertOutput,
   vcMT_ConvertTempDirectory,
+  vcMT_screenshot,
```
OsiJr: This is inconsistent with the other modal names (but I would encourage removing the modal anyway).

src/vcModals.cpp
```diff
@@ -746,29 +749,154 @@ void vcModals_DrawUnsupportedFiles(vcState *pProgramState)
...
+
+  udFree(pProgramState->screenshot.pPixels);
+  pProgramState->screenshot.taking = false;
+
+  return true;
+}
+
+// If path + filename exists, adds next available number suffix before the extension, starting from pSuffix
+bool vcModals_SequencialFilename(char* pBuffer, uint32_t bufLen, int* pSuffix)
+{
```
peterAdams1014: As per our discussion, this function could be simplified greatly.

src/vcModals.cpp
```diff
@@ -746,29 +749,154 @@ void vcModals_DrawUnsupportedFiles(vcState *pProgramState)
...
+    break;
+
+  case 2:
+    stbi_write_tga(pFilename, pProgramState->sceneResolution.x, pProgramState->sceneResolution.y, 4, pProgramState->screenshot.pPixels);
+    break;
+
+  case 3:
+    stbi_write_jpg(pFilename, pProgramState->sceneResolution.x, pProgramState->sceneResolution.y, 4, pProgramState->screenshot.pPixels, 100);
+    break;
+
```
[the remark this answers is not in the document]
samsinsane: I would assume the warning only applies to `enum`s due to the specific list of valid values.

src/vcModals.h
```diff
@@ -1,6 +1,8 @@
 #ifndef vcModals_h__
 #define vcModals_h__
 
+#include "udMath.h"
```
samsinsane: Why does this header need this include now?

builds/assets/lang/enAU.json
```diff
@@ -545,5 +545,20 @@
...
+  "screenshotResLabel": "Resolution",
+  "screenshotRes720p": "720p Resolution",
+  "screenshotRes1080p": "1080p Resolution",
+  "screenshotRes4K": "4K Resolution",
+  "screenshotFormatLabel": "Image Format",
+  "screenshotHideLabels": "Hide Labels",
+  "screenshotFilename": "Filename",
+  "screenshotSettings": "Setup Screenshot",
+  "screenshotTake": "Take Screenshot",
+  "screenshotDefaultName": "Screen Capture"
```
samsinsane: None of these are documented in the translation document.

src/main.cpp
DamianMadden: This block never opened the modal, but either way I've mostly changed to move settings into the settings dock and removed it from the modal, and I suppose I'll hook it up for show screenshots to bring up the classic image viewer

src/vcModals.cpp
```diff
@@ -746,29 +749,154 @@ void vcModals_DrawUnsupportedFiles(vcState *pProgramState)
...
+  }
+
+  udFree(pProgramState->screenshot.pPixels);
+  pProgramState->screenshot.taking = false;
+
+  return true;
+}
+
+// If path + filename exists, adds next available number suffix before the extension, starting from pSuffix
+bool vcModals_SequencialFilename(char* pBuffer, uint32_t bufLen, int* pSuffix)
```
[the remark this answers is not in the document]
DamianMadden: So the student becomes the teacher

src/main.cpp
DamianMadden: Printscreen was being intercepted by the system/not working, might be to do with this keyboard's prntscreen being fn+ insert

DamianMadden: Printscreen seems to get taken up by the OS, at least on this keyboard

src/vcModals.cpp
DamianMadden: This will be a little complex as we need the valid attached framebuffer to do so in this way, for earlier versions of opengl

src/vcState.h
```diff
@@ -174,6 +174,17 @@ struct vcState
   vcTranslationInfo languageInfo;
   bool showUI;
   vcDocks changeActiveDock;
+
+  struct screenshot
```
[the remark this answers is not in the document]
DamianMadden: the way it's layed out now there is a lot of ambiguity in some of the names

PR #175, opened 2019-11-21
Updated to latest VDK + Session Resume Support

OsiJr:


premake5.lua
```diff
@@ -211,7 +211,7 @@ solution "vaultClient"
 		filter { "kind:*App" }
 			targetextension ".js"
 		filter { "files:**.cpp" }
-			buildoptions { "-std=c++11" }
+			buildoptions { "-std=c++14" }
```
peterAdams1014: Why was this change necessary?

samsinsane: Yes, `udCallback` uses a C++14 feature, without this change it won't compile.

src/main.cpp
```diff
@@ -572,11 +575,8 @@ void vcMain_MainLoop(vcState *pProgramState)
     }
 
     // Ping the server every 30 seconds
-    if (udGetEpochSecsUTCf() > pProgramState->lastServerAttempt + 30.0)
-    {
-      pProgramState->lastServerAttempt = udGetEpochSecsUTCf();
+    if (pProgramState->hasContext && udGetEpochSecsUTCf() > pProgramState->sessionInfo.expiresTimestamp - 300)
```
samsinsane: Parenthesis around the `-` operation.

src/vcBPA.cpp
```diff
@@ -831,7 +846,11 @@ uint32_t vcBPA_GridGeneratorThread(void *pDataPtr)
 
   vcBPAGrid *pGrid = nullptr;
   int i = 0;
-  while (vcBPA_GetGrid(pData->pManifold, pData->pNewModel, pData->standardContent, &pGrid, false) && pData->running)
+
+  vdkPointCloudHeader header;
+  vdkPointCloud_GetHeader(pData->pNewModel, &header);
```
samsinsane: If this somehow fails, the header is going to be garbage data and the attributes will likely trigger a crash.

OsiJr: Yeah, sorry- knowing internal implementation broke this. I know that function can't fail in its current form and missed this. I'll add the checks now.

PR #181, opened 2019-11-25
Unit Conversion

Harshish5440:
[AB#220](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/220)

### DIFF
--- a/src/vcUnitConversion.cpp
+++ b/src/vcUnitConversion.cpp
@@ -4,167 +4,59 @@ double vcUnitConversion_ConvertDistance(double sourceValue, vcDistanceUnit sourc
 {
   if (sourceUnit == requiredUnit)
     return sourceValue;
-  double scalar = 1.0;
-  // Metric
-  if (sourceUnit == vcDistance_Metres)
-    scalar = 1.0;
-  else if (sourceUnit == vcDistance_Kilometres)
-    scalar = 1000.0;
-  else if (sourceUnit == vcDistance_Centimetres) 
-    scalar = 1 / 100.0;
-  else if (sourceUnit == vcDistance_Millimetres) 
-    scalar = 1 / 1000.0;
-  // Imperial
-  else if (sourceUnit == vcDistance_USSurveyFeet) 
-    scalar = 1 / 3.2808398950131225646;
-  else if (sourceUnit == vcDistance_USSurveyMiles)  
-    scalar = 1609.3440000000000509;
-  else if (sourceUnit == vcDistance_USSurveyInches)   
-    scalar = 1 / 39.370078740157481433;
-  // Nautical
-  else if (sourceUnit == vcDistance_NauticalMiles) 
-    scalar = 1852.0;
-  // Metric
-  if (requiredUnit == vcDistance_Metres)
-    scalar *= 1.0;
-  else if (requiredUnit == vcDistance_Kilometres)   
-    scalar *= 1 / 1000.0;
-  else if (requiredUnit == vcDistance_Centimetres)  
-    scalar *= 100.0;
-  else if (requiredUnit == vcDistance_Millimetres) 
-    scalar *= 1000.0;
-  // Imperial_US
-  else if (requiredUnit == vcDistance_USSurveyFeet) 
-    scalar *= 3.2808398950131225646;
-  else if (requiredUnit == vcDistance_USSurveyMiles)   
-    scalar *= 1 / 1609.3440000000000509;
-  else if (requiredUnit == vcDistance_USSurveyInches)
-    scalar *= 39.370078740157481433;
-  // Nautical
-  else if (requiredUnit == vcDistance_NauticalMiles)
-    scalar *= 1 / 1852.0;
-  return sourceValue * scalar;
+
+  const double metreTable[vcDistance_Count] = { 1.0, 1000.0, 0.01, 0.001, 1200.0 / 3937.0, 5280.0 * 1200.0 / 3937.0, 1.0 / 39.37, 1852.0 };
+
+  return sourceValue * metreTable[sourceUnit] / metreTable[requiredUnit];
 }
+
 double vcUnitConversion_ConvertArea(double sourceValue, vcAreaUnit sourceUnit, vcAreaUnit requiredUnit) 
 {
   if (sourceUnit == requiredUnit)
     return sourceValue;
-  double scalar = 1.0;
-  //Source Unit 
-  if (sourceUnit == vcAreaUnit_SquareMetres)
-    scalar = 1.0;
-  else if (sourceUnit == vcAreaUnit_SquareFoot)
-    scalar = 1 / 10.763910416709721928;
-  else if (sourceUnit == vcAreaUnit_SquareKilometers)
-    scalar = 1000000;
-  else if (sourceUnit == vcAreaUnit_SquareMiles)
-    scalar = 2589988.1000000000931;
-  else if (sourceUnit == vcAreaUnit_Acre)
-    scalar = 4046.8564000000001215;
-  else if (sourceUnit == vcAreaUnit_Hectare)
-    scalar = 1000;
-  //Required Unit
-  if (requiredUnit == vcAreaUnit_SquareMetres)
-    scalar *= 1.0;
-  else if (requiredUnit == vcAreaUnit_SquareFoot)
-    scalar *= 10.763910416709721928;
-  else if (requiredUnit == vcAreaUnit_SquareKilometers)
-    scalar *= 1 / 1000000.0;
-  else if (requiredUnit == vcAreaUnit_SquareMiles)
-    scalar *= 1 / 2589988.1000000000931;
-  else if (requiredUnit == vcAreaUnit_Acre)
-    scalar *= 1 / 4046.8564000000001215;
-  else if (requiredUnit == vcAreaUnit_Hectare)
-    scalar *= 1 / 10000.0;
-  return sourceValue * scalar;
+
+  const double sqmTable[vcArea_Count] = { 1.0, 1000000.0, 1000.0, 1200.0 / 3937.0 * 1200.0 / 3937.0, 5280.0 * 1200.0 / 3937.0 * 5280.0 * 1200.0 / 3937.0, 5280.0 * 1200.0 / 3937.0 * 5280.0 * 1200.0 / 3937.0 * 1.0 / 640.0 };
+
+  return sourceValue * sqmTable[sourceUnit] / sqmTable[requiredUnit];
 }
 double vcUnitConversion_ConvertVolume(double sourceValue, vcVolumeUnit sourceUnit, vcVolumeUnit requiredUnit)
 {
   if (sourceUnit == requiredUnit)
     return sourceValue;
-  double scalar = 1.0;
-  //Source Unit
-  if (sourceUnit == vcVolumeUnit_CubicMeter)
-    scalar = 1.0;
-  else if (sourceUnit == vcVolumeUnit_USSGallons)
-    scalar = 1 / 264.17203728418462560;
-  else if (sourceUnit == vcVolumeUnit_USSQuart)
-    scalar = 1 / 1056.6882607957347;
-  else if (sourceUnit == vcVolumeUnit_Liter)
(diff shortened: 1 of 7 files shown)


src/vcUnitConversion.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+double vcUnitConversion_ConvertDistance(double sourceValue, vcDistanceUnit sourceUnit, vcDistanceUnit requiredUnit)
+{
+  double scalar = 1.0;
+
+  // Metric
+  if (sourceUnit == vcDistance_Metres)
+    scalar = 1.0;
+  else if (sourceUnit == vcDistance_Kilometres) // If source unit is in Meter than Multiply by 1000 to get kilometer.
+    scalar = 1000;
```
samsinsane: This should end with `.0`, similarly with all the other doubles in this file that are whole numbers.

src/vcUnitConversion.cpp
```diff
@@ -0,0 +1,174 @@
...
+  double scalar = 1.0;
+
+  if (sourceUnit == vcAreaUnit_SquareMetres)
+    scalar = 1.0;
+  else if (sourceUnit == vcAreaUnit_SquareFoot)
+    scalar = 1 / 10.76;
+  else if (sourceUnit == vcAreaUnit_SquareKilometers)
+    scalar = 1000000;
+  else if (sourceUnit == vcAreaUnit_SquareMiles)
+    scalar = 2589988.10;
```
samsinsane: The trailing `0` here is unnecessary.

src/vcUnitConversion.cpp
```diff
@@ -0,0 +1,174 @@
...
+    scalar = 3.2;
+  return sourceValue * scalar;
+}
+
+double vcUnitConversion_ConvertTemperature(double sourceValue, vcTemperatureUnit sourceUnit, vcTemperatureUnit requiredUnit)
+{
+  double scalar = 1.0;
+
+  if (sourceUnit == vcTemperature_Celcius)
+    scalar = 1.00;
```
samsinsane: The extra `0` here is unnecessary.

src/vcUnitConversion.cpp
```diff
@@ -0,0 +1,174 @@
...
+  else if (sourceUnit == vcTemperature_Farenheit)
+    scalar = (sourceValue - 32) * (5 / 9);
+  if (requiredUnit == vcTemperature_Celcius)
+    scalar = 1.00;
+  else if (requiredUnit == vcTemperature_Kelvin)
+    scalar = sourceValue + 273.15;
+  else if (requiredUnit == vcTemperature_Farenheit)
+    scalar = (sourceValue * (5 / 9)) + 32;
+
+  return sourceValue = scalar;
```
samsinsane: This is a strange operation, this means when `requireUnit` is `vcTemperature_Celcius` this function will always return `1.0`.

src/vcUnitConversion.cpp
```diff
@@ -0,0 +1,174 @@
...
+  double scalar = 1.0;
+
+  if (sourceUnit == vcTemperature_Celcius)
+    scalar = 1.00;
+  else if (sourceUnit == vcTemperature_Kelvin)
+    scalar = sourceValue - 273.15;
+  else if (sourceUnit == vcTemperature_Farenheit)
+    scalar = (sourceValue - 32) * (5 / 9);
+  if (requiredUnit == vcTemperature_Celcius)
+    scalar = 1.00;
```
samsinsane: The extra `0` here is also unnecessary.

src/vcUnitConversion.cpp
```diff
@@ -0,0 +1,174 @@
...
+  else if (requiredUnit == vcDistance_Kilometres)   //If require unit is in Meter than divide by 1000 to get kilometer.
+    scalar = 1 / 1000;
+  else if (requiredUnit == vcDistance_Centimetres)  //Here one meter is equal to 100 Centimeter.
+    scalar = 100;
+  else if (requiredUnit == vcDistance_Millimetres) // 1 Meter = 10000 millimeter.
+    scalar = 1000;
+
+  // Imperial_US
+  else if (requiredUnit == vcDistance_USSurveyFeet) //  1 meter = 3.28 Feet
+    scalar = 3.28;
```
peterAdams1014: Probably need more precision than this. 1 meter equals `3.28084` feet.

src/vcUnitConversion.cpp
```diff
@@ -0,0 +1,174 @@
...
+  else if (requiredUnit == vcDistance_USSurveyMiles)   //1 mile = 1609.34 meter
+    scalar = 1609.34;
+  else if (requiredUnit == vcDistance_USSurveyInches)
+    scalar = 39.37;
+
+  // Nautical
+  else if (sourceUnit == vcDistance_NauticalMiles)
```
peterAdams1014: Did you intend `requiredUnit == vcDistance_NauticalMiles` here?

src/vcUnitConversion.cpp
```diff
@@ -0,0 +1,174 @@
...
+  else if (requiredUnit == vcVolumeUnit_CubicInch)
+    scalar = 61023.74;
+  else if (requiredUnit == vcVolumeUnit_CubicFoot)
+    scalar = 35.31;
+  else if (requiredUnit == vcVolumeUnit_MegaLiter)
+    scalar = 1 / 1000;
+  else if (requiredUnit == vcVolumeUnit_CubicYard)
+    scalar = 1 / 1.3;
+
+  return sourceValue;
```
peterAdams1014: Missing the actual conversion calculation here. `return sourceValue * scalar;`.

src/vcUnitConversion.cpp
```diff
@@ -0,0 +1,174 @@
...
+  else if (sourceUnit == vcTemperature_Kelvin)
+    scalar = sourceValue - 273.15;
+  else if (sourceUnit == vcTemperature_Farenheit)
+    scalar = (sourceValue - 32) * (5 / 9);
+  if (requiredUnit == vcTemperature_Celcius)
+    scalar = 1.00;
+  else if (requiredUnit == vcTemperature_Kelvin)
+    scalar = sourceValue + 273.15;
+  else if (requiredUnit == vcTemperature_Farenheit)
+    scalar = (sourceValue * (5 / 9)) + 32;
```
peterAdams1014: This line has an issue.  In strongly typed languages (c++), `5` (integer) and `5.0` (double) are different types. 
`5 / 9` will truncate to `0`, because integers are whole numbers. Whereas `5.0 / 9.0` will equal `0.5555556`. 

This is a simple fix, if you simply change the line to only have doubles: `scalar = (sourceValue * (5.0 / 9.0)) + 32.0;`

src/vcUnitConversion.cpp
Harshish5440: Yup, its nautical miles.

DamianMadden:
> I'd have an enum and a 2d table like vcMeasure_cM, vcMeasure_M, vcMeasure_kM
> then
> vc...cm vc...m vc...km
> vc...cm 1 100 100,000
> vc...m .01 1 1000
> 
> so row is source unit, column is destination unit, then lookup is table[vc...m][vc...km]
> 
> You can still wrap it in a function if you want

To clarify my position here, it's my expectation that in the configuration I've described the solution can be looked up by 2 known offsets into an array, so it should be more efficient notwithstanding unexpected compiler optimizations to dereference a value at x+y+z than to iterate through the series of possible units and conditionally do some processing based on that.

Actually I've just noticed additional processing will be needed for Fahrenheit, and other non-linear conversions, still though i think this is likely a better option with a handful of edge cases handled in the function

src/vcUnitConversion.cpp
```diff
@@ -56,26 +61,27 @@ double vcUnitConversion_ConvertArea(double sourceValue, vcAreaUnit sourceUnit, v
   if (sourceUnit == vcAreaUnit_SquareMetres)
     scalar = 1.0;
   else if (sourceUnit == vcAreaUnit_SquareFoot)
-    scalar = 1 / 10.76;
+
```
peterAdams1014: Unnecessary newline.

src/vcUnitConversion.cpp
peterAdams1014: `return sourceValue * scalar;`

src/vcUnitConversion.cpp
```diff
@@ -37,16 +38,20 @@ double vcUnitConversion_ConvertDistance(double sourceValue, vcDistanceUnit sourc
...
+    scalar *= 39.370078740157481433925568126142024993896484375;
 
   // Nautical
-  else if (sourceUnit == vcDistance_NauticalMiles)
-    scalar *= 1852.0;
+  else if (requiredUnit == vcDistance_NauticalMiles)
+    scalar *= 1 / 1852.0;
+
   return sourceValue * scalar;
+
```
peterAdams1014: Unnecessary newlines.

src/vcUnitConversion.h
```diff
@@ -0,0 +1,78 @@
...
+  // Imperial
+  vcDistance_USSurveyFeet,
+  vcDistance_USSurveyMiles,
+  vcDistance_USSurveyInches,
+
+  // Nautical
+  vcDistance_NauticalMiles,
+
+  vcDistance_Count
+};
```
DamianMadden: You should try to consider data structures to help you rather than pure logic, but you can determine what you need logically, so for each input unit you require a different output multiplier value, multiplied by the input value. This suits an N<sup>2</sup> table of multiplier values, OR a conversion to and from a standard unit for each other unit of a set... the first requires N<sup>2</sup> values but lookup is a single calculation, the standard intermediate unit style is N*2 values, but lookup can be as bad as N*2, additionally doing an extra calculation. I'd suggest as well you can add `if (source == required) return input value;`

OsiJr:

src/vcUnitConversion.cpp
```diff
@@ -0,0 +1,65 @@
+#include "vcUnitConversion.h"
+
+double vcUnitConversion_ConvertDistance(double sourceValue, vcDistanceUnit sourceUnit, vcDistanceUnit requiredUnit)
+{
+  if (sourceUnit == requiredUnit)
+    return sourceValue;
+
+  const double metreTable[vcDistance_Count] = { 1.0, 1000.0, 0.01, 0.001, 1200.0 / 3937.0, 5280.0 * 1200.0 / 3937.0, 1.0 / 39.37, 1852.0 };
```
DamianMadden: This is much better, I was going to say it would still be an improvement to pre-calculate these values into a table, but probably not as the lookup constitutes an additional multiplication anyway

vcTesting/src/vcUnitConversionTests.cpp
```diff
@@ -0,0 +1,53 @@
...
+  EXPECT_DOUBLE_EQ(223.69362920544023154, vcUnitConversion_ConvertSpeed(100, vcSpeed_MetresPerSecond, vcSpeed_USSurveyMilesPerHour));
+  EXPECT_DOUBLE_EQ(194.38444924406047675, vcUnitConversion_ConvertSpeed(100, vcSpeed_MetresPerSecond, vcSpeed_NauticalMilesPerHour));
+  EXPECT_DOUBLE_EQ(328.08398950131231686100363, vcUnitConversion_ConvertSpeed(100, vcSpeed_MetresPerSecond, vcSpeed_FeetPerSecond));
+  EXPECT_DOUBLE_EQ(340.29, vcUnitConversion_ConvertSpeed(1.0, vcSpeed_Mach, vcSpeed_MetresPerSecond));
+}
+
+TEST(UnitConversion, Temperature)
+{
+  EXPECT_DOUBLE_EQ(212, vcUnitConversion_ConvertTemperature(100.0, vcTemperature_Celcius, vcTemperature_Farenheit));
+  EXPECT_DOUBLE_EQ(373.15, vcUnitConversion_ConvertTemperature(100.0, vcTemperature_Celcius, vcTemperature_Kelvin));
```
peterAdams1014: Based on the dubious implementation of the `vcUnitConversion_ConvertTemperature()`, could you add more tests please.

PR #184, opened 2019-11-27
Prevents possible double allocation

DamianMadden:


src/scene/vcModel.cpp
```diff
@@ -44,15 +44,7 @@ void vcModel_LoadMetadata(vcState *pProgramState, vcModel *pModel)
...
-      if (udGeoZone_SetFromSRID(pModel->m_pPreferredProjection, srid) != udR_Success)
-        udFree(pModel->m_pPreferredProjection);
-    }
-
-    if (pModel->m_pPreferredProjection == nullptr && pWKT != nullptr)
-    {
-      pModel->m_pPreferredProjection = udAllocType(udGeoZone, 1, udAF_Zero);
-
-      if (udGeoZone_SetFromWKT(pModel->m_pPreferredProjection, pWKT) != udR_Success)
+      if (udGeoZone_SetFromSRID(pModel->m_pPreferredProjection, srid) != udR_Success && (pWKT != nullptr && udGeoZone_SetFromWKT(pModel->m_pPreferredProjection, pWKT) != udR_Success))
```
OsiJr: Its possible to not his this block if `pSRID` is null; which is true on older models

OsiJr: When `SRID:0` is non-geolocated space (which it partially is now but will definitely be the case after AB#479 and required fixes for non-geolocated projects); The `SetFromSRID` needs to be guarded with `pSRID != nullptr &&` which gets you to just a rolled up version of what you started with but with extra branches and longer lines.

PR #186, opened 2019-11-27
Fixed a potential udMempDup() with size 0.

peterAdams1014:
[AB#114](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/114)

### DIFF
--- a/src/vcSceneLayer.cpp
+++ b/src/vcSceneLayer.cpp
@@ -596,10 +596,13 @@ udResult vcSceneLayer_LoadTextureData(const vcSceneLayer *pSceneLayer, vcSceneLa
         continue;
 
       pPixelData = (uint8_t*)stbi_load_from_memory((stbi_uc*)pFileData, (int)fileLen, (int*)&width, (int*)&height, (int*)&channelCount, 4);
-
-      pNode->pTextureData[i].width = width;
-      pNode->pTextureData[i].height = height;
-      pNode->pTextureData[i].pData = (uint8_t*)udMemDup(pPixelData, sizeof(uint32_t)*width*height, 0, udAF_None);
+      
+      if (pPixelData != nullptr)
+      {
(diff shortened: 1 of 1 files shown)


src/vcSceneLayer.cpp
```diff
@@ -596,10 +596,13 @@ udResult vcSceneLayer_LoadTextureData(const vcSceneLayer *pSceneLayer, vcSceneLa
         continue;
```
DamianMadden: pPathBuffer doesn't get allocated, can udSprintf work like this?

src/vcSceneLayer.cpp
```diff
@@ -596,10 +596,13 @@ udResult vcSceneLayer_LoadTextureData(const vcSceneLayer *pSceneLayer, vcSceneLa
...
+      
+      if (pPixelData != nullptr)
+      {
+        pNode->pTextureData[i].width = width;
+        pNode->pTextureData[i].height = height;
+        pNode->pTextureData[i].pData = (uint8_t *)udMemDup(pPixelData, sizeof(uint32_t) * width * height, 0, udAF_None);
+      }
 
       udFree(pFileData);
       stbi_image_free(pPixelData);
```
DamianMadden: Why do this memdup at all?

src/vcSceneLayer.cpp
samsinsane: `pPathBuffer` is allocated via `udSprintf`, when you pass in a `char **` as the first parameter it will allocate a buffer big enough for you.

src/vcSceneLayer.cpp
samsinsane: The memdup avoids needing to use `stbi_image_free` to free `pData` later, this means that the data could come from anything instead of just an STBI API.

PR #187, opened 2019-11-28
Screenshot

DamianMadden:
bound to F8 Key because printscreen key is taken up by the OS


src/vcSettings.cpp
```diff
@@ -91,6 +91,25 @@ bool SDL_filewrite(const char* filename, const char *pStr, size_t bytes)
...
+    return nullptr;
+
+  udFilename temp(pName);
+  const char* pExtension = temp.GetExt();
+
+  int i = 1;
+  while (udFileExists(temp.GetPath()) == udR_Success)
+  {
+    temp.SetFromFullPath(pName);
+    temp.SetExtension(udTempStr("%d%s", i++, pExtension));
```
samsinsane: `i++` should be on another line.

src/vcState.h
```diff
@@ -166,7 +166,8 @@ struct vcState
     int height;
 
     vcImageType imageType;
-  } image;
+  } image[2];
```
samsinsane: Did this need to be an array? Could it have been two different variables instead? `} image, screenshot` or something?

src/main.cpp
```diff
@@ -55,6 +55,11 @@
 #include "udStringUtil.h"
 #include "udUUID.h"
 
+#include "udPlatformUtil.h"
+#include "stb_image.h"
+#define STB_IMAGE_WRITE_IMPLEMENTATION
+#include "stb_image_write.h"
```
OsiJr: `vcTexture` is a better place for this

src/main.cpp
```diff
@@ -188,6 +193,77 @@ void vcMain_PresentationMode(vcState *pProgramState)
     pProgramState->lastEventTime = udGetEpochSecsUTCd();
 }
 
+bool vcMain_TakeScreenshot(vcState* pProgramState, const char* pFilename = nullptr)
```
OsiJr: Most of this function should be `vcTexture_SaveTexture(vcTexture *pTexture, const char *pFilename);` or `vcFrameBuffer_Save...`

Where do you prefer, @peterAdams1014?

src/vcSettings.cpp
```diff
@@ -428,6 +447,14 @@ bool vcSettings_Load(vcSettings *pSettings, bool forceReset /*= false*/, vcSetti
     udFree(pFileContents);
   }
 
+  if (group == vcSC_Screenshot || group == vcSC_All)
+  {
+    pSettings->screenshot.hideLabels = data.Get("screenshot.hideLabels").AsBool(false);
+    pSettings->screenshot.format = (vcImageFormats)data.Get("screenshot.format").AsInt(vcIF_PNG);
+    pSettings->screenshot.res = (vcImageResolutions)data.Get("screenshot.res").AsInt(vcIR_720p);
```
OsiJr: These need to be a strings so the enum can change in future

src/vcSettings.h
```diff
@@ -95,6 +97,21 @@ enum vcTileRendererFlags
   vcTRF_OnlyRequestVisibleTiles = 0x1,
 };
 
+enum vcImageFormats
+{
+  vcIF_PNG,
+  vcIF_BMP,
+  vcIF_TGA,
+  vcIF_JPG
```
OsiJr: Just because we _can_ offer these doesn't mean we should. Just having PNG and JPG is enough.

src/vcSettings.h
```diff
@@ -95,6 +97,21 @@ enum vcTileRendererFlags
...
+};
+
+enum vcImageResolutions
```
OsiJr: This name is not suitable. `vcScreenshotOutputResolution` is better. There should also be a 'current scene window size' option

src/vcSettings.h
```diff
@@ -259,6 +276,19 @@ struct vcSettings
     char license[vcMetadataMaxLength];
   } convertdefaults;
 
+  struct
+  {
+    bool taking;
+    bool resize;
```
OsiJr: `resizeRequired` provides a lot more information

src/vcSettingsUI.h
```diff
@@ -3,6 +3,8 @@
 
 struct vcState;
 
+static const char* ScreenshotExportFormats[4] = { ".PNG", ".BMP", ".TGA", ".JPG" };
```
OsiJr: These screenshot arrays are getting spread out over the whole program, bringing them together is going to help later

src/vcState.h
OsiJr: Separate variables will clean up a lot of issues.

src/vcModals.cpp
```diff
@@ -5,6 +5,7 @@
 #include "vcThirdPartyLicenses.h"
 #include "vcPOI.h"
 #include "gl/vcTexture.h"
+#include "gl/vcFramebuffer.h"
```
OsiJr: This should not be required here- viewing the framebuffer makes no sense; just save the image and load it back in?

That also means the array in `vcState` can be removed?

src/main.cpp
DamianMadden: I've moved it to vcTexture, as the framebuffer isn't even entirely necessary except on opengl

OsiJr:

src/vcModals.cpp
```diff
@@ -766,24 +766,25 @@ void vcModals_DrawImageViewer(vcState *pProgramState)
   if (pProgramState->openModals & (1 << vcMT_ImageViewer))
     ImGui::OpenPopup(vcString::Get("sceneImageViewerTitle"));
 
-  // Use 75% of the window
-  int maxX, maxY;
-  SDL_GetWindowSize(pProgramState->pWindow, &maxX, &maxY);
-  ImGui::SetNextWindowSize(ImVec2(maxX * 0.75f, maxY * 0.75f), ImGuiCond_Always);
-
+  ImGui::SetNextWindowSizeConstraints(ImVec2(50.f, 50.f), ImVec2((float)pProgramState->settings.window.width, (float)pProgramState->settings.window.height));
+  ImGui::SetNextWindowSize(ImVec2((float)pProgramState->image.width + 25, (float)pProgramState->image.height + 50), ImGuiCond_Appearing);
```
samsinsane: This is very different to what was there before, this uses the entire screen instead of just 75%, why? Also, why is there a discrepancy between the magic numbers? Minimum size is 50x50, but you add 25x50?

src/vcSettings.h
```diff
@@ -95,6 +99,24 @@ enum vcTileRendererFlags
...
+  vcIR_SceneSize
+};
+
+enum class vcImageFormats
+{
+  vcIF_PNG,
+  vcIF_JPG
+};
+
+static udUInt2 ScreenshotResolutions[] = { {1280, 720}, {1920, 1080}, {4096, 2160}, {1920, 1080} };
```
samsinsane: Why is the default Scene Size `1920, 1080` when the default in the settings is actually `1280, 720`?

src/vcSettings.h
```diff
@@ -95,6 +99,24 @@ enum vcTileRendererFlags
...
+static const char *ScreenshotResolutionStrings[4] = { "720", "1080", "4k", "Scene" };
+static const char *ScreenshotExportFormats[2] = { ".PNG", ".JPG" };
```
samsinsane: Why uppercase? I don't believe this was done for the map tile code?

PR #188, opened 2019-11-29
Headless tile server requests

DamianMadden:
I have avoided using the udFOF_ flag in vcWebFile to achieve this because this way avoids an unnecessary allocation for the buffer used in vcWebFile_SeekRead, still added the flag for completeness though

### DIFF
--- a/src/rendering/vcTileRenderer.cpp
+++ b/src/rendering/vcTileRenderer.cpp
@@ -8,8 +8,6 @@
 #include "gl/vcShader.h"
 #include "gl/vcMesh.h"
 
-#include "vdkWeb.h"
-
 #include "udThread.h"
 #include "udFile.h"
 #include "udPlatformUtil.h"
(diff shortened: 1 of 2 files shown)


OsiJr:
Customers have tile servers locally, on network drives and on non-http protocols that are all supported currently via the udFileHandlers.

src/vcWebFile.cpp
```diff
@@ -64,20 +64,23 @@ udResult vcWebFile_Open(udFile **ppFile, const char *pFilename, udFileOpenFlags
...
-  else if (responseCode == 503)
-    UD_ERROR_SET(udR_Pending);
-  else if (responseCode >= 500 && responseCode <= 599)
-    UD_ERROR_SET(udR_ServerError);
-  else if (responseCode < 200 || responseCode >= 300)
-    UD_ERROR_SET(udR_OpenFailure);
-
-  pFile->totalBytes = dataLength;
-  pFile->fileLength = dataLength;
+  if (!(flags & udFOF_FastOpen))
```
samsinsane: I feel like we've looked into this before and it created problems, does this just work? How much testing have you done on this? I think the tile server was really one of the only places this was used.

OsiJr:
@samsinsane, removing the head request is supposed to be optional + if you could be confident of the buffer size (like with tiles or config files).

The actual performance improvement for tiles is quite negligible; on a standard scene with Australian internet it takes around 30 seconds to fully resolve the map. Removing the head request is only about ~1-3 seconds improvement. All the other work @peterAdams1014 has done offers a far superior improvement than the HEAD changes do.

samsinsane:
@OsiJr after more thought, it might be a good idea for the Emscripten port. In the short term, it reduces the number of web requests we do which is one of our least stable areas, but I expect long term this to have no impact as the web requests code paths are fixed.

src/vcWebFile.cpp
DamianMadden: Yes, with the change to make vcTileRenderer use this it does appear to work

DamianMadden: Not much testing though, only really ran it for like 10 minutes

src/rendering/vcTileRenderer.cpp
```diff
@@ -223,15 +223,17 @@ uint32_t vcTileRenderer_LoadThread(void *pThreadData)
...
       udResult result = udR_Failure_;
-      void *pFileData = nullptr;
-      int64_t fileLen = -1;
+      uint32_t amountRead = 0;
       int width = 0;
       int height = 0;
       int channelCount = 0;
       uint8_t *pData = nullptr;
+      udFile *pFile = nullptr;
+      char buffer[20000] = {};
```
OsiJr: Some tiles might be bigger than this especially if they are higher res than 256² or full colour satellite- they could be much larger.

@samsinsane and I had an offline discussion that the best way to do this might be for file handlers to be a bit more flexible.

After an additional discussion with @dpev, he suggested the simplest/cleanest option would be if file-handlers could provide an (optional) load file function pointer; current behaviour continues if that is nullptr, otherwise it uses that function to load the whole file.

OsiJr:

OsiJr:
Closing due to offline discussions about how to handle this.

PR #195, opened 2019-12-05
Select ImGui explorer

DamianMadden:
Makes clicks onto scene item imgui elements select the item
Resolves [AB#544](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/544)

### DIFF
--- a/src/scene/vcFolder.cpp
+++ b/src/scene/vcFolder.cpp
@@ -342,7 +342,12 @@ void vcFolder::HandleImGui(vcState *pProgramState, size_t *pItemID)
         ImGui::Indent();
         ImGui::PushID(udTempStr("SXIExpanded%zu", *pItemID));
 
+        ImGui::BeginGroup();
         pSceneItem->HandleImGui(pProgramState, pItemID);
+        ImGui::EndGroup();
+
+        if (ImGui::IsItemClicked(0))
+          vcProject_SelectItem(pProgramState, m_pNode, pNode);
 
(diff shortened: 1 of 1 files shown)


src/scene/vcFolder.cpp
```diff
@@ -342,7 +342,12 @@ void vcFolder::HandleImGui(vcState *pProgramState, size_t *pItemID)
         ImGui::Indent();
         ImGui::PushID(udTempStr("SXIExpanded%zu", *pItemID));
 
+        ImGui::BeginGroup();
         pSceneItem->HandleImGui(pProgramState, pItemID);
+        ImGui::EndGroup();
+
+        if (ImGui::IsItemClicked(0))
+          vcProject_SelectItem(pProgramState, m_pNode, pNode);
```
OsiJr: I would probably recommend not doing it this way.

The majority the UI does not need to do the selection and this behaviour is not consistent with how other applications behave. The bug can be better fixed by showing the elements when the selected item is `-1`. Changing the speed of an attached model shouldn't select the item for example.

DamianMadden: So, you're suggesting just closing the PR and the AB item?

DamianMadden: or you want selection to open the treenode for an items handleimgui?

OsiJr:
Going to close this and close the issue as well.

PR #196, opened 2019-12-05
Makes missed tolltips reflect current bindings

DamianMadden:
Resolves [AB#686](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/686)

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2244,7 +2244,10 @@ void vcRenderWindow(vcState *pProgramState)
     {
       if (ImGui::Begin(udTempStr("%s###sceneExplorerDock", vcString::Get("sceneExplorerTitle")), &pProgramState->settings.window.windowsOpen[vcDocks_SceneExplorer]))
       {
-        if (vcMenuBarButton(pProgramState->pUITexture, vcString::Get("sceneExplorerAddUDS"), vcString::Get("sceneExplorerAddUDSKey"), vcMBBI_AddPointCloud, vcMBBG_FirstItem) || vcHotkey::IsPressed(vcB_AddUDS))
+        char buffer[50] = {};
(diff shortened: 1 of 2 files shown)


src/main.cpp
```diff
@@ -2244,7 +2244,10 @@ void vcRenderWindow(vcState *pProgramState)
     {
       if (ImGui::Begin(udTempStr("%s###sceneExplorerDock", vcString::Get("sceneExplorerTitle")), &pProgramState->settings.window.windowsOpen[vcDocks_SceneExplorer]))
       {
-        if (vcMenuBarButton(pProgramState->pUITexture, vcString::Get("sceneExplorerAddUDS"), vcString::Get("sceneExplorerAddUDSKey"), vcMBBI_AddPointCloud, vcMBBG_FirstItem) || vcHotkey::IsPressed(vcB_AddUDS))
+        char buffer[50] = {};
+        vcHotkey::GetKeyName(vcB_AddUDS, buffer, udLengthOf(buffer));
```
samsinsane: It might be better to add a template overload similar to how `udLengthOf` works so that you can just call `vcHotkey::GetKeyName(vcB_AddUDS, buffer)`. This reduces the chances of the two parameters getting out of sync which has happened before.

src/vcHotkey.h
```diff
@@ -41,6 +41,13 @@ namespace vcHotkey
   bool IsDown(vcBind key);
   bool IsPressed(vcBind key);
   void GetKeyName(vcBind key, char *pBuffer, uint32_t bufferLen);
+
+  template <size_t N>
+  void GetKeyName(vcBind key, char(&buffer)[N])
```
samsinsane: This should probably marked as `inline`, but also, you can just put this on a single line - see `udStringUtil.h` in udCore for examples of how this is done.

PR #201, opened 2019-12-05
Write failure reset

DamianMadden:
Resolves [AB#337](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/337)

### DIFF
--- a/src/vcConvert.cpp
+++ b/src/vcConvert.cpp
@@ -398,7 +398,7 @@ void vcConvert_ShowUI(vcState *pProgramState)
   }
   else
   {
-    if ((pSelectedJob->status == vcCQS_Preparing || pSelectedJob->status == vcCQS_Cancelled))
+    if (pSelectedJob->status == vcCQS_Preparing || pSelectedJob->status == vcCQS_Cancelled || pSelectedJob->status == vcCQS_WriteFailed || pSelectedJob->status == vcCQS_ParseFailed || pSelectedJob->status == vcCQS_ImageParseFailed || pSelectedJob->status == vcCQS_Failed)
     {
       if (ImGui::Button(vcString::Get("convertAddFile"), ImVec2(200, 40)))
         vcModals_OpenModal(pProgramState, vcMT_ConvertAdd);


src/vcConvert.cpp
```diff
@@ -421,7 +421,7 @@ void vcConvert_ShowUI(vcState *pProgramState)
         ImGui::Button(localizationBuffer, ImVec2(-1, 40));
       }
     }
-    else if (pSelectedJob->status == vcCQS_Completed)
+    else if (pSelectedJob->status == vcCQS_Completed || pSelectedJob->status == vcCQS_WriteFailed)
```
peterAdams1014: This doesn't seem right...On a failure, it opens up the UI to "add to scene" and what not?

Should `vcCQS_WriteFailed` be with `vcCQS_Preparing` and `vcCQS_Cancelled` above?

src/vcConvert.cpp
```diff
@@ -398,7 +398,7 @@ void vcConvert_ShowUI(vcState *pProgramState)
   }
   else
   {
-    if ((pSelectedJob->status == vcCQS_Preparing || pSelectedJob->status == vcCQS_Cancelled))
+    if (pSelectedJob->status == vcCQS_Preparing || (pSelectedJob->status >= vcCQS_Cancelled && pSelectedJob->status <= vcCQS_Failed))
```
samsinsane: I'm not a huge fan of this range check, it goes from a non-error condition to an error condition. I can only see problems arising from this kind of assumption.

OsiJr: Also can cause non-obvious errors if you modify the enum to add another value in that range

src/vcConvert.cpp
```diff
@@ -398,7 +398,7 @@ void vcConvert_ShowUI(vcState *pProgramState)
   }
   else
   {
-    if ((pSelectedJob->status == vcCQS_Preparing || pSelectedJob->status == vcCQS_Cancelled))
+    if (pSelectedJob->status == vcCQS_Preparing || pSelectedJob->status >= vcCQS_Cancelled || pSelectedJob->status == vcCQS_WriteFailed || pSelectedJob->status == vcCQS_ParseFailed || pSelectedJob->status == vcCQS_ImageParseFailed || pSelectedJob->status == vcCQS_Failed)
```
samsinsane: I assume you meant to undo the `>=` here?

src/vcProject.cpp
```diff
@@ -158,10 +158,15 @@ void vcProject_Save(vcState *pProgramState, const char *pPath, bool allowOverrid
       projectError.source = vcES_ProjectChange;
       projectError.pData = udStrdup(exportFilename.GetFilenameWithExt());
 
-      if (udFile_Save(exportFilename.GetPath(), (void *)pOutput, udStrlen(pOutput)) == udR_Success)
-        projectError.resultCode = udR_Success;
-      else
-        projectError.resultCode = udR_WriteFailure;
+      udFile *pOutFile;
+      if (udFile_Open(&pOutFile, exportFilename.GetPath(), udFOF_Create) == udR_Success)
```
samsinsane: You don't use the `pOutFile` that's created here, why is this done?

OsiJr:

src/vcProject.cpp
DamianMadden: I really don't remember doing this, and in my git log it appears out of order with other changes, just gonna revert it I suppose

PR #204, opened 2019-12-09
Fixed a bug when resetting a SLPK conversion

peterAdams1014:
[AB#699](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/699)

### DIFF
--- a/src/vcSceneLayer_Internal.h
+++ b/src/vcSceneLayer_Internal.h
@@ -48,7 +48,7 @@ struct vcSceneLayerNode
 
   int level;
   char *pID;
-  const char *pURL;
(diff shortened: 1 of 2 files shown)


src/vcSceneLayer_Internal.h
```diff
@@ -48,7 +48,7 @@ struct vcSceneLayerNode
 
   int level;
   char *pID;
-  const char *pURL;
+  const char *pURL; // Note: the parent of this node is responsible for this memory.
```
samsinsane: Can you clarify why this is the case?

peterAdams1014: Here? Or extend the comment?

PR #206, opened 2019-12-09
Key Modifier Fix

peterAdams1014:
[AB#713](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/713)

### DIFF
--- a/src/vcHotkey.cpp
+++ b/src/vcHotkey.cpp
@@ -77,18 +77,7 @@ namespace vcHotkey
     if (target != -1)
       return false;
 
-    ImGuiIO io = ImGui::GetIO();
-
-    if (((keyNum & vcMOD_Shift) == vcMOD_Shift) != io.KeyShift)
-      return false;
-    if (((keyNum & vcMOD_Ctrl) == vcMOD_Ctrl) != io.KeyCtrl)
-      return false;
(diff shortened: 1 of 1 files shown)


src/vcHotkey.cpp
```diff
@@ -77,18 +77,7 @@ namespace vcHotkey
...
-      return false;
-    if (((keyNum & vcMOD_Ctrl) == vcMOD_Ctrl) != io.KeyCtrl)
-      return false;
-    if (((keyNum & vcMOD_Alt) == vcMOD_Alt) != io.KeyAlt)
-      return false;
-    if (((keyNum & vcMOD_Super) == vcMOD_Super) != io.KeySuper)
-      return false;
-
-    return io.KeysDown[keyNum];
+    return ImGui::GetIO().KeysDown[keyNum];
```
OsiJr: This will mean that `Ctrl + A` and `Shift + A` as different bindings will both trigger.

OsiJr: I've had a discussion with @DamianMadden, we should add a warning on the modal that movement keys are handled differently and that Shift and Ctrl modify the movement keys.

OsiJr: This is tracked as AB#738

PR #209, opened 2019-12-10
Tile Server supports Google maps

zengweicheng666:
https://mt1.google.cn/maps/vt?lyrs=s&hl=zh-CN&gl=CN&x=0&y=0&z=0

### DIFF
--- a/src/rendering/vcTileRenderer.cpp
+++ b/src/rendering/vcTileRenderer.cpp
@@ -209,22 +209,23 @@ uint32_t vcTileRenderer_LoadThread(void *pThreadData)
       char localFileName[vcMaxPathLength];
       char serverAddress[vcMaxPathLength];
 
-	  switch (pRenderer->pSettings->maptiles.serverFlag)
-	  {
-	  case vcMTSF_Unknown:
-		  break;
-	  case vcMTSF_Slippy:
-		  udSprintf(localFileName, "%s/%s/%d/%d/%d.%s", pRenderer->pSettings->cacheAssetPath, udUUID_GetAsString(pRenderer->pSettings->maptiles.tileServerAddressUUID), pBestNode->slippyPosition.z, pBestNode->slippyPosition.x, pBestNode->slippyPosition.y, pRenderer->pSettings->maptiles.tileServerExtension);
-		  udSprintf(serverAddress, "%s/%d/%d/%d.%s", pRenderer->pSettings->maptiles.tileServerAddress, pBestNode->slippyPosition.z, pBestNode->slippyPosition.x, pBestNode->slippyPosition.y, pRenderer->pSettings->maptiles.tileServerExtension);
-		  break;
-	  case vcMTSF_Google:
-		  udSprintf(localFileName, "%s?lyrs=%s&x=%d&y=%d&z=%d", pRenderer->pSettings->cacheAssetPath, vcSettings_GetMapTileTypeName(pRenderer->pSettings), pBestNode->slippyPosition.x, pBestNode->slippyPosition.y, pBestNode->slippyPosition.z);
-		  udSprintf(serverAddress, "%s?lyrs=%s&x=%d&y=%d&z=%d", pRenderer->pSettings->maptiles.tileServerAddress, vcSettings_GetMapTileTypeName(pRenderer->pSettings), pBestNode->slippyPosition.x, pBestNode->slippyPosition.y, pBestNode->slippyPosition.z);
-		  break;
-	  default:
-		  break;
-	  }
-	  udReleaseMutex(pCache->pMutex);
+      switch (pRenderer->pSettings->maptiles.serverFlag)
+      {
+      case vcMTSF_Slippy:
+         udSprintf(localFileName, "%s/%s/%d/%d/%d.%s", pRenderer->pSettings->cacheAssetPath, udUUID_GetAsString(pRenderer->pSettings->maptiles.tileServerAddressUUID), pBestNode->slippyPosition.z, pBestNode->slippyPosition.x, pBestNode->slippyPosition.y, pRenderer->pSettings->maptiles.tileServerExtension);
+         udSprintf(serverAddress, "%s/%d/%d/%d.%s", pRenderer->pSettings->maptiles.tileServerAddress, pBestNode->slippyPosition.z, pBestNode->slippyPosition.x, pBestNode->slippyPosition.y, pRenderer->pSettings->maptiles.tileServerExtension);
+         break;
+      case vcMTSF_Google:
+         udSprintf(localFileName, "%s?lyrs=%s&x=%d&y=%d&z=%d", pRenderer->pSettings->cacheAssetPath, vcSettings_GetMapTileTypeName(pRenderer->pSettings), pBestNode->slippyPosition.x, pBestNode->slippyPosition.y, pBestNode->slippyPosition.z);
+         udSprintf(serverAddress, "%s?lyrs=%s&x=%d&y=%d&z=%d", pRenderer->pSettings->maptiles.tileServerAddress, vcSettings_GetMapTileTypeName(pRenderer->pSettings), pBestNode->slippyPosition.x, pBestNode->slippyPosition.y, pBestNode->slippyPosition.z);
+         break;
+      default:
+         udSprintf(localFileName, "%s", pRenderer->pSettings->cacheAssetPath);
+         udSprintf(serverAddress, "%s", pRenderer->pSettings->maptiles.tileServerAddress);
+         break;
+      }
+      
+      udReleaseMutex(pCache->pMutex);
 
       bool downloadingFromServer = true;
       char *pTileURL = serverAddress;
--- a/src/rendering/vcTileRenderer.cpp
+++ b/src/rendering/vcTileRenderer.cpp
@@ -218,8 +218,8 @@ uint32_t vcTileRenderer_LoadThread(void *pThreadData)
 		  udSprintf(serverAddress, "%s/%d/%d/%d.%s", pRenderer->pSettings->maptiles.tileServerAddress, pBestNode->slippyPosition.z, pBestNode->slippyPosition.x, pBestNode->slippyPosition.y, pRenderer->pSettings->maptiles.tileServerExtension);
 		  break;
 	  case vcMTSF_Google:
-		  udSprintf(localFileName, "%s?lyrs=s%40781&hl=zh-CN&gl=CN&x=%d&y=%d&z=%d", pRenderer->pSettings->cacheAssetPath, pBestNode->slippyPosition.x, pBestNode->slippyPosition.y, pBestNode->slippyPosition.z);
-		  udSprintf(serverAddress, "%s?lyrs=s%40781&hl=zh-CN&gl=CN&x=%d&y=%d&z=%d", pRenderer->pSettings->maptiles.tileServerAddress, pBestNode->slippyPosition.x, pBestNode->slippyPosition.y, pBestNode->slippyPosition.z);
+		  udSprintf(localFileName, "%s?\lyrs=s\%40781&\hl=\zh-CN&\gl=CN&x=%d&y=%d&z=%d", pRenderer->pSettings->cacheAssetPath, pBestNode->slippyPosition.x, pBestNode->slippyPosition.y, pBestNode->slippyPosition.z);
+		  udSprintf(serverAddress, "%s?\lyrs=s\%40781&\hl=\zh-CN&\gl=CN&x=%d&y=%d&z=%d", pRenderer->pSettings->maptiles.tileServerAddress, pBestNode->slippyPosition.x, pBestNode->slippyPosition.y, pBestNode->slippyPosition.z);
 		  break;
 	  default:
 		  break;
--- a/src/rendering/vcTileRenderer.cpp
+++ b/src/rendering/vcTileRenderer.cpp
@@ -218,8 +218,8 @@ uint32_t vcTileRenderer_LoadThread(void *pThreadData)
 		  udSprintf(serverAddress, "%s/%d/%d/%d.%s", pRenderer->pSettings->maptiles.tileServerAddress, pBestNode->slippyPosition.z, pBestNode->slippyPosition.x, pBestNode->slippyPosition.y, pRenderer->pSettings->maptiles.tileServerExtension);
 		  break;
 	  case vcMTSF_Google:
-		  udSprintf(localFileName, "%s?\lyrs=s\%40781&\hl=\zh-CN&\gl=CN&x=%d&y=%d&z=%d", pRenderer->pSettings->cacheAssetPath, pBestNode->slippyPosition.x, pBestNode->slippyPosition.y, pBestNode->slippyPosition.z);
-		  udSprintf(serverAddress, "%s?\lyrs=s\%40781&\hl=\zh-CN&\gl=CN&x=%d&y=%d&z=%d", pRenderer->pSettings->maptiles.tileServerAddress, pBestNode->slippyPosition.x, pBestNode->slippyPosition.y, pBestNode->slippyPosition.z);
+		  udSprintf(localFileName, "%s?lyrs=s&hl=zh-CN&gl=CN&x=%d&y=%d&z=%d", pRenderer->pSettings->cacheAssetPath, pBestNode->slippyPosition.x, pBestNode->slippyPosition.y, pBestNode->slippyPosition.z);
+		  udSprintf(serverAddress, "%s?lyrs=s&hl=zh-CN&gl=CN&x=%d&y=%d&z=%d", pRenderer->pSettings->maptiles.tileServerAddress, pBestNode->slippyPosition.x, pBestNode->slippyPosition.y, pBestNode->slippyPosition.z);
 		  break;
 	  default:
 		  break;
--- a/src/rendering/vcTileRenderer.cpp
+++ b/src/rendering/vcTileRenderer.cpp
@@ -218,8 +218,8 @@ uint32_t vcTileRenderer_LoadThread(void *pThreadData)
 		  udSprintf(serverAddress, "%s/%d/%d/%d.%s", pRenderer->pSettings->maptiles.tileServerAddress, pBestNode->slippyPosition.z, pBestNode->slippyPosition.x, pBestNode->slippyPosition.y, pRenderer->pSettings->maptiles.tileServerExtension);
 		  break;
 	  case vcMTSF_Google:
-		  udSprintf(localFileName, "%s?lyrs=s&hl=zh-CN&gl=CN&x=%d&y=%d&z=%d", pRenderer->pSettings->cacheAssetPath, pBestNode->slippyPosition.x, pBestNode->slippyPosition.y, pBestNode->slippyPosition.z);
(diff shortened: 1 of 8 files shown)


src/rendering/vcTileRenderer.cpp
```diff
@@ -209,9 +209,22 @@ uint32_t vcTileRenderer_LoadThread(void *pThreadData)
...
+	  {
+	  case vcMTSF_Unknown:
+		  break;
+	  case vcMTSF_Slippy:
+		  udSprintf(localFileName, "%s/%s/%d/%d/%d.%s", pRenderer->pSettings->cacheAssetPath, udUUID_GetAsString(pRenderer->pSettings->maptiles.tileServerAddressUUID), pBestNode->slippyPosition.z, pBestNode->slippyPosition.x, pBestNode->slippyPosition.y, pRenderer->pSettings->maptiles.tileServerExtension);
+		  udSprintf(serverAddress, "%s/%d/%d/%d.%s", pRenderer->pSettings->maptiles.tileServerAddress, pBestNode->slippyPosition.z, pBestNode->slippyPosition.x, pBestNode->slippyPosition.y, pRenderer->pSettings->maptiles.tileServerExtension);
+		  break;
+	  case vcMTSF_Google:
+		  udSprintf(localFileName, "%s?lyrs=s&hl=zh-CN&gl=CN&x=%d&y=%d&z=%d", pRenderer->pSettings->cacheAssetPath, pBestNode->slippyPosition.x, pBestNode->slippyPosition.y, pBestNode->slippyPosition.z);
+		  udSprintf(serverAddress, "%s?lyrs=s&hl=zh-CN&gl=CN&x=%d&y=%d&z=%d", pRenderer->pSettings->maptiles.tileServerAddress, pBestNode->slippyPosition.x, pBestNode->slippyPosition.y, pBestNode->slippyPosition.z);
```
OsiJr: The `zh-CN`, `zh`, etc. should really come from the localisation system

src/vcSettings.h
```diff
@@ -59,6 +59,21 @@ enum
   vcMetadataMaxLength = 256,
 };
 
+enum vcMapTileServerFlag
+{
+	vcMTSF_Unknown,
```
OsiJr: `Unknown` isn't required; defaulting to slippy is acceptable

src/vcSettings.h
```diff
@@ -59,6 +59,21 @@ enum
...
+	vcMTSF_Google,
+
+	vcMTSF_Count
+};
+
+const char g_vcMapTileServerName[vcMTSF_Count][64] = {
+	{ "unknown" },
+	{ "slippy" },
+	{ "google" }
+};
```
OsiJr: This is a weird format that is inconsistent with the other strings. We generally use:

```
const char *g_vcMapTileServerName[] = { "unknown", "slippy", "google" };
UDCOMPILEASSERT(udLengthOf(g_vcMapTileServerName) == vcMTSF_Count, "Update array");
```

(sometimes to split individual values on each line for larger enums)

src/vcModals.cpp
```diff
@@ -242,7 +242,19 @@ void vcModals_SetTileImage(void *pProgramStatePtr)
...
+  case vcMTSF_Unknown:
+	  break;
+  case vcMTSF_Slippy:
+	  udSprintf(buf, "%s/0/0/0.%s", pProgramState->settings.maptiles.tileServerAddress, pProgramState->settings.maptiles.tileServerExtension);
+	  break;
+  case vcMTSF_Google:
+	  udSprintf(buf, "%s?lyrs=s&hl=zh-CN&gl=CN&x=0&y=0&z=0", pProgramState->settings.maptiles.tileServerAddress);
+	  break;
+  default:
+	  break;
```
OsiJr: When iterating an enum; we encourage using all values and not using `default` in order to cause compilation issues when adding additional enum values in future

src/vcSettings.cpp
```diff
@@ -822,3 +824,16 @@ udResult vcSettings_UpdateLanguageOptions(vcSettings *pSetting)
 epilogue:
   return result;
 }
+
+void vcSettings_UpdateServerFlag(vcSettings *pSettings)
+{
+	pSettings->maptiles.serverFlag = vcMTSF_Unknown;
+	for (size_t i = 0; i < vcMTSF_Count; i++)
+	{
+		if (udStrstr(pSettings->maptiles.tileServerAddress, udStrlen(pSettings->maptiles.tileServerAddress), g_vcMapTileServerName[i]) != 0)
```
OsiJr: You can't do this check. There are no rules for slippy servers; _we_ have `slippy` in our domain but no one else does.

You need to provide a combo box to select between Slippy (default) and other providers (like Google, Bing, etc). Those platforms need to know the users API KEY but don't need the domain.

[See this list](https://wiki.openstreetmap.org/wiki/Tile_servers) for domain names for other servers.

src/rendering/vcTileRenderer.cpp
```diff
@@ -209,9 +209,22 @@ uint32_t vcTileRenderer_LoadThread(void *pThreadData)
       char localFileName[vcMaxPathLength];
       char serverAddress[vcMaxPathLength];
 
-      udSprintf(localFileName, "%s/%s/%d/%d/%d.%s", pRenderer->pSettings->cacheAssetPath, udUUID_GetAsString(pRenderer->pSettings->maptiles.tileServerAddressUUID), pBestNode->slippyPosition.z, pBestNode->slippyPosition.x, pBestNode->slippyPosition.y, pRenderer->pSettings->maptiles.tileServerExtension);
-      udSprintf(serverAddress, "%s/%d/%d/%d.%s", pRenderer->pSettings->maptiles.tileServerAddress, pBestNode->slippyPosition.z, pBestNode->slippyPosition.x, pBestNode->slippyPosition.y, pRenderer->pSettings->maptiles.tileServerExtension);
-      udReleaseMutex(pCache->pMutex);
+	  switch (pRenderer->pSettings->maptiles.serverFlag)
```
DamianMadden: This is indented once extra

src/vcSettings.h
zengweicheng666: vcMapTileServerFlag gives the vcMTSF_Unknown type by default before it is used to better determine whether it is assigned to another type.

src/vcSettings.cpp
zengweicheng666: Ok, I'll provide a combo box listing the available providers to choose from.The API KEY needs to be explored.

zengweicheng666: This feature will create a new branch instead of adding it.

zengweicheng666: This feature will recreate a branch and not be added to it.

OsiJr:

src/vcSettings.cpp
```diff
@@ -822,3 +827,24 @@ udResult vcSettings_UpdateLanguageOptions(vcSettings *pSetting)
 epilogue:
   return result;
 }
+
+void vcSettings_UpdateServerFlag(vcSettings *pSettings)
+{
+    UDCOMPILEASSERT(udLengthOf(g_vcMapTileServerName) == vcMTSF_Count, "Update array");
```
DamianMadden: This seems an unusual place for the compile assert, since it runs at compilation time you can place it anywhere and it shouldn't make a difference, but generally elsewhere in the code we have them immediately after the definition of the array or enum

src/vcSettings.cpp
```diff
@@ -822,3 +827,24 @@ udResult vcSettings_UpdateLanguageOptions(vcSettings *pSetting)
...
+
+    pSettings->maptiles.serverFlag = vcMTSF_Count;
+    for (size_t i = 0; i < vcMTSF_Count; i++)
+    {
+        if (udStrstr(pSettings->maptiles.tileServerAddress, udStrlen(pSettings->maptiles.tileServerAddress),g_vcMapTileServerName[i]) != 0)
```
DamianMadden: In this instance, since we know that tileServerAddress is null terminated, providing strlen is less efficient, preferably I'd pass a 0 in its place

src/vcSettings.cpp
```diff
@@ -822,3 +827,24 @@ udResult vcSettings_UpdateLanguageOptions(vcSettings *pSetting)
 epilogue:
   return result;
 }
+
+void vcSettings_UpdateServerFlag(vcSettings *pSettings)
+{
+    UDCOMPILEASSERT(udLengthOf(g_vcMapTileServerName) == vcMTSF_Count, "Update array");
+
+    pSettings->maptiles.serverFlag = vcMTSF_Count;
+    for (size_t i = 0; i < vcMTSF_Count; i++)
```
DamianMadden: ++i as per our coding standard

src/vcSettings.cpp
```diff
@@ -822,3 +827,24 @@ udResult vcSettings_UpdateLanguageOptions(vcSettings *pSetting)
...
+        {
+            pSettings->maptiles.serverFlag = (vcMapTileServerFlag)i;
+            break;
+        }
+    }
+}
+
+const char * vcSettings_GetMapTileTypeName(vcSettings *pSettings)
+{
+    UDCOMPILEASSERT(udLengthOf(g_vcMapTileTypeName) == vcMTT_Count, "Update array");
```
DamianMadden: Same comment as above

src/vcSettings.h
```diff
@@ -339,4 +359,8 @@ udResult vcSettings_RegisterAssetFileHandler();
 // Various settings helpers
 udResult vcSettings_UpdateLanguageOptions(vcSettings *pSettings);
 
+void vcSettings_UpdateServerFlag(vcSettings *pSettings);
+
+const char * vcSettings_GetMapTileTypeName(vcSettings *pSettings);
```
DamianMadden: These functions are perfect candidates to be inlined

src/vcModals.cpp
```diff
@@ -325,24 +338,39 @@ void vcModals_DrawTileServer(vcState *pProgramState)
     static bool s_isDirty = false;
     static int s_currentItem = -1;
 
-    const char *pItems[] = { "png", "jpg" };
+    const char *pSlippyItems[] = { "png", "jpg" };
+	const char *pGoogleItems[] = { "Plan","Image","Terrain" };
```
DamianMadden: indentation & blank line after this one

src/vcModals.cpp
```diff
@@ -325,24 +338,39 @@ void vcModals_DrawTileServer(vcState *pProgramState)
...
-      {
-        if (udStrEquali(pItems[i], pProgramState->settings.maptiles.tileServerExtension))
-          s_currentItem = i;
-      }
+        if (pProgramState->settings.maptiles.serverFlag == vcMTSF_Slippy)
+        {
+            for (int i = 0; i < (int)udLengthOf(pSlippyItems); ++i)
+            {
+                if (udStrEquali(pSlippyItems[i], pProgramState->settings.maptiles.tileServerExtension))
+                    s_currentItem = i;
```
DamianMadden: should break after this

src/vcModals.cpp
```diff
@@ -325,24 +338,39 @@ void vcModals_DrawTileServer(vcState *pProgramState)
...
+            s_currentItem = pProgramState->settings.maptiles.mapTileType;
+        }
     }
 
     if (ImGui::InputText(vcString::Get("settingsMapsTileServer"), pProgramState->settings.maptiles.tileServerAddress, vcMaxPathLength))
       s_isDirty = true;
 
-    if (ImGui::Combo(vcString::Get("settingsMapsTileServerImageFormat"), &s_currentItem, pItems, (int)udLengthOf(pItems)))
+    if (pProgramState->settings.maptiles.serverFlag == vcMTSF_Slippy&&
+        ImGui::Combo(vcString::Get("settingsMapsTileServerImageFormat"), &s_currentItem, pSlippyItems, (int)udLengthOf(pSlippyItems)))
```
DamianMadden: A space between _Slippy and && would be much appreciated

src/vcModals.cpp
```diff
@@ -325,24 +338,39 @@ void vcModals_DrawTileServer(vcState *pProgramState)
...
     {
-      udStrcpy(pProgramState->settings.maptiles.tileServerExtension, pItems[s_currentItem]);
+      udStrcpy(pProgramState->settings.maptiles.tileServerExtension, pSlippyItems[s_currentItem]);
       s_isDirty = true;
     }
+    else if (pProgramState->settings.maptiles.serverFlag == vcMTSF_Google&&
```
DamianMadden: Same comment as above

src/vcSettings.cpp
zengweicheng666: Ok, thanks for your comments.

src/vcSettings.cpp
zengweicheng666: Ok, thanks for your comments.I'm not sure what you mean, but I've optimized it: reduce the number of strlen calls.

src/vcSettings.cpp
zengweicheng666: Ok, thanks for your reminding, I will read our compilation standard carefully.

src/vcModals.cpp
zengweicheng666: All right.

src/vcSettings.h
zengweicheng666: All right.

src/vcModals.cpp
zengweicheng666: All right.

src/vcSettings.h
zengweicheng666: The cost of inlining here is too high, so you need to change where g_vcMapTileTypeName and g_vcMapTileServerName are declared

src/vcSettings.cpp
zengweicheng666: All right,Assertions have been placed where variables are declared.

src/rendering/vcTileRenderer.cpp
```diff
@@ -209,8 +209,22 @@ uint32_t vcTileRenderer_LoadThread(void *pThreadData)
...
+         udSprintf(serverAddress, "%s/%d/%d/%d.%s", pRenderer->pSettings->maptiles.tileServerAddress, pBestNode->slippyPosition.z, pBestNode->slippyPosition.x, pBestNode->slippyPosition.y, pRenderer->pSettings->maptiles.tileServerExtension);
+         break;
+      case vcMTSF_Google:
+         udSprintf(localFileName, "%s?lyrs=%s&x=%d&y=%d&z=%d", pRenderer->pSettings->cacheAssetPath, vcSettings_GetMapTileTypeName(pRenderer->pSettings), pBestNode->slippyPosition.x, pBestNode->slippyPosition.y, pBestNode->slippyPosition.z);
+         udSprintf(serverAddress, "%s?lyrs=%s&x=%d&y=%d&z=%d", pRenderer->pSettings->maptiles.tileServerAddress, vcSettings_GetMapTileTypeName(pRenderer->pSettings), pBestNode->slippyPosition.x, pBestNode->slippyPosition.y, pBestNode->slippyPosition.z);
+         break;
+      default:
+         udSprintf(localFileName, "%s", pRenderer->pSettings->cacheAssetPath);
+         udSprintf(serverAddress, "%s", pRenderer->pSettings->maptiles.tileServerAddress);
+         break;
```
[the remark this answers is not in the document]
zengweicheng666: All right.

src/vcModals.cpp
```diff
@@ -325,24 +338,42 @@ void vcModals_DrawTileServer(vcState *pProgramState)
     static bool s_isDirty = false;
     static int s_currentItem = -1;
 
-    const char *pItems[] = { "png", "jpg" };
+    const char *pSlippyItems[] = { "png", "jpg" };
+    const char *pGoogleItems[] = { "Plan","Image","Terrain" };
```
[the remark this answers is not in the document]
samsinsane: Additionally, neither of these should have the `p` prefix, they're both arrays while their elements are pointers.

src/vcModals.cpp
```diff
@@ -325,24 +338,42 @@ void vcModals_DrawTileServer(vcState *pProgramState)
...
+              if (udStrEquali(pSlippyItems[i], pProgramState->settings.maptiles.tileServerExtension))
+              {
+                s_currentItem = i;
+                break;
+              }
+            }
+        }
+        else if (pProgramState->settings.maptiles.serverFlag == vcMTSF_Google)
+        {
+            s_currentItem = pProgramState->settings.maptiles.mapTileType;
```
samsinsane: No. `mapTileType` is an index into the `pGoogleItems` array that is saved to disk. This means that we can never modify the `pGoogleItems` array, or if we do, we have to create a whole bunch of nonsense to handle this lazy option. Mimic what the Slippy code does for the file extension, it's the same concept. Don't be lazy.

src/vcModals.cpp
```diff
@@ -325,24 +338,42 @@ void vcModals_DrawTileServer(vcState *pProgramState)
...
+        }
     }
 
     if (ImGui::InputText(vcString::Get("settingsMapsTileServer"), pProgramState->settings.maptiles.tileServerAddress, vcMaxPathLength))
       s_isDirty = true;
 
-    if (ImGui::Combo(vcString::Get("settingsMapsTileServerImageFormat"), &s_currentItem, pItems, (int)udLengthOf(pItems)))
+    if (pProgramState->settings.maptiles.serverFlag == vcMTSF_Slippy &&
+        ImGui::Combo(vcString::Get("settingsMapsTileServerImageFormat"), &s_currentItem, pSlippyItems, (int)udLengthOf(pSlippyItems)))
```
samsinsane: We don't break an `if` conditional across multiple lines unless absolutely neccessary - which it isn't in this case. Please put it all back on one line.

src/vcModals.cpp
```diff
@@ -325,24 +338,42 @@ void vcModals_DrawTileServer(vcState *pProgramState)
...
     {
-      udStrcpy(pProgramState->settings.maptiles.tileServerExtension, pItems[s_currentItem]);
+      udStrcpy(pProgramState->settings.maptiles.tileServerExtension, pSlippyItems[s_currentItem]);
       s_isDirty = true;
     }
+    else if (pProgramState->settings.maptiles.serverFlag == vcMTSF_Google &&
+        ImGui::Combo(vcString::Get("settingsMapsTileType"), &s_currentItem, pGoogleItems, (int)udLengthOf(pGoogleItems)))
```
samsinsane: Same as above, don't break up `if` conditionals across multiple lines.

src/vcSettings.cpp
```diff
@@ -822,3 +830,22 @@ udResult vcSettings_UpdateLanguageOptions(vcSettings *pSetting)
...
   return result;
 }
+
+void vcSettings_UpdateServerFlag(vcSettings *pSettings)
+{
+    size_t sLength = udStrlen(pSettings->maptiles.tileServerAddress);
+    pSettings->maptiles.serverFlag = vcMTSF_Count;
+    for (size_t i = 0; i < vcMTSF_Count; ++i)
+    {
+        if (udStrstr(pSettings->maptiles.tileServerAddress, sLength, g_vcMapTileServerName[i]) != 0)
```
samsinsane: See [this comment](https://github.com/Euclideon/vaultclient/pull/209#discussion_r355850084) for why this needs to change.

OsiJr:
@zengweicheng666, I'm closing this because our satellite system is nearly running using the slippy coordinate conventions.

PR #210, opened 2019-12-11
Added 'Saturation' Appearance Option

peterAdams1014:
FXAA is now always run as a post process
[AB#546](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/546)

### DIFF
--- a/src/gl/metal/vcRenderShaders.cpp
+++ b/src/gl/metal/vcRenderShaders.cpp
@@ -3,8 +3,8 @@
 const char *const g_VisualizationFragmentShader = "visualizationFragmentShader";
 const char *const g_VisualizationVertexShader = "visualizationVertexShader";
 
-const char *const g_FXAAFragmentShader = "fxaaFragmentShader";
-const char *const g_FXAAVertexShader = "fxaaVertexShader";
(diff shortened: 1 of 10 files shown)


src/gl/metal/vcRenderShaders.cpp
```diff
@@ -3,8 +3,8 @@
 const char *const g_VisualizationFragmentShader = "visualizationFragmentShader";
 const char *const g_VisualizationVertexShader = "visualizationVertexShader";
 
-const char *const g_FXAAFragmentShader = "fxaaFragmentShader";
-const char *const g_FXAAVertexShader = "fxaaVertexShader";
+const char *const g_PostEffectsFragmentShader = "postEffectsFragmentShader";
+const char *const g_PostEffectsVertexShader = "postEffectsVertexShader";
```
samsinsane: Don't you need to modify the `.metal` file too?

peterAdams1014: Nope.

With the expectation that we'll get shader cross-compilation working soon, getting these shaders working for metal is a bit redundant.

PR #211, opened 2019-12-11
Hotkey fixes/Numeric settings

DamianMadden:
Resolves [AB#740](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/740)
Resolves [AB#696](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/696)


OsiJr:

OsiJr:
I like the changes but dislike the commit messages and PR title. Please work on these.

src/vcHotkey.cpp
```diff
@@ -152,8 +162,8 @@ namespace vcHotkey
     }
 
     pStrings[4] = SDL_GetScancodeName((SDL_Scancode)(mappedKey & 0x1FF));
-   
-    vcStringFormat(pBuffer, bufferLen, "{0}{1}{2}{3}{4}", pStrings, 5);
+    pStrings[5] = key >= vcB_Forward && key <= vcB_Down ? ", Shift Spd++, Ctrl Spd--" : "";
```
OsiJr: These strings are strange, intentional?

DamianMadden: I was conserving space, but probably don't need to, changed

DamianMadden:
This is now based on my other keybindings branch to avoid significant conflicts

src/vcHotkey.cpp
```diff
@@ -87,39 +91,50 @@ namespace vcHotkey
...
+    if (keyNum == vcMOD_Shift)
+      return io.KeyShift;
+    if (keyNum == vcMOD_Ctrl)
+      return io.KeyCtrl;
+    if (keyNum == vcMOD_Alt)
+      return io.KeyAlt;
+    if (keyNum == vcMOD_Super)
+      return io.KeySuper;
+
+    return ImGui::GetIO().KeysDown[keyNum & 0x1FF];
```
samsinsane: This is a magic number, this should be `vcMOD_Mask` or something.

src/vcHotkey.cpp
```diff
@@ -87,39 +91,50 @@ namespace vcHotkey
...
-      return false;
-    if (((keyNum & vcMOD_Ctrl) == vcMOD_Ctrl) != io.KeyCtrl)
-      return false;
-    if (((keyNum & vcMOD_Alt) == vcMOD_Alt) != io.KeyAlt)
-      return false;
-    if (((keyNum & vcMOD_Super) == vcMOD_Super) != io.KeySuper)
-      return false;
+    if (unique)
+    {
+      if (((keyNum & vcMOD_Shift) == vcMOD_Shift) != io.KeyShift)
```
samsinsane: This is very strange, is this what you intended to do?

DamianMadden: Yes, the unique parameter is to prevent keypresses like shift+x from triggering x for that bind, and the other part checks if the key includes a shift modifier and compares that to whether shift is pressed

samsinsane: It's just a very strange conditional; bitwise AND and _two_ comparisons for a single conditional.

DamianMadden: well the first part reflects how you guys told me was correct, my preference initially was to let the result of the AND suffice without comparison to the flag

samsinsane: Also, this feature seems broken now. This is all for the Altitude Lock, which uses Spacebar, correct?

When pressing Shift+Spacebar, this block is bypassed. However, when you rebind the altitude lock to use any modifier and Spacebar, it will always trigger with just spacebar.

You've got quite a lot of complexity in the conditional which makes it hard to expand the logic correctly, but ultimately you need to do:
```c++
bool containsShift = (keyNum & vcMOD_Shift) == vcMOD_Shift;
if (containsShift && !io.KeyShift || (unique && !containsShift && io.KeyShift))
  return false;
```

I'm not sure `unique` is the right name here?

DamianMadden: altitude lock passes false, so that you can trigger it while flying around with ctrl or shift held, because that was the only use case that carl mentioned was bad, the intention is that you pass false for any key, that wouldn't be bound as a modifier key + key, that you wish to be able to trigger while navigating, it's working as I intended... your way works too, but it should never be the case that someone passes false for a key with a modifier, because then effectively the key ultimately pressed would be mod + mod + key, and fail for each bind

samsinsane: Can you change the altitude lock key binding? When the altitude lock binding is set to Alt+Spacebar, does it trigger when just Spacebar is pressed?

DamianMadden: unique >> checkMod? *fail for the first bind be it mod + key*

samsinsane: I don't know what you're trying to tell me here? I ended up checking out your branch and what I was saying does happen, you can set the altitude lock to some mod key + spacebar, and pressing spacebar still triggers the altitude lock to toggle. This is broken. Also, when I first opened it none of the key bindings were mapped, I had to manually click "Load Defaults", this branch can't be merged if this continues to happen.

DamianMadden: So there's a problem, if we have 2 bindings: [alt+w] goes forward and [shift + a] goes left, then you press [alt+shift+w+a] then what do you want to happen? it takes it as [alt+shift+w] & [alt+shift+a]? or it ignores extra modifier keys and just checks that the one corresponding to them is down? do keys that don't have modifier keys trigger if any modifier is hit? and how would I enable carl to fly around holding space bar and still trigger altitude lock with it bound to space bar?

DamianMadden: If I 'consume' keypresses it *might* work depending on where in the code altitude lock is triggered vs movement

src/vcHotkey.cpp
```diff
@@ -91,39 +96,60 @@ namespace vcHotkey
...
+      contains = (keyNum & vcMOD_Alt) == vcMOD_Alt;
+      if (contains && !io.KeyAlt || (checkMod && !contains && io.KeyAlt))
+        return false;
+
+      contains = (keyNum & vcMOD_Super) == vcMOD_Super;
+      if (contains && !io.KeySuper || (checkMod && !contains && io.KeySuper))
+        return false;
+    }
+
+    return ImGui::IsKeyPressed((keyNum & vcMOD_Mask), false);
```
DamianMadden: Checkmod not working as intended, seems to have no effect

PR #222, opened 2020-01-10
Prevent binding Alt+F4

DamianMadden:
Resolves [AB#761](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/761)


src/vcHotkey.cpp
```diff
@@ -213,7 +213,7 @@ namespace vcHotkey
     {
       errors |= (1 << vcKES_Select);
 
-      if (pProgramState->currentKey)
+      if (pProgramState->currentKey && pProgramState->currentKey != vcMOD_Alt + SDL_SCANCODE_F4)
```
samsinsane: `(vcMOD_Alt + SDL_SCANCODE_F4)`.

DamianMadden: > I'm not sure if we should be blocking specific combinations, or just not saving immediately upon user input instead. Saving when the user clicks "Close" would probably be better, this would also give them the option to click "Cancel" and have the changes revert.

The saving is triggered by quitting the application, which I don't think should change

DamianMadden: Will do, but I think we should probably block specific combinations as well, binding alt-f4 is probably never useful/intended

samsinsane: > binding alt-f4 is probably never useful/intended

Alternatively, let the user decide what is useful/intended - from memory Alt+F4 is fine on macOS.

If you stop writing the key changes directly into the settings struct and instead have a struct for the "new bindings", when the user clicks the Close button you can then stamp the new bindings struct over the top of the settings struct. When the user uses Alt+F4 on Windows, it might momentarily display this change but the Close button isn't triggered. There's no need to block the user from setting anything if they're unable to click the Close button.

src/vcHotkey.cpp
```diff
@@ -13,7 +13,7 @@
 namespace vcHotkey
 {
   static int target = -1;
-  static int keyBinds[vcB_Count] = {};
+  static int keyBinds[2][vcB_Count] = {};
```
samsinsane: This just feels hacky, two unique variables would be much clearer too. `keyBinds` and `pendingKeyBinds` - the `bool right` used below should also change to `bool pending`, left/right terminology doesn't fit in this context.

src/vcHotkey.h
```diff
@@ -38,23 +38,24 @@ enum vcBind
 
 namespace vcHotkey
 {
+  void StampBinds(bool right);
```
samsinsane: I would recommend changing this too:
```c++
void FinalisePendingChanges();
void CancelPendingChanges();
```

src/vcModals.cpp
```diff
@@ -851,16 +851,20 @@ void vcModals_DrawBindings(vcState *pProgramState)
     vcHotkey::DisplayBindings(pProgramState);
 
     ImGui::Columns(3, "ButtonRow");
-    if (ImGui::Button(udTempStr("%s###bindingsClose" , vcString::Get("bindingsModalClose")), ImVec2(-1, 25)) || vcHotkey::IsPressed(vcB_Close))
-      ImGui::CloseCurrentPopup();
+    if (ImGui::Button(udTempStr("%s###bindingsRevert", vcString::Get("bindingsModalRevert")), ImVec2(-1, 25)) || vcHotkey::IsPressed(vcB_Cancel))
+      vcHotkey::StampBinds(true);
```
samsinsane: Cancel implies the popup closes - otherwise it would be called Revert.

DamianMadden: That naming is to operate as much as possible as a catch-all for anything we'd map to the escape key, so closing popups or reverting changes

samsinsane: My apologies, I misunderstood what was happening here.

src/vcHotkey.h
```diff
@@ -33,28 +33,30 @@ enum vcBind
...
 
 namespace vcHotkey
 {
+  void FinalisePendingChanges(bool apply);
   bool IsDown(int keyNum);
   bool IsDown(vcBind key);
   bool IsPressed(int keyNum);
   bool IsPressed(vcBind key);
-  void GetKeyName(vcBind key, char *pBuffer, uint32_t bufferLen);
+  void GetKeyName(vcBind key, char *pBuffer, uint32_t bufferLen, bool pending);
```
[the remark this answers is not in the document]
DamianMadden: The call in DisplayBindings has true

PR #227, opened 2020-01-14
Restart convert/Fix start convert with no input

DamianMadden:
Resolves [AB#337](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/337)


docs/TranslationGuide.md
```diff
@@ -3,16 +3,18 @@
 ## Changelist
 - Version 0.5.0
   - Added:
-    - Menus
-      - `menuBindings`, `settingsAppearanceSaturation`
```
DamianMadden: Ordered this alphabetically

src/vcConvert.cpp
```diff
@@ -398,18 +398,19 @@ void vcConvert_ShowUI(vcState *pProgramState)
   }
   else
   {
-    if (pSelectedJob->status == vcCQS_Preparing || pSelectedJob->status == vcCQS_Cancelled || pSelectedJob->status == vcCQS_WriteFailed || pSelectedJob->status == vcCQS_ParseFailed || pSelectedJob->status == vcCQS_ImageParseFailed || pSelectedJob->status == vcCQS_Failed)
+    if (pSelectedJob->status == vcCQS_Preparing || pSelectedJob->status == vcCQS_Cancelled || (pSelectedJob->status >= vcCQS_WriteFailed && pSelectedJob->status <= vcCQS_Failed))
```
samsinsane: We've discussed this previously, using ranges isn't a good idea as you're assuming that all "Failed" conditions are contiguous.

DamianMadden: Since there are 4 fail conditions already and potentially more to come, what if I added a flag value to the enum values for failure conditions and left a comment documenting it instead?

samsinsane: We discussed this in the other PR too, I don't recall what the outcome of that discussion was. @OsiJr thoughts?

DamianMadden: or just leave a comment to ensure failure conditions are kept contiguous?

samsinsane: Definitely not the comment option. There's no guarantee that this will happen, we can guarantee that a bit is set, but we can't guarantee that people pay attention to comments.

PR #230, opened 2020-01-16
Added load handler to vcWebFile

samsinsane:
Resolves [AB#251](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/251)

### DIFF
--- a/src/vcWebFile.cpp
+++ b/src/vcWebFile.cpp
@@ -7,6 +7,29 @@
 #include "udFileHandler.h"
 #include "udStringUtil.h"
 
+static udResult vcWebFile_Load(udFile *pFile, void **ppBuffer, int64_t *pBufferLength)
+{
+  UDTRACE();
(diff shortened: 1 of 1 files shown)


src/vcWebFile.cpp
```diff
@@ -7,6 +7,29 @@
 #include "udFileHandler.h"
 #include "udStringUtil.h"
 
+static udResult vcWebFile_Load(udFile *pFile, void **ppBuffer, int64_t *pBufferLength)
+{
+  UDTRACE();
```
OsiJr: This is odd; consistent with the rest of this file though

samsinsane: Yeah, I thought the same thing but figured I'd just keep it - it's probably because we just copied the udCore HTTP implementation and changed it.

PR #239, opened 2020-01-24
Prevent conversion failure when unloading models

DamianMadden:
Resolves [AB#715](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/715)

### DIFF
--- a/src/scene/vcModel.cpp
+++ b/src/scene/vcModel.cpp
@@ -14,6 +14,7 @@
 #include "vdkPointCloud.h"
 
 #include "udStringUtil.h"
+#include "udThread.h"
 
 struct vcModelLoadInfo
 {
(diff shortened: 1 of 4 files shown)


DamianMadden:
@samsinsane I might need you to elaborate, I agree about swapping to a reference counter in case there are multiple conversions using the pointcloud, but I'm not clear on what order of operations here is problematic in the base case of 1 conversion, putting aside the failure to account for the old model properly as well

samsinsane:
@DamianMadden You don't use a synchronization primitive to ensure that two threads don't access `destroyPointClouds`/`m_pConverting` at the same time. This means, that this: https://github.com/Euclideon/vaultclient/blob/912180302a41c61a3a3c194f9e485fef91a697dc/src/vcBPA.cpp#L1043-L1051
can fail the condition and end up in the else case. However, there's no guarantee that between those two things happening that this: https://github.com/Euclideon/vaultclient/blob/912180302a41c61a3a3c194f9e485fef91a697dc/src/scene/vcModel.cpp#L447-L450
won't be executed on another thread. This means that it's possible that neither thread is going to clean up the model, which is the race condition. While that doesn't sound so bad, it gets worse because that first block goes on to free `pBPA`, which could happen before the second block hits the else case. This will modify "random" memory - random in that we don't know what is there - or in the worst case, trigger a segfault and crash the program.

OsiJr:
Replaced by #312

PR #240, opened 2020-01-24
Use model SRID in displacement

DamianMadden:
Resolves [AB#573](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/573)

### DIFF
--- a/src/vcBPA.cpp
+++ b/src/vcBPA.cpp
@@ -1064,6 +1064,13 @@ void vcBPA_CompareExport(vcState *pProgramState, vdkPointCloud *pOldModel, vdkPo
   udJSON metadata = {};
   metadata.Parse(pMetadata);
 
+  for (uint32_t i = 0; i < metadata.MemberCount() ; ++i)
+  {
+    const udJSON *pElement = metadata.GetMember(i);
+    // Removed error checking because convertInfo metadata triggers vE_NotSupported
+    vdkConvert_SetMetadata(pConvertItem->pConvertContext, metadata.GetMemberName(i), pElement->AsString());
+  }
+
   vdkConvertCustomItem item = {};
   item.pName = "DisplacementComparison";
 


src/scene/vcModel.cpp
```diff
@@ -395,9 +395,13 @@ void vcModel::ContextMenuListModels(vcState *pProgramState, vdkProjectNode *pPar
       {
         vcModel *pOldModel = (vcModel *)pChildNode->pUserData;
         const double ballRadius = 0.15; // TODO: Expose this to the user
-        udWorkerPoolCallback callback = [this, pProgramState, pOldModel, ballRadius](void*)
+
+        int SRID = (this->m_pPreferredProjection ? this->m_pPreferredProjection->srid : (pOldModel->m_pPreferredProjection ? pOldModel->m_pPreferredProjection->srid : -1));
+        SRID = (SRID == -1 ? (this->m_pCurrentZone ? this->m_pCurrentZone->srid : (pOldModel->m_pCurrentZone ? pOldModel->m_pCurrentZone->srid : -1)) : SRID);
```
samsinsane: The point of the comparison feature is to duplicate the model and just add an attribute for displacement. This code should only be getting the SRID that is stored in the UDS, not an override value, nor the `pOldModel` value.

src/vcBPA.cpp
```diff
@@ -1067,6 +1067,9 @@ void vcBPA_CompareExport(vcState *pProgramState, vdkPointCloud *pOldModel, vdkPo
   vdkConvertCustomItem item = {};
   item.pName = "DisplacementComparison";
 
+  if (SRID != -1)
+    item.srid = SRID;
```
samsinsane: Couldn't this be pulled out of `pNewModel` instead of being passed in, we already have the `pNewModel` and we don't want the SRID to be overridden by anything ever.

DamianMadden: it could, as in vcModel_LoadMetadata, but it is a lot of work that's already done when loading the model
`if (vdkPointCloud_GetMetadata(pModel->m_pPointCloud, &pMetadata) == vE_Success)
  {
    pModel->m_metadata.Parse(pMetadata);
    pModel->m_hasWatermark = pModel->m_metadata.Get("Watermark").IsString();

    pModel->m_meterScale = pModel->m_pointCloudHeader.unitMeterScale;
    pModel->m_pivot = udDouble3::create(pModel->m_pointCloudHeader.pivot[0], pModel->m_pointCloudHeader.pivot[1], pModel->m_pointCloudHeader.pivot[2]);

    vcSRID srid = 0;
    const char *pSRID = pModel->m_metadata.Get("ProjectionID").AsString();
    const char *pWKT = pModel->m_metadata.Get("ProjectionWKT").AsString();

    if (pSRID != nullptr)
    {
      pSRID = udStrchr(pSRID, ":");
      if (pSRID != nullptr)
        srid = udStrAtou(&pSRID[1]);

      pModel->m_pPreferredProjection = udAllocType(udGeoZone, 1, udAF_Zero);
      if (udGeoZone_SetFromSRID(pModel->m_pPreferredProjection, srid) != udR_Success)
        udFree(pModel->m_pPreferredProjection);
    }

    if (pModel->m_pPreferredProjection == nullptr && pWKT != nullptr)
    {
      pModel->m_pPreferredProjection = udAllocType(udGeoZone, 1, udAF_Zero);

      if (udGeoZone_SetFromWKT(pModel->m_pPreferredProjection, pWKT) != udR_Success)
        udFree(pModel->m_pPreferredProjection);
    }...`

DamianMadden: Would you prefer I do that in vcBPA_CompareExport?

samsinsane: You need to use three backticks when posting multiple lines of code, you can also add `c++` after the first three backticks and it will format it as C++ code.

You might be able to forgo the SRID variable entirely and just feed all of the metadata back into the conversion system with `vdkConvert_SetMetadata` for each key value pair in the JSON object. Additionally, if the SRID is defined through the WKT, then we should use the WKT not the SRID as that's what the original model did too.

DamianMadden:
Untested presently, as the conversion would take too long, confirmed that nothing breaks though

samsinsane:
@DamianMadden you can test this in one of two ways, use a smaller model and compare it to itself, or return after the first grid has completed/generated (not sure which one I was doing, there's a loop you can just break out of somewhere). You don't need to verify that the comparison works.

DamianMadden:
True, tested with an early breakout, appears to be working

PR #241, opened 2020-01-24
Default naming & range setting for displacement

DamianMadden:
Resolves [AB#572](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/572)
Resolves [AB#703](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/703)


src/scene/vcModel.cpp
```diff
@@ -395,9 +395,21 @@ void vcModel::ContextMenuListModels(vcState *pProgramState, vdkProjectNode *pPar
...
         vcModel *pOldModel = (vcModel *)pChildNode->pUserData;
         const double ballRadius = 0.15; // TODO: Expose this to the user
-        udWorkerPoolCallback callback = [this, pProgramState, pOldModel, ballRadius](void*)
+
+        char newName[vcMaxPathLength] = {};
+        char oldName[vcMaxPathLength] = {};
+        int newLen = udFilename(this->m_pNode->pName).ExtractFilenameOnly(newName, sizeof(newName));
+        int oldLen = udFilename(pOldModel->m_pNode->pName).ExtractFilenameOnly(oldName, sizeof(oldName));
+
+        char nameBuffer[24] = { 'D', '_', '\0' };
```
samsinsane: Why 24? Also, looking at the code, you might be better off using `udSprintf` and injecting null terminators into `newname` and `oldName` where you want the cut offs, for simplicity. `udSprintf(nameBuffer, "D_%s-%s", oldName, newName);` or something similar. Alternatively, you could change this to be:
```c++
char *pNameBuffer = nullptr;
udSprintf(&pNameBuffer, "D_%s_%s", oldName, newname);
```
And just pass ownership of the string to `vcBPA_CompareExport` which then doesn't need to use `udStrdup`.

src/vcBPA.cpp
```diff
@@ -1037,7 +1037,7 @@ void vcBPA_ConvertDestroy(vdkConvertCustomItem *pConvertInput)
   udFree(pConvertInput->pData);
 }
 
-void vcBPA_CompareExport(vcState *pProgramState, vdkPointCloud *pOldModel, vdkPointCloud *pNewModel, double ballRadius)
+void vcBPA_CompareExport(vcState *pProgramState, vdkPointCloud *pOldModel, vdkPointCloud *pNewModel, double ballRadius, const char *name)
```
samsinsane: `pName` and same in the header.

src/vcSettingsUI.cpp
```diff
@@ -368,7 +370,13 @@ void vcSettingsUI_Show(vcState *pProgramState)
           }
 
           if (pProgramState->settings.visualization.mode == vcVM_Displacement)
-            ImGui::InputFloat2(vcString::Get("settingsVisDisplacementRange"), &pProgramState->settings.visualization.displacement.x);
+          {
+            if (ImGui::InputFloat2(vcString::Get("settingsVisDisplacementRange"), &pProgramState->settings.visualization.displacement.x))
```
samsinsane: From memory, this won't trigger if someone uses Ctrl+Click to enter the value?

DamianMadden: these aren't sliders, just input floats

samsinsane: Oh right, sorry.

src/scene/vcModel.cpp
DamianMadden: 24 is precisely the amount needed to include 2x10 strings and 'D', '_', (file) '_', (file2) '\0'

samsinsane: Yeah, I can see how you got to 24, but I don't see why you chose 24 - more specifically why you chose 10 characters of each string. Why 10?

DamianMadden: No particular reason, just that's about how big I'd want the output name to be, for it to fit in the side panel of the convert window or the scene explorer at default size and this resolution

samsinsane: Okay, I can understand why you've made this decision. However, I think naming conventions are far more likely to differ in the last 10 characters than the first 10 - I would expect the "what" to be the first part of the filename with the "when" being the last part, to group like files together.

Grabbing any amount of the filename is going to be tricky, due to all the issues related to path limits and what I've mentioned above. I think the safest thing would be to either:
- Increase the 10 limit to 50 or 100, to grab the most out of the filename.
- Leave it to the user to create a meaningful name. Change the default name to `DisplacementComparison_{UnixTimestamp}.uds`.

We should also add two items into the metadata; the names of the two files used to generate the comparison. I'm not sure what the best name would be, but I think `displacement.oldfile` and `displacement.newfile` are the best options I can think of right now.

PR #243, opened 2020-01-24
Remove duplicated code in vcStringFormat

DamianMadden:
Probably would like to test this a bit more before merging
Resolves [AB#826](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/826)

### DIFF
--- a/src/vcStringFormat.cpp
+++ b/src/vcStringFormat.cpp
@@ -3,12 +3,11 @@
 #include "udPlatformUtil.h"
 #include "udStringUtil.h"
 
-const char *vcStringFormat(const char *pFormatString, const char **ppStrings, size_t numStrings)
+inline uint32_t vcStringFormat_CalculateNewLength(const char *pFormatString, const char **ppStrings, size_t numStrings, uint32_t *pOldLength)
 {
-  int len;
   int newChars = 0;
-  int currArg;
-  int i;
+  uint32_t currArg;
+  uint32_t i;
 
   for (i = 0; pFormatString[i] != '\0'; ++i)
   {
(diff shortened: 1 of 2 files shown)


src/vcStringFormat.cpp
```diff
@@ -56,7 +57,7 @@ const char *vcStringFormat(const char *pFormatString, const char **ppStrings, si
     // Format Specifier
     if (pFormatString[i] == '{')
     {
-      if (len <= i + 1 || pFormatString[i + 1] < '0' || pFormatString[i + 1] > '9')
+      if (pFormatString[i + 1] == '\0' || pFormatString[i + 1] < '0' || pFormatString[i + 1] > '9')
```
samsinsane: I'm fairly certain that `\0` lies outside of `0` and `9`, this seems like a redundant check? Is this for the verbosity?

src/vcStringFormat.cpp
```diff
@@ -26,25 +25,27 @@ const char *vcStringFormat(const char *pFormatString, const char **ppStrings, si
 
       int length;
       currArg = udStrAtoi((pFormatString + i + 1), &length);
-      if (currArg > (int) numStrings || currArg < 0 || pFormatString[length + i + 1] != '}')
+      if (currArg > numStrings || pFormatString[length + i + 1] != '}')
```
samsinsane: It's very confusing to see `udStrAtoi` and the removal of the `< 0` check. I'm genuinely surprised that this doesn't generate any warnings, `currArg` is unsigned but `udStrAtoi` is returning a signed value. You should really change all of the `udStrAtoi` usages to `udStrAtou` for clarity.

src/vcStringFormat.cpp
```diff
@@ -26,25 +25,27 @@ const char *vcStringFormat(const char *pFormatString, const char **ppStrings, si
...
+    *pOldLength = i;
 
-  if (len == 0)
-    return pFormatString;
+  return i + newChars + 1;
+}
 
-  int newLength = len + newChars + 1;
-  char *pBuf = udAllocType(char, newLength, udAF_None);
+const char *vcStringFormat_FillBuffer(char *pBuf, const char *pFormatString, const char **ppStrings, size_t numStrings)
```
samsinsane: This should really take a buffer length, to avoid overrunning the buffer. It also seems unnecessary to have this function return `pBuf`?

src/vcStringFormat.cpp
DamianMadden: just didn't occur to me because I was only reworking the previous line to check the length without having the length

src/vcStringFormat.cpp
DamianMadden: In every instance of its use it's called after constructing a buffer of the perfect size or checking the length of the input buffer, this isn't intended for use outside of vcStringFormat.cpp

samsinsane: I'm not really worried about the usage otuside of this file - if it's not intented to be used outside of the file, you should mark them as `static` though. I'm far more worried about the two functions ending up out of sync with each other, data corruption is not fun to track down and I'd rather avoid it entirely.

DamianMadden: I've refactored the buffer argument version of the function to no longer calculate output size beforehand, in this iteration it will just output as much as it can and report how many characters were written, making this no longer a concern

samsinsane:
The latest changes don't really satisfy the linked issue of removing the duplicated code. It seems like the allocating function could just allocate the buffer and then call the buffered call? With whatever other additional logic it might have.

PR #244, opened 2020-01-28
SettingsUI & Hotkey interface changes

DamianMadden:
Removed 'revert' and related code from vcHotkey, moved error system from vcHotkey to a top bar on the settingsUI, changed to 'verbose' loading for settings & moved defaults into the code
Resolves [AB#837](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/837)


DamianMadden:
Sry, I figured it wouldn't be an issue given the simplicity of the unrelated changes

src/vcSettingsUI.cpp
```diff
@@ -21,7 +21,62 @@
...
+  vcSC_Visualization,
+  vcSC_Bindings,
+  vcSC_Convert,
+  vcSC_Count,
+  vcSC_Count,
+  vcSC_Count,
+  vcSC_Count
+};
+UDCOMPILEASSERT(vcSC_Count == 10, "Update the above mapping if necessary");
+UDCOMPILEASSERT(vcSR_Count == 11, "Update the above mapping if necessary");
```
samsinsane: These are very strange asserts and it's very difficult to tell why you've got both, your message doesn't explain anything. Based on counting the number of items in `categoryMapping`, I assume that `vcSR_Count == 11` is supposed to be ensuring that `categoryMapping` has 11 items? Did you consider having `categoryMapping[vcSR_Count]` to enforce the length of the array instead? I assume that `vcSC_Count == 10` is to make sure the values are correct, I would suggest using a switch statement which will make sure that `vcSC_Appearance` is always mapped to `vcSR_Appearance`, or have a number of asserts that ensure that `vcSR_Appearance == 0` and so on. Overall, these asserts aren't making it clear to anyone what is supposed to happen when they fail, and having to decipher meaning is really counter productive.

src/vcSettingsUI.cpp
```diff
@@ -21,7 +21,62 @@
...
+
+static int errors = 0;
```
samsinsane: This really should be in the `pProgramState` structure, global variables are strongly discouraged.

src/vcSettingsUI.cpp
DamianMadden: It's to ensure that if a settings category or settings ui region is added that the dev will create an entry if necessary in the mapping table such that the appropriate setting category can be reset or otherwise invoked in relation to the corresponding settings ui region in the settings page, I've documented this better

PR #245, opened 2020-01-28
Verbose Key Settings

DamianMadden:


DamianMadden:
I raised that concern as well, so I'll just direct you to Paul

OsiJr:
> This goes against the idea of the `defaultsettings.json` file, I also can't see any reason for this change; there's no linked issue. I'm inclined to reject this change as it doesn't really enforce anything, and I can only see mistakes being made when adding new bindings options.

I did ask for this change to occur. With the loop version, when we add new keys they won't get default bindings without resetting _all_ keys back to default (which doesn't (and shouldn't) happen by default). Some notification that there is a key conflict at load time would be beneficial but is outside the scope of this MR.

This change also brings key bindings in line with every other settings system; `defaultsettings.json` is used at first launch (and can be configured by enterprise IT departments at deployment) and settings that are added in future versions use the internal defaults to avoid users settings being destroyed.

@DamianMadden, The `defaultsettings.json` version should still exist though.

PR #246, opened 2020-01-28
Removed 'revert' and related code from vcHotkey

DamianMadden:


DamianMadden:
I think Paul said people were confused by it or something, the save button was just missed in my separating this from the bigger branch, I guess I could just add a condition like !programComplete to the save function?

PR #251, opened 2020-02-03
BPA Octree for grid discovery

DamianMadden:


src/vcBPA.cpp
```diff
@@ -792,97 +732,126 @@ size_t vcBPA_FindClosestTriangle(vcBPAGrid *pOldGrid, udDouble3 position)
...
+    {
+      pOctData = udAllocType(vcBPAOctantData, 1, udAF_Zero);
+      pOctData->pData = pData;
+      pOctData->ppNode = &node.pChildren + i;
+      pOctData->extents = node.extents * .5;
+      pOctData->center = node.center + (octantPositions[i] * pOctData->extents);
 
-        udSleep(500);
-      } while (pData->running);
+      udWorkerPool_AddTask(pData->pManifold->pPool, vcBPA_ProcessOctant, pOctData, true);
```
samsinsane: As I was suggesting above with the capturing lambda, you could change this block to be:
```c++
vcBPAOctNode **ppNode = &node.pChildren + i;
udDouble3 extents = node.extents * 0.5;
udDouble3 center = node.center + (octantPositions[i] * pOctData->extents);

udWorkerPool_AddTask(pData->pManifold->pPool, [pData, ppNode, extents, center](void*) { vcBPA_ProcessOctant(pData, ppNode, extents, center); }, nullptr, true);
```
You can probably change that `true` to a `false` - I can't remember what it's for - and then omit it entirely I think?

Also, you might be able to pass `node` by value, and then use the lambda to clean it up. If you do this, it's probably best to put the lambda on its own line since it's going to be multiple lines long (I don't recall how we do this in other places, it might be done within the function call itself rather than as another variable):
```c++
auto callback = […](…) { // or udCallback<void(void*)>
  ...
};
udWorkerPool_AddTask(…, callback, …);
```

DamianMadden:
Currently having issues, it looks promising and fairly quick but stalls after some time, I've probably just done something mindless

src/vcBPA.cpp
```diff
@@ -924,86 +954,97 @@ vdkError vcBPA_ConvertReadPoints(vdkConvertCustomItem *pConvertInput, vdkPointBu
...
 
-  if (pData->activeItem.pointIndex == 0 && pData->activeItem.pGrid == nullptr)
-    goto epilogue;
-
-  error = vdkAttributeSet_GetOffsetOfNamedAttribute(&pBuffer->attributes, "udDisplacement", &displacementOffset);
-  if (error != vE_Success)
-    return error;
-
-  for (size_t i = pData->activeItem.pointIndex; i < pData->activeItem.pGrid->vertices.length; ++i)
+  while (pBuffer->pointCount < pBuffer->pointsAllocated)
```
samsinsane: Unless I'm mistaken, this is changing the callback to process multiple grids per callback instead of just one? I highly recommend reverting this change, this is adding too much complexity to the PR. I don't think it's a necessary optimization at this point either, but for now we need to focus on the octree changes and nothing else.

src/vcBPA.cpp
```diff
@@ -924,86 +954,97 @@ vdkError vcBPA_ConvertReadPoints(vdkConvertCustomItem *pConvertInput, vdkPointBu
...
     pData->activeItem.pointIndex = 0;
 
+    int i = 0;
     do
     {
       if (udSafeDeque_PopFront(pData->pConvertItemData, &pData->activeItem) == udR_Success)
         break;
+
+      if (++i > 25)
+        return vE_Success;
```
samsinsane: This seems unnecessary to me, also, it really should `goto epilogue;` instead. Our coding standard is very clear on `return` statements, there's very few cases where we make exceptions on this.

src/vcBPA.cpp
```diff
@@ -1014,16 +1055,19 @@ void vcBPA_ConvertClose(vdkConvertCustomItem *pConvertInput)
...
   pBPA->activeItem.oldGrid.Deinit();
   pBPA->activeItem = {};
 
-  vcBPAConvertItemData itemData;
-  while (udSafeDeque_PopFront(pBPA->pQueueItems, &itemData) == udR_Success)
-    itemData.oldGrid.Deinit();
-  udSafeDeque_Destroy(&pBPA->pQueueItems);
+  vcBPAOctNode node;
+  while (udSafeDeque_PopFront(pBPA->pLoadList, &node) == udR_Success)
+    udFree(node.pChildren);
```
samsinsane: I don't believe that `pLoadList` contains all of the nodes at the end of execution? Above you call `PopBack` but you don't push that node back on and then the scope ends. I would recommend that you use the `rootNode` instead.

src/vcBPA.cpp
```diff
@@ -387,9 +336,6 @@ vcBPATriangle vcBPA_FindSeedTriangle(vcBPAGrid *pGrid, double ballRadius)
         {
           triangle.x = (uint64_t)nearbyPoints[i];
           triangle.y = (uint64_t)pointIndex;
-          a = p0 - p1;
-          b = p2 - p1;
-          triangleNormal = udNormalize(udCross(a, b));
```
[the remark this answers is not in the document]
DamianMadden: it's not referenced after this for anything

DamianMadden:
> I think this PR is starting to contain unrelated changes, we need to focus on the grid generation changes, the rest is just going to slow down the review process and make it harder to find bugs.

Sorry, but I've added even more changes now, I think in the past the threading model for this must have been glitching in such a way as to end the conversion somewhere along the way 'successfully', the old code was actually only getting a third of the points in the test files (pevs subtraction ones)... so I've changed it to allocate and free convert data in full as it goes and load limited the convertitemdata array, and some other stuff as part of reducing the memory usage - because it was topping out and dying at 60gb

src/vcBPA.cpp
```diff
@@ -69,15 +69,14 @@ struct vcBPAGrid
   udDouble3 minPos;
   udDouble3 maxPos;
 
-  static vcBPAGrid create(udDouble3 center, udDouble3 minPos, udDouble3 maxPos)
+  static vcBPAGrid *create(udDouble3 center, udDouble3 minPos, udDouble3 maxPos)
```
samsinsane: This against the coding standard.

samsinsane:
> Sorry, but I've added even more changes now,

This needs to stop happening - if you're finding problems with the system as a whole. Go back to master, fix them and submit a PR for each one. Don't just pile all the changes into a single PR, I don't have hours to pour over all of your changes every time you push the branch.

> I think in the past the threading model for this must have been glitching in such a way as to end the conversion somewhere along the way 'successfully', the old code was actually only getting a third of the points in the test files (pevs subtraction ones)...

To my knowledge, this was working fine with my original code - and not just on my machine. I'm very concerned that since I've stopped working on it, changes have gone in to break this.

> so I've changed it to allocate and free convert data in full as it goes and load limited the convertitemdata array, and some other stuff as part of reducing the memory usage - because it was topping out and dying at 60gb

I'm unsure how this is happening, there was a limit to the number of grids (100 at a time was the limit) when I last touched this code. Does this still exist? Are grids significantly larger now? Or is the memory consumption coming from somewhere else?

src/vcBPA.cpp
```diff
@@ -899,23 +919,35 @@ vdkError vcBPA_ConvertOpen(vdkConvertCustomItem *pConvertInput, uint32_t everyNt
...
   vdkPointCloud_GetHeader(pData->pNewModel, &header);
   udDouble4x4 storedMatrix = udDouble4x4::create(header.storedMatrix);
-  udDouble3 startAABBCenter = (storedMatrix * udDouble4::create(header.pivot[0], header.pivot[1], header.pivot[2], 1.0)).toVector3();
+  storedMatrix.extractTransforms(translation, scale, ypr);
+  udDouble3 center = (storedMatrix * udDouble4::create(header.boundingBoxCenter[0], header.boundingBoxCenter[1], header.boundingBoxCenter[2], 1.0)).toVector3();
+  udDouble3 extents = scale * udDouble3::create(header.boundingBoxExtents[0], header.boundingBoxExtents[1], header.boundingBoxExtents[2]);
 
-  udDouble3 halfGrid = udDouble3::create(pData->pManifold->gridSize / 2.0);
-  pData->pManifold->grids.PushBack(vcBPAGrid::create(startAABBCenter, startAABBCenter - halfGrid, startAABBCenter + halfGrid));
+  for (uint8_t i = 1; i < 3; ++i)
```
peterAdams1014: Just confirming, you're assuming `pData->shortAxis` is always `0`, so you're essentially skipping the first element?

src/vcBPA.cpp
```diff
@@ -792,97 +735,174 @@ size_t vcBPA_FindClosestTriangle(vcBPAGrid *pOldGrid, udDouble3 position)
...
+  while (pData->running)
+  {
+    if (udSafeDeque_PopBack(pData->pLoadList, &node) != udR_Success)
+    {
+      if (pData->pManifold->pointsCollected == pData->pManifold->targetPoints)
+        break;
 
-  vdkPointCloudHeader header;
-  vdkPointCloud_GetHeader(pData->pNewModel, &header);
+      udSleep(250);
```
peterAdams1014: Is this necessary?

src/vcBPA.cpp
DamianMadden: Yes

src/vcBPA.cpp
DamianMadden: It doesn't but close can be called partway through the process

src/vcBPA.cpp
DamianMadden: Not entirely, but the chokepoint is memory blowout due to load in convertData, so this can also throttle that to an extent

src/vcBPA.cpp
samsinsane: My point is that you're only cleaning up the nodes that are in the `pLoadList` - which isn't all of them. This is a memory leak that needs to be fixed.

DamianMadden:
Update: The memory issues were the result of work accumulating due to the thread processing it getting stuck in udConvert::TransformPointBuffer, which is still preventing the process completing despite the memory now being managed very conservatively, I'm presently making unit test(s) to illustrate the issue after consulting with Sam

PR #252, opened 2020-02-03
Update watermark to new logo

DamianMadden:
Resolves [AB#789](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/789)


DamianMadden:
> I'm not sure why EuclideonLogo.png was deleted? I thought there was still a case where it was used?

Not that I can see, searching brings up nothing, I think it was vestigial before and the actual data used was stored in internal textures data

PR #253, opened 2020-02-10
open profile model by clicking user name

asunawu:
Fixes [AB#832](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/832)

### DIFF
--- a/builds/assets/lang/enAU.json
+++ b/builds/assets/lang/enAU.json
@@ -280,9 +280,9 @@
   "menuLogoutMessage": "You were logged out.",
 
   "menuProfileTitle": "Profile",
-  "menuUserNameLabel": "userName",
-  "menuRealNameLabel": "realName",
-  "menuEmailLabel": "email",
+  "menuUserNameLabel": "User Name",
+  "menuRealNameLabel": "Real Name",
+  "menuEmailLabel": "Email",
 
   "menuReleaseNotes": "Release Notes",
   "menuReleaseNotesTitle": "Release Notes",
--- a/builds/assets/lang/enAU.json
+++ b/builds/assets/lang/enAU.json
@@ -280,6 +280,9 @@
   "menuLogoutMessage": "You were logged out.",
 
   "menuProfileTitle": "Profile",
+  "menuUserNameLabel": "userName",
+  "menuRealNameLabel": "realName",
+  "menuEmailLabel": "email",
 
   "menuReleaseNotes": "Release Notes",
   "menuReleaseNotesTitle": "Release Notes",
--- a/builds/assets/lang/enAU.json
+++ b/builds/assets/lang/enAU.json
@@ -279,6 +279,8 @@
   "menuLogoutTitle": "Logged Out",
   "menuLogoutMessage": "You were logged out.",
 
+  "menuProfileTitle": "Profile",
+
   "menuReleaseNotes": "Release Notes",
   "menuReleaseNotesTitle": "Release Notes",
   "menuReleaseNotesFail": "Unable to load release notes from package.",
--- a/builds/assets/lang/zhCN.json
+++ b/builds/assets/lang/zhCN.json
@@ -287,6 +287,9 @@
   "menuLogoutCloseButton": "关闭",
 
   "menuProfileTitle": "资料",
+  "menuUserNameLabel": "用户名",
+  "menuRealNameLabel": "全名",
+  "menuEmailLabel": "邮箱",
 
   "menuReleaseNotes": "版本说明",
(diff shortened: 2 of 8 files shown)


builds/assets/lang/zhCN.json
```diff
@@ -9,6 +9,8 @@
     "targetVersion": "0.5.0"
   },
 
+  "popupClose": "关闭",
```
DamianMadden: This will probably need to be in the english translation if used also

src/vcModals.cpp
```diff
@@ -563,6 +564,35 @@ void vcModals_DrawImageViewer(vcState *pProgramState)
   }
 }
 
+void vcModals_DrawProfile(vcState* pProgramState)
+{
+  const char* userName = pProgramState->settings.loginInfo.username;
+  const char* profile = vcString::Get("menuProfileTitle");
```
DamianMadden: To match the coding standard these should be formatted as 
const char *pUserName 
const char *pProfile

src/vcModals.cpp
```diff
@@ -563,6 +564,35 @@ void vcModals_DrawImageViewer(vcState *pProgramState)
...
+
+  if (pProgramState->openModals & (1 << vcMT_Profile))
+    ImGui::OpenPopup(profile);
+
+  float width = 300.f;
+  float height = 250.f;
```
samsinsane: I don't believe we normally have local variables like this, however, perhaps we should reconsider for the additional clarity in the UI code.

src/vcModals.cpp
```diff
@@ -563,6 +564,35 @@ void vcModals_DrawImageViewer(vcState *pProgramState)
...
+  float btw = 80.f;
+  ImGui::SetNextWindowSize(ImVec2(width, height), ImGuiCond_Always);
+
+  if (ImGui::BeginPopupModal(profile, nullptr, ImGuiWindowFlags_NoResize))
+  {
+    pProgramState->modalOpen = true;
+    ImGui::TextUnformatted(userName);
+
+    if (ImGui::Button(vcString::Get("menuLogout"), ImVec2(btw, 0)))
+      vcSession_Logout(pProgramState);
```
DamianMadden: Just to check what behaviour is intended? Not sure, but I expect this will also close the popup due to losing the context, and open the logged out modal, not entirely sure we want to see that after clicking to logout, maybe check with Paul, but you can use pProgramState->openModals = 0. or openModals &= ^vcMT_LoggedOut I think

src/vcModals.cpp
```diff
@@ -563,6 +564,35 @@ void vcModals_DrawImageViewer(vcState *pProgramState)
...
+{
+  const char* userName = pProgramState->settings.loginInfo.username;
+  const char* profile = vcString::Get("menuProfileTitle");
+
+  if (pProgramState->openModals & (1 << vcMT_Profile))
+    ImGui::OpenPopup(profile);
+
+  float width = 300.f;
+  float height = 250.f;
+  float btw = 80.f;
```
samsinsane: We try to avoid shortened names like this, `buttonWidth` would be the preferred name here.

builds/assets/lang/zhCN.json
asunawu: There is English already.

src/vcModals.cpp
OsiJr: There is a variable to trigger that does this more cleanly- From memory its `pProgramState->forceLogout`

src/vcModals.cpp
OsiJr: Yeah, I said to @asunawu verbally that the UI tends to have magic numbers embedded in the ImGui calls but also that its not really standardised for UI magic numbers.

samsinsane: Sounds good :+1: - I might just resolve this conversation since I didn't want this to be changed.

builds/assets/lang/enAU.json
```diff
@@ -280,6 +280,9 @@
   "menuLogoutMessage": "You were logged out.",
 
   "menuProfileTitle": "Profile",
+  "menuUserNameLabel": "userName",
```
peterAdams1014: Should these be `"User Name"` and `"Real Name"`?

asunawu: I think you are right. I'll fix it soon.

asunawu: Fixed.

src/vcModals.cpp
asunawu: Fixed.

PR #254, opened 2020-02-11
Added command line wildcard support and dongle support

OsiJr:


vcConvertCMD/src/vcConvertCMD.cpp
```diff
@@ -174,6 +165,8 @@ int main(int argc, const char **ppArgv)
   vdkConvertContext *pModel = nullptr;
   vcConvertCMDSettings settings = {};
 
+  settings.files.Init(256);
```
samsinsane: There's no call to `Deinit` this?

vcConvertCMD/src/vcConvertCMD.cpp
```diff
@@ -56,19 +57,18 @@ struct vcConvertCMDSettings
...
   const char *pProxyPassword;
-  const char **ppInputFiles;
-  int32_t inputFileCount;
   int32_t srid;
   double globalOffset[3];
   bool pause;
   bool pauseOnError;
   bool autoOverwrite;
+
+  udChunkedArray<const char*> files;
```
peterAdams1014: This needs to be in a separate PR.

vcConvertCMD/src/vcConvertCMD.cpp
OsiJr: Will fix.

PR #255, opened 2020-02-11
Potentially fix logout modal repopping

DamianMadden:


OsiJr:
I can't see how this fixes the linked issue; its more likely that force logout isn't reset properly? This will remove the logout modal but the user is still erroneously logged out?

DamianMadden:
> I can't see how this fixes the linked issue; its more likely that force logout isn't reset properly? This will remove the logout modal but the user is still erroneously logged out?

Yeah actually it probably doesn't, but this makes the forced logout follow the same steps as a regular logout, the modal is called to open inside vcSession_Logout

DamianMadden:
After testing, it seems to work as expected either way so I imagine this was fixed and just not closed

PR #258, opened 2020-02-11
BPA with Octree, core changes

DamianMadden:


samsinsane:
This isn't any better than the other PR, please split this into multiple PRs. Removing or replacing a member of a struct should be its own PR, adding members to a struct should be its own PR, changing a function to allocate memory and return a pointer should its own PR, etc. One massive PR with all changes munged together makes it really difficult to understand what problem you were trying to solve.

PR #264, opened 2020-02-12
Show Chinese text in "Unsupported Files" window when loading failed.

asunawu:
Resolved [AB#160](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/160).
Resolved [AB#158](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/158).

### DIFF
--- a/src/vcModals.cpp
+++ b/src/vcModals.cpp
@@ -480,8 +480,27 @@ void vcModals_DrawUnsupportedFiles(vcState *pProgramState)
 
       ImGui::GetCurrentWindow()->DC.CurrentLineTextBaseOffset = offset;
 
-      const char *errorStr = udTempStr("error%s", udResultAsString(pProgramState->errorItems[i].resultCode));
-      ImGui::TextUnformatted(vcString::Get(errorStr));
+      int errorCode = pProgramState->errorItems[i].resultCode;
+      const char *pErrorString = nullptr;
+      switch (errorCode)
+      {
+      case udR_CorruptData:
+        pErrorString = vcString::Get("errorCorruptData");
+        break;
+      case udR_Unsupported:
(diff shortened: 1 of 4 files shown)


src/vcModals.cpp
```diff
@@ -470,7 +470,9 @@ void vcModals_DrawUnsupportedFiles(vcState *pProgramState)
       ImGui::NextColumn();
 
       ImGui::GetCurrentWindow()->DC.CurrentLineTextBaseOffset = offset;
-      ImGui::TextUnformatted(udResultAsString(pProgramState->errorItems[i].resultCode));
+
+      const char *errorStr = udTempStr("error%s", udResultAsString(pProgramState->errorItems[i].resultCode));
```
peterAdams1014: We typically prefix string types with `p`. Additionally we don't typically shorten words. So in this case this variable name should be `const char *pErrorString = ...`.

asunawu: Got it. Thanks.

OsiJr: This will also cause most errors to not be translated- I would recommend a switch of the specifically handled error with a `default` for generic errors

asunawu: The udResultAsString is used only in the unsupported files window currently, I translated those errors related with files for this quick fix. I think if we don't use it elsewhere, that's fine, otherwise we need to create a new interface.

asunawu: And there is another issue, we show "open failure" for a empty folder, it's a litter strange. A default message can resolve this problem, I'll talk to our QA about this fix.

asunawu: Fixed both variable name and a default string for generic errors.

PR #269, opened 2020-02-12
Define structs and storage for Octree components

DamianMadden:


src/vcBPA.cpp
```diff
@@ -1024,6 +1046,9 @@ void vcBPA_ConvertClose(vdkConvertCustomItem *pConvertInput)
     itemData.oldGrid.Deinit();
   udSafeDeque_Destroy(&pBPA->pQueueItems);
 
+  while (udSafeDeque_PopFront(pBPA->pLoadList, &pBPA->rootNode) == udR_Success);
```
samsinsane: I don't believe this is necessary and it's kind of confusing how you've written it - for empty while loops the `;` goes on the next line so that it's obvious what the intention was.

DamianMadden: Hmmn, yeah I suppose this is a vestige of when octree nodes had pointers to children

DamianMadden: There's definitely no issue calling Deinit() on a full chunked array though?

PR #270, opened 2020-02-12
Miscellaneous Bug Fixes

DamianMadden:
Ignoring whitespace changes for this eliminates most of the changes, alot just gets wrapped in a while loop

### DIFF
--- a/src/vcBPA.cpp
+++ b/src/vcBPA.cpp
@@ -324,7 +324,7 @@ void vcBPA_GetNearbyPoints(vcBPAGrid *pGrid, udChunkedArray<uint64_t> *pPoints, 
     size_t j = 0;
     for (; j < pPoints->length; ++j)
     {
-      udDouble3 p2 = pGrid->vertices[j].position;
+      udDouble3 p2 = pGrid->vertices[*pPoints->GetElement(j)].position;
 
       if (udMagSq3(p0 - p2) > d01)
         break;
@@ -387,9 +387,6 @@ vcBPATriangle vcBPA_FindSeedTriangle(vcBPAGrid *pGrid, double ballRadius)
         {
           triangle.x = (uint64_t)nearbyPoints[i];
           triangle.y = (uint64_t)pointIndex;
-          a = p0 - p1;
-          b = p2 - p1;
-          triangleNormal = udNormalize(udCross(a, b));
         }
 
         // Test that a p-ball with center in the outward halfspace touches all three vertices and contains no other points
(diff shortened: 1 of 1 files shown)


src/vcBPA.cpp
```diff
@@ -387,9 +387,6 @@ vcBPATriangle vcBPA_FindSeedTriangle(vcBPAGrid *pGrid, double ballRadius)
         {
           triangle.x = (uint64_t)nearbyPoints[i];
           triangle.y = (uint64_t)pointIndex;
-          a = p0 - p1;
-          b = p2 - p1;
-          triangleNormal = udNormalize(udCross(a, b));
```
samsinsane: This isn't a bug fix?

src/vcBPA.cpp
```diff
@@ -923,87 +920,94 @@ vdkError vcBPA_ConvertReadPoints(vdkConvertCustomItem *pConvertInput, vdkPointBu
   // Reset point count to avoid infinite loop
   pBuffer->pointCount = 0;
 
-  vcBPAConvertItem *pData = (vcBPAConvertItem *)pConvertInput->pData;
-  bool getNextGrid = true;
+  while (pBuffer->pointCount == 0)
```
samsinsane: What bug does this fix?

src/vcBPA.cpp
```diff
@@ -923,87 +920,94 @@ vdkError vcBPA_ConvertReadPoints(vdkConvertCustomItem *pConvertInput, vdkPointBu
...
 
-    udDouble3 position = pData->activeItem.pGrid->vertices[i].position;
-    size_t triangle = vcBPA_FindClosestTriangle(&pData->activeItem.oldGrid, position);
+      udDouble3 position = pData->activeItem.pGrid->vertices[i].position;
 
-    double distance = FLT_MAX;
-    if (triangle < pData->activeItem.oldGrid.triangles.length)
-      distance = udAbs(vcBPA_DistanceToTriangle(&pData->activeItem.oldGrid, triangle, position));
+      if (!udPointInAABB(position, pData->activeItem.pGrid->minPos, pData->activeItem.pGrid->maxPos))
+          continue;
```
samsinsane: What bug does this fix?

DamianMadden: multiple inclusion of points that were outside the AABB for the sake of the algorithm to complete boundary triangles

src/vcBPA.cpp
DamianMadden: with the exclusion of points outside the AABB of a grid it's possible that grids are present with 0 points, this prevents early termination of the conversion by proceeding to subsequent grids

src/vcBPA.cpp
DamianMadden: Not precisely...

src/vcBPA.cpp
samsinsane: `pGrid` shouldn't contain any points outside the AABB - we don't mesh `pGrid`, we mesh `pOldGrid`. If `pGrid` contains any points outside the AABB, that should be fixed instead.

OsiJr:
Replaced by #291

PR #272, opened 2020-02-12
Sanitise scene item names

DamianMadden:
Resolves [AB#743](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/743)
Resolves [AB#680](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/680)

### DIFF
--- a/src/vcStringFormat.cpp
+++ b/src/vcStringFormat.cpp
@@ -3,6 +3,10 @@
 #include "udPlatformUtil.h"
 #include "udStringUtil.h"
 
+#define VC_STRING_FORMAT_BUFFER_COUNT 16
+static volatile int32_t s_vcStringBufferIndex = 0;
+static char s_vcStringBuffers[VC_STRING_FORMAT_BUFFER_COUNT][64];
+
 const char *vcStringFormat(const char *pFormatString, const char **ppStrings, size_t numStrings)
 {
   int len;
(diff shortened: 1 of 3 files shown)


src/vcStringFormat.cpp
```diff
@@ -186,3 +190,67 @@ const char *vcStringFormat(char *pBuffer, size_t bufLen, const char *pFormatStri
 {
   return vcStringFormat(pBuffer, bufLen, pFormatString, &pString, 1);
 }
+
+// ****************************************************************************
+// Author: Damian Madden, February 2020
+const char *vcStringFormat_ImGuiClean(const char *pStr)
```
samsinsane: This function name doesn't really make any sense?

DamianMadden: Yes I suppose a better fix would be to change udFilename, but solving it on either end works, and this fix is needed to for the other issue linked, the name implies that it makes a string clean for imgui

samsinsane: > Yes I suppose a better fix would be to change udFilename, but solving it on either end works

Just to clarify, `udFilename` crashes when given a path with a `%` symbol in it?

> the name implies that it makes a string clean(sanitised) for imgui

I guess I wasn't very clear in why it didn't make any sense - the issues indicate that the problem is with `udFilename` and yet this function implies that it's a problem with ImGui, this doesn't make any sense.

DamianMadden: Yeah it crashes if passed a % symbol without a corresponding argument

samsinsane: Then this should be fixed in `udFilename` instead of here - it looks like calls to `udFilename::SetFromFullPath` need to use `("%s", args)` instead of just `(args)`. Can you also add a unit test for this, call each Set function with a path that contains `%s` or something and expect the path to still be `%s`.

DamianMadden: "...but solving it on either end works, and this fix is needed to for the other issue linked", which crashes in imgui

DamianMadden: I'd think since as far as QA is concerned this solves both issues, then we could add one like "Fix udFiename handling % characters"

samsinsane: > "...but solving it on either end works, and this fix is needed to for the other issue linked", which crashes in imgui

I'm not finding this crash in ImGui you're talking about, neither linked issue says anything about ImGui. I've patched `udFilename` and an oversight in the `vcWebFile` load function, and I'm not seeing any crashes in ImGui? I was assuming this was going to be the same problem, passing in the filename as a format string instead of as an argument, however, it doesn't seem to be an issue at all.

> I'd think since as far as QA is concerned this solves both issues, then we could add one like "Fix udFiename handling % characters"

Tricking QA into thinking something is fixed isn't how you should be approaching bug fixes.

PR #274, opened 2020-02-13
Fixed object file loading.

asunawu:
Resolved [AB#784](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/784)

Allow to load the object file without a material file.
Allow to create meshes without face matrix.

### DIFF
--- a/src/parsers/vcOBJ.cpp
+++ b/src/parsers/vcOBJ.cpp
@@ -404,7 +404,10 @@ udResult vcOBJ_Load(vcOBJ **ppOBJ, const char *pFilename)
       pTextBuffer[bufferPos + charCount] = temp;
       pOBJ->basePath.SetFilenameWithExt(f.GetFilenameWithExt());
 
-      UD_ERROR_CHECK(vcOBJ_LoadMtlLib(pOBJ, pOBJ->basePath.GetPath()));
+      //Allow to load an object file without a material file.
+      result = vcOBJ_LoadMtlLib(pOBJ, pOBJ->basePath.GetPath());
+      if (result != udR_OpenFailure || result != udR_Success)
+        UD_ERROR_SET(result);
     }
     else
     {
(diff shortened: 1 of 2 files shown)


src/parsers/vcOBJ.cpp
```diff
@@ -404,7 +404,10 @@ udResult vcOBJ_Load(vcOBJ **ppOBJ, const char *pFilename)
       pTextBuffer[bufferPos + charCount] = temp;
       pOBJ->basePath.SetFilenameWithExt(f.GetFilenameWithExt());
 
-      UD_ERROR_CHECK(vcOBJ_LoadMtlLib(pOBJ, pOBJ->basePath.GetPath()));
+      //Allow to load an object file without a material file.
+      result = vcOBJ_LoadMtlLib(pOBJ, pOBJ->basePath.GetPath());
+      if (result != udR_OpenFailure || result != udR_Success)
```
OsiJr:
```suggestion
      if (result != udR_OpenFailure && result != udR_Success)
```

asunawu: Fixed

src/rendering/vcPolygonModel.cpp
```diff
@@ -308,7 +308,7 @@ udResult vcPolygonModel_CreateFromOBJ(vcPolygonModel **ppPolygonModel, const cha
   {
     vcOBJ::Face *pFace = &pOBJReader->faces[f];
     if (pFace->mat == -1)
-      continue;
+      pFace->mat = 0;
```
DamianMadden: It looks to me like the vcOBJ_Load function doesn't set a default material at index 0, it might be prudent to define one, otherwise it's less clear that the model isn't likely supposed to look like that, so above where it says
``` // always have one
  if (pOBJReader->materials.length == 0)
  {
    vcOBJ::Material *pMat = pOBJReader->materials.PushBack();
    pMat->Ka = udFloat3::create(1.f);
    pMat->Kd = udFloat3::create(1.f);
  }```
Maybe push that back regardless of the condition and here set pFace->mat to ->materials.length - 1

DamianMadden: Or I could be concerning myself with an improper use of the obj format, idk

peterAdams1014: Could work yeah. I'd rather change as little as possible though.

PR #278, opened 2020-02-13
Added colour setting for compass to resolve visibly lost in yellow maps.

asunawu:
Resolve [AB#219](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/219)


peterAdams1014:
The compass is being removed at a later stage. So this PR is not longer needed.

PR #281, opened 2020-02-14
Added 3D Textures, and Texture Arrays

peterAdams1014:
[AB#974](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/974)

### DIFF
--- a/src/gl/directx11/vcTexture.cpp
+++ b/src/gl/directx11/vcTexture.cpp
@@ -473,7 +473,7 @@ bool vcTexture_EndReadPixels(vcTexture *pTexture, uint32_t x, uint32_t y, uint32
   pPixelData = ((uint32_t *)msr.pData) + (x + y * pTexture->width);
   for (int i = 0; i < (int)height; ++i)
   {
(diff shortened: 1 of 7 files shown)


src/gl/directx11/vcTexture.cpp
```diff
@@ -421,7 +473,7 @@ bool vcTexture_EndReadPixels(vcTexture *pTexture, uint32_t x, uint32_t y, uint32
   pPixelData = ((uint32_t *)msr.pData) + (x + y * pTexture->width);
   for (int i = 0; i < (int)height; ++i)
   {
-    memcpy((uint8_t*)pPixels + (i * pTexture->width), pPixelData, width * pixelBytes);
+    memcpy((uint8_t *)pPixels + (i * pTexture->width), pPixelData, width * pixelBytes);
```
OsiJr: There are quite a lot of these throughout this MR- I think your VS settings are set to "right only" instead of "leave alone"

peterAdams1014:
Metal support will be added later.

PR #282, opened 2020-02-14
Fix vcHotkey button borders

DamianMadden:
Use ImGui variables/color enum values instead of draw calls for button borders
Resolves [AB#865](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/865)


OsiJr:
I have an alternative solution to this incoming that doesn't use the error values- unsure why @peterAdams1014 unassigned me.

PR #287, opened 2020-02-17
vcBPA now uses octree for grid generation

samsinsane:


src/vcBPA.cpp
```diff
@@ -137,47 +143,40 @@ void vcBPA_Deinit(vcBPAManifold **ppManifold)
...
-    udDouble3 boundingBoxCenter = udDouble3::create(header.boundingBoxCenter[0], header.boundingBoxCenter[1], header.boundingBoxCenter[2]);
-    udDouble3 boundingBoxExtents = udDouble3::create(header.boundingBoxExtents[0], header.boundingBoxExtents[1], header.boundingBoxExtents[2]);
-    udDouble3 modelMin = (storedMatrix * udDouble4::create(boundingBoxCenter - boundingBoxExtents, 1.0)).toVector3();
-    udDouble3 modelMax = (storedMatrix * udDouble4::create(boundingBoxCenter + boundingBoxExtents, 1.0)).toVector3();
-
-    if (center.x < modelMin.x || center.y < modelMin.y || center.z < modelMin.z || center.x > modelMax.x || center.y > modelMax.y || center.z > modelMax.z)
-      return;
-  }
+  bool hasData = false;
+  vdkPointBufferF64 *pBuffer;
```
OsiJr: Not initialised

src/vcBPA.cpp
```diff
@@ -97,33 +97,39 @@ struct vcBPAGrid
   }
 };
 
+struct vcBPAOctNode
+{
+  udDouble3 center;
+  udDouble3 extents;
+  bool visited;
+
+  static vcBPAOctNode create(udDouble3 center, udDouble3 extents)
```
OsiJr: ProperCase for function names

samsinsane: This is consistent with the rest of the file and udMath, which is where I got the `create` idea from. To my knowledge, this is one of the only places we use this pattern in this repo too, however, I'm willing to change it in a future PR if you want?

src/vcBPA.cpp
samsinsane: I'll fix this up, I had planned for it to be a member in `vcBPAManifold` to avoid the cost of creating and destroying it half a million times, but I managed to avoid calling this function for the leaves of the octree instead. The optimisation could still be made, but having vdkQuery support a "ContainsPoints" instead of having to query for a single point would be better.

samsinsane: This has been fixed.

PR #290, opened 2020-02-18
Added very rough attribute selection system

OsiJr:


builds/assets/lang/enAU.json
```diff
@@ -163,25 +163,25 @@
...
+  "settingsVisClassReserved": "Reserved",
+  "settingsVisClassWireGuard": "Wire Guard / Shield",
+  "settingsVisClassWireConductor": "Wire Conductor / Phase",
+  "settingsVisClassTransmissionTower": "Transmission Tower",
+  "settingsVisClassWireStructureConnector": "Wire Structure Connector",
+  "settingsVisClassBridgeDeck": "Bridge Deck",
+  "settingsVisClassHighNoise": "High Noise",
   "settingsVisClassReservedColours": "19 - 63 Reserved",
   "settingsVisClassReservedLabels": "Reserved",
   "settingsVisClassUserDefinable": "64 - 255 User definable",
```
DamianMadden: Should these numbers still be present?

src/main.cpp
```diff
@@ -1242,6 +1245,93 @@ void vcRenderScene_HandlePicking(vcState *pProgramState, vcRenderData &renderDat
...
+
+              pProgramState->udModelNodeAttributes.Set("%s = '%s'", pHeader->attributes.pDescriptors[i].name, pClassificationName);
+              continue;
+            }
+
+            // Do Stuff
+            switch (pHeader->attributes.pDescriptors[i].typeInfo)
+            {
+            case vdkAttributeTypeInfo_uint8:
+              pProgramState->udModelNodeAttributes.Set("%s = %u", pHeader->attributes.pDescriptors[i].name, *(uint8_t *)pAttributePtr);
```
samsinsane: Not a huge fan of unaligned reads, is it possible to avoid this? Perhaps a template that `memcpy`s into a temporary variable that is then returned?

src/main.cpp
```diff
@@ -1242,6 +1245,93 @@ void vcRenderScene_HandlePicking(vcState *pProgramState, vcRenderData &renderDat
...
+              break;
+            case vdkAttributeTypeInfo_color32:
+              pProgramState->udModelNodeAttributes.Set("%s = \"RGB(%u, %u, %u)\"", pHeader->attributes.pDescriptors[i].name, pAttributePtr[3], pAttributePtr[2], pAttributePtr[1]); //BGRA internally
+              pAttributePtr += 4;
+              break;
+            case vdkAttributeTypeInfo_normal32:
+              pProgramState->udModelNodeAttributes.Set("%s = 'NORMAL'", pHeader->attributes.pDescriptors[i].name);
+              break;
+            case vdkAttributeTypeInfo_vec3f32:
+              pProgramState->udModelNodeAttributes.Set("%s = '[ %f, %f, %f ]'", pHeader->attributes.pDescriptors[i].name, *(float *)&pAttributePtr[0], *(float *)&pAttributePtr[4], *(float *)&pAttributePtr[8]); //BGRA internally
```
samsinsane: You could have wrapped this and achieved more readable code, `((float*)pAttributePtr)[0]`, `[1]`, `[2]`.

src/main.cpp
```diff
@@ -1242,6 +1245,93 @@ void vcRenderScene_HandlePicking(vcState *pProgramState, vcRenderData &renderDat
...
+            case vdkAttributeTypeInfo_float64:
+              pProgramState->udModelNodeAttributes.Set("%s = %f", pHeader->attributes.pDescriptors[i].name, *(double *)pAttributePtr);
+              pAttributePtr += 8;
+              break;
+            case vdkAttributeTypeInfo_color32:
+              pProgramState->udModelNodeAttributes.Set("%s = \"RGB(%u, %u, %u)\"", pHeader->attributes.pDescriptors[i].name, pAttributePtr[3], pAttributePtr[2], pAttributePtr[1]); //BGRA internally
+              pAttributePtr += 4;
+              break;
+            case vdkAttributeTypeInfo_normal32:
+              pProgramState->udModelNodeAttributes.Set("%s = 'NORMAL'", pHeader->attributes.pDescriptors[i].name);
```
peterAdams1014: Is this intended to not actually set data?

src/vcSettingsUI.cpp
```diff
@@ -666,3 +648,44 @@ bool vcSettingsUI_LangCombo(vcState *pProgramState)
...
+      "settingsVisClassWireConductor",
+      "settingsVisClassTransmissionTower",
+      "settingsVisClassWireStructureConnector",
+      "settingsVisClassBridgeDeck",
+      "settingsVisClassHighNoise"
+    };
+
+    return udTempStr("%d. %s", classification, vcString::Get(localizations[classification]));
+  }
+  else if (classification <= 64)
```
peterAdams1014: `else if (classification < 64)` ?

src/main.cpp
OsiJr: Yeah, those formats don't work

src/vcSettingsUI.cpp
OsiJr: Fixed

src/vcSettingsUI.cpp
```diff
@@ -762,80 +803,28 @@ void vcSettingsUI_VisualizationSettings(vcState *pProgramState, vcVisualizationS
...
-      }
-
       ImGui::SameLine();
-      if (ImGui::Button(vcString::Get("settingsVisClassDisableAll")))
-      {
-        for (size_t i = 0; i < udLengthOf(pVisualizationSettings->customClassificationToggles); i++)
-          pVisualizationSettings->customClassificationToggles[i] = false;
-      }
+      if (ImGui::Button(udTempStr("%s##RestoreClassificationColors", vcString::Get("settingsRestoreDefaults"))))
+        memcpy(pProgramState->settings.visualization.customClassificationColors, GeoverseClassificationColours, sizeof(pProgramState->settings.visualization.customClassificationColors));
```
int-Frank: I've seen a lot of `ud*` functions for memory management, but never for `memcpy`. Why is this?

samsinsane: It's just not really necessary. Most other functions are wrapped because either they don't have a consistent API (and in some cases name, see `_udAlloc`) across the various platforms.

src/vcSettingsUI.cpp
```diff
@@ -762,80 +803,28 @@ void vcSettingsUI_VisualizationSettings(vcState *pProgramState, vcVisualizationS
...
 
-      for (size_t i = 0; i < udLengthOf(s_customClassifications); i++)
-      {
-        ImGui::PushID((int)i);
-        ImGui::Checkbox("", &pVisualizationSettings->customClassificationToggles[i]);
-        ImGui::PopID();
-        ImGui::SameLine();
-        vcIGSW_ColorPickerU32(vcString::Get(s_customClassifications[i]), &pVisualizationSettings->customClassificationColors[i], ImGuiColorEditFlags_NoAlpha);
-      }
+      for (uint8_t i = 0; i < 19; ++i)
```
int-Frank: This will break if the size of `customClassificationColors` changes. Maybe use 
```for (uint8_t i = 0; i < UDARRAYSIZE(settings.visualization.customClassificationColors); ++i)``` 
here as well.

File: src/vcSettingsUI.cpp [hunk 8]
```diff
@@ -762,80 +803,28 @@ void vcSettingsUI_VisualizationSettings(vcState *pProgramState, vcVisualizationS
...
       {
-        for (int i = 19; i < 64; ++i)
-        {
-          ImGui::PushID(i);
-          ImGui::Checkbox("", &pVisualizationSettings->customClassificationToggles[i]);
-          ImGui::PopID();
-          ImGui::SameLine();
-          vcIGSW_ColorPickerU32(udTempStr("%d. %s", i, vcString::Get("settingsVisClassReservedLabels")), &pVisualizationSettings->customClassificationColors[i], ImGuiColorEditFlags_NoAlpha);
-        }
+        for (uint8_t i = 19; i < 64; ++i)
```
int-Frank: And here

src/vcSettingsUI.cpp
```diff
@@ -762,80 +803,28 @@ void vcSettingsUI_VisualizationSettings(vcState *pProgramState, vcVisualizationS
...
+          vcIGSW_ColorPickerU32(vcSettingsUI_GetClassificationName(pProgramState, i), &pProgramState->settings.visualization.customClassificationColors[i], ImGuiColorEditFlags_NoAlpha);
         ImGui::TreePop();
       }
 
       if (ImGui::TreeNode(vcString::Get("settingsVisClassUserDefinable")))
       {
-        for (int i = 64; i <= 255; ++i)
+        for (int xi = 64; xi <= 255; ++xi)
```
int-Frank: It doesn't look like `xi` is used anywhere. Should be able to use `for (uint8_t i =...`

src/vcSettingsUI.cpp
```diff
@@ -744,6 +744,47 @@ bool vcSettingsUI_LangCombo(vcState *pProgramState)
   return true;
 }
 
+const char *vcSettingsUI_GetClassificationName(vcState *pProgramState, uint8_t classification)
+{
+  if (classification < 19)
```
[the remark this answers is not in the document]
OsiJr: We don't define these but I've moved the magic numbers to an enum

src/vcSettingsUI.cpp
OsiJr: Can't be, it will overflow.

PR #294, opened 2020-02-19
Added scene window size saving

OsiJr:
[AB#990](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/990)
[AB#991](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/991)

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2241,6 +2241,15 @@ void vcRenderWindow(vcState *pProgramState)
         case vcWL_SceneLeft:
           ImGui::Columns(2);
 
+          if (!pProgramState->settings.presentation.columnSizeCorrect)
+          {
+            ImGui::SetColumnWidth(0, size.x - pProgramState->settings.presentation.sceneExplorerSize);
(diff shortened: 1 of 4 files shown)


src/main.cpp
```diff
@@ -2241,6 +2241,15 @@ void vcRenderWindow(vcState *pProgramState)
         case vcWL_SceneLeft:
           ImGui::Columns(2);
 
+          if (!pProgramState->settings.presentation.columnSizeCorrect)
+          {
+            ImGui::SetColumnWidth(0, size.x - pProgramState->settings.presentation.sceneExplorerSize);
+            pProgramState->settings.presentation.columnSizeCorrect = true;
+          }
+
+          if (ImGui::GetColumnWidth(0) != size.x - pProgramState->settings.presentation.sceneExplorerSize)
```
peterAdams1014: `float` comparison. You sure this will work?

OsiJr: Worst case it does the if every time which does nothing anyway

PR #297, opened 2020-02-19
Screenshots

OsiJr:


src/gl/directx11/vcTexture.cpp
```diff
@@ -470,11 +470,21 @@ bool vcTexture_EndReadPixels(vcTexture *pTexture, uint32_t x, uint32_t y, uint32
...
+  else
   {
-    memcpy((uint8_t*)pPixels + (i * pTexture->width), pPixelData, width * pixelBytes);
-    pPixelData += pTexture->width;
+    //TODO: This block is bugged. Does not do a complete copy.
+
+    pPixelData = ((uint32_t *)msr.pData) + (x + y * pTexture->width);
+    for (int i = 0; i < (int)height; ++i)
+    {
+      memcpy((uint8_t*)pPixels + (i * pTexture->width), pPixelData, width * pixelBytes);
```
samsinsane: `pPixels` is a `void*` unlike `pPixelData` which is a `uint32_t*`, so this add here is probably missing a `4 *`?

src/vcState.h
```diff
@@ -172,7 +172,8 @@ struct vcState
     int height;
 
     vcImageType imageType;
-  } image;
+  } image, screenshot;
+  bool destroyImage;
```
OsiJr: Gets set to false but otherwise not used.

src/vcHotkey.h
```diff
@@ -15,7 +15,7 @@ enum modifierFlags
 
 enum vcBind
 {
-  vcB_Forward,
+  vcB_Forward,
```
samsinsane: Lots of whitespace here?

src/gl/directx11/vcTexture.cpp
OsiJr: You are correct, that fixes it. I'll leave the optimised route anyway- I imagine that will be the much more common case.

src/gl/directx11/vcTexture.cpp
```diff
@@ -470,11 +470,21 @@ bool vcTexture_EndReadPixels(vcTexture *pTexture, uint32_t x, uint32_t y, uint32
...
+  if (x == 0 && y == 0 && width == (uint32_t)pTexture->width && height == (uint32_t)pTexture->height)
   {
-    memcpy((uint8_t*)pPixels + (i * pTexture->width), pPixelData, width * pixelBytes);
-    pPixelData += pTexture->width;
+    pPixelData = (uint32_t*)msr.pData;
+    memcpy((uint8_t*)pPixels, pPixelData, height * width * pixelBytes);
+  }
+  else
+  {
+    //TODO: This block is bugged. Does not do a complete copy.
```
samsinsane: Is this still the case?

OsiJr:
There are lots of issues caused by this- they are all logged in AzDO to fix throughout the next few PRs

PR #311, opened 2020-02-21
updating udCore

int-Frank:


OsiJr:
@int-Frank, you opened this into the wrong master, I've updated to target `master_vdk0.5.1`

PR #315, opened 2020-02-21
Added ability to create an outpath when taking a screenshot

int-Frank:
[AB#1028](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1028)

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -181,6 +181,17 @@ bool vcMain_TakeScreenshot(vcState *pProgramState)
   if (currSize.x - pProgramState->settings.screenshot.resolution.x > 32 || currSize.y - pProgramState->settings.screenshot.resolution.y > 32)
     return true;
 
+  udFindDir* poutputPath(nullptr);
+  udResult result = udOpenDir(&poutputPath, pProgramState->settings.screenshot.outputPath);
+  if (result != udR_Success)
+  {
+    //Folder may not exist, try to create.
+    result = udCreateDir(pProgramState->settings.screenshot.outputPath);
+    if (result != udR_Success)
+      return false;
+  }
+  udCloseDir(&poutputPath);
+
   char buffer[vcMaxPathLength];
   time_t rawtime;
   tm timeval;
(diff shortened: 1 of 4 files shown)


src/main.cpp
```diff
@@ -181,6 +181,17 @@ bool vcMain_TakeScreenshot(vcState *pProgramState)
   if (currSize.x - pProgramState->settings.screenshot.resolution.x > 32 || currSize.y - pProgramState->settings.screenshot.resolution.y > 32)
     return true;
 
+  udFindDir* poutputPath(nullptr);
+  udResult result = udOpenDir(&poutputPath, pProgramState->settings.screenshot.outputPath);
+  if (result != udR_Success)
+  {
+    //Folder may not exist, try to create.
+    result = udCreateDir(pProgramState->settings.screenshot.outputPath);
```
samsinsane: This function is not recursive, if a user were to manually type in a path that contained multiple folders that don't exist this would fail. However, there is a PR pending in udCore for this functionality.

src/main.cpp
```diff
@@ -181,6 +181,17 @@ bool vcMain_TakeScreenshot(vcState *pProgramState)
   if (currSize.x - pProgramState->settings.screenshot.resolution.x > 32 || currSize.y - pProgramState->settings.screenshot.resolution.y > 32)
     return true;
 
+  udFindDir* poutputPath(nullptr);
```
peterAdams1014: This is fine, but we typically don't initialize this way.
`udFindDir* pOutputPath = nullptr;`

int-Frank: Understood

src/main.cpp
int-Frank: I understand. What would you suggest? Wait for this function to come in and redo this item?

samsinsane: I think it's fine to merge as-is - @peterAdams1014 can you get the udCore work prioritised? As far as I'm aware, there's a couple issues with the unit tests, but otherwise it's good to go.

PR #320, opened 2020-02-21
Presentation layer now reponds to 'show UI' control

int-Frank:
[AB#1004](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1004)


int-Frank:
I don't think so. I see no other places where it is used.

PR #322, opened 2020-02-24
Watermark images now display in the proper aspect ratio in the UI

int-Frank:
[AB#843](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/843)

### DIFF
--- a/src/vcSettingsUI.cpp
+++ b/src/vcSettingsUI.cpp
@@ -517,7 +517,29 @@ void vcSettingsUI_Show(vcState *pProgramState)
 
             if (pProgramState->settings.convertdefaults.watermark.pTexture != nullptr)
             {
-              ImGui::Image(pProgramState->settings.convertdefaults.watermark.pTexture, ImVec2(256, 256));
+              //Since we're allowing images of any dimensions, we need to make sure it fits in the UI.
+              int dimension[2] = { pProgramState->settings.convertdefaults.watermark.width,
+                                   pProgramState->settings.convertdefaults.watermark.height };
+
+              int maxDimension[2] = {512, 256};
+              int minDimension = 2;
+
+              for (int i = 0; i < 2; i++)
+              {
+                if (dimension[i] > maxDimension[i])
+                {
+                  float factor = float(dimension[i]) / maxDimension[i];
+                  dimension[i] = int(float(dimension[i]) / factor);
+                  int j = (i + 1) % 2;
+                  dimension[j] = int(float(dimension[j]) / factor);
+
+                  //Is the other dimension too thin now?
+                  if (dimension[j] < minDimension)
+                    dimension[j] = minDimension;
+                }
+              }
+
+              ImGui::Image(pProgramState->settings.convertdefaults.watermark.pTexture, ImVec2(dimension[0], dimension[1]));
 
               if (ImGui::Button(vcString::Get("convertRemoveWatermark")))
               {
--- a/src/vcSettingsUI.cpp
+++ b/src/vcSettingsUI.cpp
@@ -518,20 +518,17 @@ void vcSettingsUI_Show(vcState *pProgramState)
             if (pProgramState->settings.convertdefaults.watermark.pTexture != nullptr)
             {
               //Since we're allowing images of any dimensions, we need to make sure it fits in the UI.
-              int dimension[2] = { pProgramState->settings.convertdefaults.watermark.width,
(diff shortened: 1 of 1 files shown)


src/vcSettingsUI.cpp
```diff
@@ -517,7 +517,29 @@ void vcSettingsUI_Show(vcState *pProgramState)
...
+                if (dimension[i] > maxDimension[i])
+                {
+                  float factor = float(dimension[i]) / maxDimension[i];
+                  dimension[i] = int(float(dimension[i]) / factor);
+                  int j = (i + 1) % 2;
+                  dimension[j] = int(float(dimension[j]) / factor);
+
+                  //Is the other dimension too thin now?
+                  if (dimension[j] < minDimension)
+                    dimension[j] = minDimension;
```
samsinsane: Doesn't this break the aspect ratio? Though I guess anyone with a `512+:1` or `1:256+` ratio "watermark" isn't really going to have a fun time at all.

int-Frank: Yes it does. My intent is to display something, even if the aspect ratio of the image is such that one dimension vanishes after resizing to fit in the UI. Perhaps another solution could be to simply enlarge and clip such images, thus only showing a portion of the watermark, but still retaining the correct aspect ratio?

src/vcSettingsUI.cpp
```diff
@@ -517,7 +517,29 @@ void vcSettingsUI_Show(vcState *pProgramState)
 
             if (pProgramState->settings.convertdefaults.watermark.pTexture != nullptr)
             {
-              ImGui::Image(pProgramState->settings.convertdefaults.watermark.pTexture, ImVec2(256, 256));
+              //Since we're allowing images of any dimensions, we need to make sure it fits in the UI.
+              int dimension[2] = { pProgramState->settings.convertdefaults.watermark.width,
```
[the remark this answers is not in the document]
int-Frank: Righto

src/vcSettingsUI.cpp
```diff
@@ -517,7 +517,29 @@ void vcSettingsUI_Show(vcState *pProgramState)
...
+                                   pProgramState->settings.convertdefaults.watermark.height };
+
+              int maxDimension[2] = {512, 256};
```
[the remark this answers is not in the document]
int-Frank: Righto

src/vcSettingsUI.cpp
```diff
@@ -517,7 +517,29 @@ void vcSettingsUI_Show(vcState *pProgramState)
...
+              int minDimension = 2;
+
+              for (int i = 0; i < 2; i++)
```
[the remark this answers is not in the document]
int-Frank: Good point

src/vcSettingsUI.cpp
```diff
@@ -517,7 +517,29 @@ void vcSettingsUI_Show(vcState *pProgramState)
...
+              {
+                if (dimension[i] > maxDimension[i])
+                {
+                  float factor = float(dimension[i]) / maxDimension[i];
+                  dimension[i] = int(float(dimension[i]) / factor);
+                  int j = (i + 1) % 2;
```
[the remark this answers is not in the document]
int-Frank: Good point

src/vcSettingsUI.cpp
int-Frank: Also note this is only for the preview in settings, not the actual watermark

samsinsane: I'm not too fussed on this - I'm not sure it's a realistic situation anyway. This comment probably didn't need to be made, I just wanted confirmation that the aspect ratio was intentionally broken.

src/vcSettingsUI.cpp
```diff
@@ -517,7 +517,26 @@ void vcSettingsUI_Show(vcState *pProgramState)
 
             if (pProgramState->settings.convertdefaults.watermark.pTexture != nullptr)
             {
-              ImGui::Image(pProgramState->settings.convertdefaults.watermark.pTexture, ImVec2(256, 256));
+              //Since we're allowing images of any dimensions, we need to make sure it fits in the UI.
+              udInt2 dimension = {pProgramState->settings.convertdefaults.watermark.width, pProgramState->settings.convertdefaults.watermark.height};
```
samsinsane: Technically you create these `udInt2::create` but this is fine.

PR #323, opened 2020-02-24
Added missing strings for enAu and zhCn.

asunawu:
Fixed [AB#476](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/476)


asunawu:
Sorry I added the missing strings reported by the bug into the json files only. Should all strings be added to TranslationGuide.md, even if they have been translated?

PR #324, opened 2020-02-24
Limit the maximum customized screenshot resolution from 16384x16384 t…

asunawu:
Fixed [AB#1014](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1014)

### DIFF
--- a/src/vcSettingsUI.cpp
+++ b/src/vcSettingsUI.cpp
@@ -542,6 +542,8 @@ void vcSettingsUI_Show(vcState *pProgramState)
           vcSettingsUI_ShowHeader(pProgramState, vcString::Get("settingsScreenshot"), vcSC_Screenshot);
 
           static udUInt2 ScreenshotResolutions[] = { { 1280, 720 }, { 1920, 1080 }, { 4096, 2160 } };
+          static udUInt2 ScreenshotResolutionsMax = { 4096, 4096 };
+          static udUInt2 ScreenshotResolutionsMin = { 32, 32 };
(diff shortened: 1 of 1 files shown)


src/vcSettingsUI.cpp
```diff
@@ -566,8 +568,10 @@ void vcSettingsUI_Show(vcState *pProgramState)
             ImGui::Indent();
             if (ImGui::InputInt2(vcString::Get("settingsScreenshotResLabel"), (int*)&pProgramState->settings.screenshot.resolution.x))
             {
-              pProgramState->settings.screenshot.resolution.x = udClamp(pProgramState->settings.screenshot.resolution.x, 32u, 16384u);
-              pProgramState->settings.screenshot.resolution.y = udClamp(pProgramState->settings.screenshot.resolution.y, 32u, 16384u);
+              pProgramState->settings.screenshot.resolution.x = udClamp(pProgramState->settings.screenshot.resolution.x
+                , ScreenshotResolutionsMin.x, ScreenshotResolutionsMax.x);
+              pProgramState->settings.screenshot.resolution.y = udClamp(pProgramState->settings.screenshot.resolution.y
+                , ScreenshotResolutionsMin.y, ScreenshotResolutionsMax.y);
```
samsinsane: This isn't really how we do formatting, we don't really break things up onto multiple lines.

Also, I think you could probably change the above to:
```c++
pProgramState->settings.screenshot.resolution = udClamp(pProgramState->settings.screenshot.resolution, screenshotResolutionsMin, ScreenshotResolutionsMax);
```
We have overloads for `udClamp` and friends to handle `udVectorX<T>`, we may as well utilise them.

asunawu: Resolved.

PR #332, opened 2020-02-26
Changed watermark from GUI to mesh.

asunawu:
Fixed [AB#1031](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1031) and there was a watermark on the screenshot image too.

### DIFF
--- a/src/vcRender.cpp
+++ b/src/vcRender.cpp
@@ -1160,26 +1160,14 @@ void vcRenderWatermark(vcState *pProgramState, vcRenderContext *pRenderContext)
   udInt2 imageSize = udInt2::zero();
   vcTexture_GetSize(pProgramState->pSceneWatermark, &imageSize.x, &imageSize.y);
 
-  udDouble3 position = udDouble3::create((double)imageSize.x / pRenderContext->sceneResolution.x - 1,
-    (double)imageSize.y / pRenderContext->sceneResolution.y - 1, 0);
-
-  vcShader_Bind(pRenderContext->watermarkShader.pProgram);
+  udDouble3 position = udDouble3::create((double)imageSize.x / pRenderContext->sceneResolution.x - 1,(double)imageSize.y / pRenderContext->sceneResolution.y - 1, 0);
   pRenderContext->watermarkShader.params.u_worldViewProjectionMatrix = udFloat4x4::create(udDouble4x4::translation(position) * udDouble4x4::scaleUniform(1.0f));
-  pRenderContext->watermarkShader.params.u_screenSize = udFloat4::create((float)imageSize.x / pRenderContext->sceneResolution.x
-    , (float)imageSize.y / pRenderContext->sceneResolution.y
-    , 1.0f, 0.0f);
+  pRenderContext->watermarkShader.params.u_screenSize = udFloat4::create((float)imageSize.x / pRenderContext->sceneResolution.x, (float)imageSize.y / pRenderContext->sceneResolution.y, 1.0f, 0.0f);
   pRenderContext->watermarkShader.params.u_colour = udFloat4::create(1, 1, 1, 1);
 
-  vcShader_BindConstantBuffer(pRenderContext->watermarkShader.pProgram
-    , pRenderContext->watermarkShader.uniform_params
-    , &pRenderContext->watermarkShader.params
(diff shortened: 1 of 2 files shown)


src/vcRender.cpp
```diff
@@ -1130,7 +1149,42 @@ bool vcRender_CreateSelectionBuffer(vcState *pProgramState, vcRenderContext *pRe
   }
   return true;
 }
+void vcRenderWatermark(vcState *pProgramState, vcRenderContext *pRenderContext)
```
samsinsane: There should be an empty line before this line - functions are separated by empty lines.

src/vcRender.cpp
```diff
@@ -1130,7 +1149,42 @@ bool vcRender_CreateSelectionBuffer(vcState *pProgramState, vcRenderContext *pRe
...
+  vcShader_BindTexture(pRenderContext->watermarkShader.pProgram
+    , pProgramState->pSceneWatermark
+    , 0
+    , pRenderContext->watermarkShader.uniform_texture);
+
+  vcMesh_Render(gInternalMeshes[vcInternalMeshType_Billboard]);
+
+  vcGLState_ResetState();
+  vcShader_Bind(nullptr);
+}
```
samsinsane: There should be an empty line after this - same reasoning as above.

src/vcRender.cpp
```diff
@@ -1130,7 +1149,42 @@ bool vcRender_CreateSelectionBuffer(vcState *pProgramState, vcRenderContext *pRe
...
+    return;
 
+  vcGLState_SetDepthStencilMode(vcGLSDM_Always, false);
+  vcGLState_SetFaceMode(vcGLSFM_Solid, vcGLSCM_None);
+
+  udInt2 imageSize = udInt2::zero();
+  vcTexture_GetSize(pProgramState->pSceneWatermark, &imageSize.x, &imageSize.y);
+
+  udDouble3 position = udDouble3::create((double)imageSize.x / pRenderContext->sceneResolution.x - 1,
+    (double)imageSize.y / pRenderContext->sceneResolution.y - 1, 0);
```
samsinsane: These should all be on one line (there's a few more cases below where you do this too).

Also, the udMath types allow you to do something like:
```c++
udDouble3 position = udDouble3::create(imageSize) / pRenderContext->sceneResolution - 1;
```
I'm not 100% certain if the create functions allow for different types - something that we should definitely take care of outside of this PR if they don't.

asunawu: Sorry, but pRenderContext->sceneResolution's type is udUint2, and 1 should be(1,1,0). Do I need to fix it still?

src/vcRender.cpp
asunawu: Fixed.

src/vcRender.cpp
asunawu: Fixed.

src/vcRender.cpp
samsinsane: Ah right, sorry. You can still utilise the various operators and create functions, something like:
```c++
udDouble2 imageSizeD = udDouble2::create(imageSize);
udDouble2 sceneResD = udDouble2::create(pRenderContext->sceneResolution);
udDouble3 position = udDouble3::create(imageSizeD / sceneResD - 1, 0);
```
You can also get rid of the local variables, but I think you have something similar below too?

asunawu: udDouble3::create(imageSizeD / sceneResD - 1, 0) won't work.

samsinsane: Is it because the constants aren't doubles, or have I forgotten something?

samsinsane: Oh I see, I'm sorry, we have the `*` and `/` operators for a scalar, but not `+` and `-`, I think you would just need to change `1` to `udDouble2::one()`

asunawu: Do you think it's really worth to do that? The requested change will will add at least 8 additional 4-byte copy operations. I have changed the code into one line for easier reading.

samsinsane: I was only considering the readability of the code, but you raise a very good point that I was not even thinking about - this is a very bad place to add unnecessary work, oops. I'm going to resolve this discussion. 👍

PR #333, opened 2020-02-26
Clamped imageRescaleDistance to ImageRescaleMin and ImageRescaleMax

LukeFrancoisSchoen:
[AB#1042](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1042)

### DIFF
--- a/src/vcSettingsUI.cpp
+++ b/src/vcSettingsUI.cpp
@@ -132,6 +132,7 @@ void vcSettingsUI_Show(vcState *pProgramState)
           if (ImGui::SliderFloat(vcString::Get("settingsAppearancePOIDistance"), &pProgramState->settings.presentation.POIFadeDistance, vcSL_POIFaderMin, vcSL_POIFaderMax, "%.3fm", 3.f))
             pProgramState->settings.presentation.POIFadeDistance = udClamp(pProgramState->settings.presentation.POIFadeDistance, vcSL_POIFaderMin, vcSL_GlobalLimitf);
           ImGui::SliderFloat(vcString::Get("settingsAppearanceImageRescale"), &pProgramState->settings.presentation.imageRescaleDistance, vcSL_ImageRescaleMin, vcSL_ImageRescaleMax, "%.3fm", 3.f);
+          pProgramState->settings.presentation.imageRescaleDistance = udClamp(pProgramState->settings.presentation.imageRescaleDistance, vcSL_ImageRescaleMin, vcSL_ImageRescaleMax);
           ImGui::Checkbox(vcString::Get("settingsAppearanceShowDiagnostics"), &pProgramState->settings.presentation.showDiagnosticInfo);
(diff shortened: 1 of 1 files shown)


src/vcSettingsUI.cpp
```diff
@@ -132,6 +132,7 @@ void vcSettingsUI_Show(vcState *pProgramState)
           if (ImGui::SliderFloat(vcString::Get("settingsAppearancePOIDistance"), &pProgramState->settings.presentation.POIFadeDistance, vcSL_POIFaderMin, vcSL_POIFaderMax, "%.3fm", 3.f))
             pProgramState->settings.presentation.POIFadeDistance = udClamp(pProgramState->settings.presentation.POIFadeDistance, vcSL_POIFaderMin, vcSL_GlobalLimitf);
           ImGui::SliderFloat(vcString::Get("settingsAppearanceImageRescale"), &pProgramState->settings.presentation.imageRescaleDistance, vcSL_ImageRescaleMin, vcSL_ImageRescaleMax, "%.3fm", 3.f);
+          pProgramState->settings.presentation.imageRescaleDistance = udClamp(pProgramState->settings.presentation.imageRescaleDistance, vcSL_ImageRescaleMin, vcSL_ImageRescaleMax);
```
samsinsane: This seems inconsistent with the slider above (which uses an if statement to perform the clamp), are we intending to have different functionality between the two?

LukeFrancoisSchoen: Hey Sam, i noticed the line above when i was writing the code, however i decided to do it this way, the functionality is effectively the same, the above if is simply ensures that the imgui sliderDraw() function result has returned true before applying the clamping (True is returned when imgui detects that the slider has been updated this frame), this potentially avoids needing to perform the clamp (slight performance benefit) however it relies on imgui always correctly returning true under all circumstances where in the slider may have been changed.
I considered removing the if from the line above but figured since i was making such a small change it was unnecessary to apply cleanup, if your convinced the clamp is too expensive and that sliderDraw() will always returns true in the relevant cases then i would be happy to include an if on this new clamp.

samsinsane: My main concern was consistency but I don't know about how ImGui works with the return value, to my knowledge using Ctrl+Click to override the value doesn't trigger it but I could be wrong here. @OsiJr @peterAdams1014 can you guys weigh in on how this works?

peterAdams1014: Ctrl + Click will trigger this. I haven't seen any evidence to suggest ImGui doesn't work as intended (return true always on any change). 

IMO just keep it consistent.

LukeFrancoisSchoen: Did a quick test looking frame by frame, Ctrl + Click reliably triggers return true so i've included the if, lets assume for now there is no other way to affect scroll bars.

PR #336, opened 2020-02-26
Export LAS and UDS models

OsiJr:
[AB#205](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/205)

### DIFF
--- a/src/scene/vcModel.cpp
+++ b/src/scene/vcModel.cpp
@@ -5,6 +5,7 @@
 #include "vcBPA.h"
 #include "vcStringFormat.h"
 #include "vcModals.h"
+#include "vcQueryNode.h"
 
 #include "gl/vcTexture.h"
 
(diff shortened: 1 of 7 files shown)


src/scene/vcModel.cpp
```diff
@@ -474,16 +476,53 @@ void vcModel::HandleContextMenu(vcState *pProgramState)
...
+      }
+
+      vcIGSW_FilePicker(pProgramState, vcString::Get("sceneExplorerExportFilename"), pProgramState->modelPath, SupportedTileTypes_QueryExport, vcFDT_SaveFile, nullptr);
+
+      if (ImGui::Button(vcString::Get("sceneExplorerExportBegin")))
+      {
+        if (udFileExists(pProgramState->modelPath) != udR_Success || vcModals_OverwriteExistingFile(pProgramState->modelPath))
+        {
+          vdkQueryFilter *pFilter = ((s_pQuery == nullptr) ? nullptr : s_pQuery->m_pFilter);
+          vdkPointCloud *pCloud = m_pPointCloud;
```
samsinsane: Is this required for the capture? I would have thought you could just use `m_pPointCloud` directly? Also, if you remove the model from the scene during an export, does this trigger a crash?

PR #338, opened 2020-02-26
Settings Window reordering when language selection has changed 

LukeFrancoisSchoen:
[AB#1002](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1002) Added code to keep focus on settings window when the language combo box selection has changed.

### DIFF
--- a/src/vcSettingsUI.cpp
+++ b/src/vcSettingsUI.cpp
@@ -109,7 +109,14 @@ void vcSettingsUI_Show(vcState *pProgramState)
         {
           vcSettingsUI_ShowHeader(pProgramState, vcString::Get("settingsAppearance"), vcSC_Appearance);
 
-          vcSettingsUI_LangCombo(pProgramState);
+          bool languageWasJustChanged = false;
+          static bool languageWasChangedLastFrame = false;
+          vcSettingsUI_LangCombo(pProgramState, &languageWasJustChanged);
+          // Reset focus to the settings window (fixes a window ordering bug when language is changed for the first time)
+          if (languageWasChangedLastFrame)
+            ImGui::SetWindowFocus();
+          languageWasChangedLastFrame = languageWasJustChanged;
+
           ImGui::SameLine();
           ImGui::TextUnformatted(vcString::Get("settingsAppearanceLanguage"));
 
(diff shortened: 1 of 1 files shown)


LukeFrancoisSchoen:
It looks like my VS text auto formatter fixed a few formatting side issues, @here should those be a separate PR ?

LukeFrancoisSchoen:
Also my solution here is pretty nasty i wonder if there is a better way, @here who knows the most about how the language switching system works ?

src/vcSettingsUI.cpp
```diff
@@ -109,7 +109,14 @@ void vcSettingsUI_Show(vcState *pProgramState)
...
           vcSettingsUI_ShowHeader(pProgramState, vcString::Get("settingsAppearance"), vcSC_Appearance);
 
-          vcSettingsUI_LangCombo(pProgramState);
+          bool languageWasJustChanged = false;
+          static bool languageWasChangedLastFrame = false;
+          vcSettingsUI_LangCombo(pProgramState, &languageWasJustChanged);
+          // Reset focus to the settings window (fixes a window ordering bug when language is changed for the first time)
+          if (languageWasChangedLastFrame)
+            ImGui::SetWindowFocus();
+          languageWasChangedLastFrame = languageWasJustChanged;
```
samsinsane: This would be a bit cleaner:
```c++
static bool languageWasChanged = false;
if (languageWasChanged) // Needs to be off by one frame
  ImGui::SetWindowFocus();

vcSettingsUI_LangCombo(pProgramState, &languageWasChanged);
```

src/vcSettingsUI.cpp
```diff
@@ -474,7 +481,7 @@ void vcSettingsUI_Show(vcState *pProgramState)
           vcIGSW_FilePicker(pProgramState, vcString::Get("convertTempDirectory"), pProgramState->settings.convertdefaults.tempDirectory, udLengthOf(pProgramState->settings.convertdefaults.tempDirectory), nullptr, 0, vcFDT_SelectDirectory, [pProgramState] {
             // Nothing needs to happen here
             udUnused(pProgramState);
-          });
+            });
```
samsinsane: You should change your VS settings for this, I think it's: `Options` -> `Text Editor` -> `C/C++` -> `Formatting` -> `Indentation` -> `In existing code, use the setting for alignment of new lines within parentheses`. Best way to test this is to reduce the indentation on this line and then apply formatting to the file - I'm fairly certain that triggers it, I remember it being kind of weird tracking down this setting.

src/vcSettingsUI.cpp
samsinsane: Oops, forgot to assign back to false after the if.

samsinsane:
This also doesn't build because you didn't change the forward declaration, how did this work when you were testing?

LukeFrancoisSchoen:
> 
> 
> This also doesn't build because you didn't change the forward declaration, how did this work when you were testing?

LukeFrancoisSchoen:
> 
> 
> This also doesn't build because you didn't change the forward declaration, how did this work when you were testing?

Not certain how but it does.

LukeFrancoisSchoen:
Woopsies i closed this one by accident while trying to leave a comment, anyway I'll try Sams ideas tomorrow and reopen then.

src/vcSettingsUI.cpp
LukeFrancoisSchoen: ta!

PR #342, opened 2020-02-27
Fhart/fixslpk

int-Frank:
[AB#421](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/421) Here is an immediate fix for vertex/index count overflow. Standardising all vertex/index count values to uint32_t is taking a little longer than I thought.


int-Frank:
The two offending functions `vcMesh_Create()` and `vcPolygonModel_CreateFromRawVertexData()` have been updated to require vertex and index counts as `uint32_t`. Furthermore 'vcPolygonModel_CreateFromRawVertexData()' now supports index data as 16 and 32 bit ints, as opposed to just 16 bit previously. Also note `vcFenceSegment` and `vcWaterVolume` still store vertex counts as int and may need to be updated in the future to align with the current convention of using 32 bit uints to store vertex/index counts.

src/rendering/vcPolygonModel.cpp
```diff
@@ -144,7 +145,12 @@ udResult vcPolygonModel_CreateFromRawVertexData(vcPolygonModel **ppPolygonModel,
   if (pPolygonModel->pMeshes[0].materialID == vcPMST_Count)
     UD_ERROR_SET(udR_Unsupported);
 
-  UD_ERROR_CHECK(vcMesh_Create(&pPolygonModel->pMeshes[0].pMesh, pMeshLayout, totalTypes, pVerts, pPolygonModel->pMeshes[0].numVertices, pIndices, indexCount, (pIndices == nullptr) ? vcMF_NoIndexBuffer : vcMF_IndexShort));
+  if (pIndices == nullptr)
+    meshFlags = meshFlags | vcMF_NoIndexBuffer;
+  else if (flags & vcPMF_IndexShort)
+    meshFlags = meshFlags | vcMF_IndexShort;
```
peterAdams1014: Could auto-detect this flag (when `vertCount <= 65535`)? Then `vcPolyModelFlags` could be removed.

Thoughts?

int-Frank: This flag tells us the format of `pIndices`. We're passing in our index data as a `void *` so we need a flag to let us know how to interpret this data.

int-Frank: @peterAdams1014 Just to update, I removed the flags and set `vcPolygonModel_CreateFromRawVertexData()` to accept index data as `uint32_t`

int-Frank:
Tested in OpenGL and DirectX builds

samsinsane:
@int-Frank Can you rebase this on master or merge master in? You have a conflict that needs to be resolved.

PR #345, opened 2020-02-28
creating VSM poly meshes back on main thread

int-Frank:
[AB#1079](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1079)

While fixing another another bug I came across this crash. When loading a VSM, we were trying to create the mesh on a worker thread. This caused a crash in (at least) OpenGL as we are making OpenGL calls on a thread that does not own the OpenGL context. One possible solution would be to transfer ownership of the context to the thread before we make our gl calls. I decided against this however as we create all other meshes on the main thread. I am not sure why VSMs are treated differently.

### DIFF
--- a/src/rendering/vcPolygonModel.cpp
+++ b/src/rendering/vcPolygonModel.cpp
@@ -84,9 +84,9 @@ struct AsyncPolygonModelLoadMeshInfo
   vcPolygonModel *pPolygonModel;
   vcPolygonModelMesh *pMesh;
   vcP3N3UV2Vertex *pVerts;
-  const vcVertexLayoutTypes* pMeshLayout;
+  const vcVertexLayoutTypes *pMeshLayout;
   int totalTypes;
-  void* pIndices;
+  void *pIndices;
   uint32_t currentIndices;
   vcMeshFlags flags;
 };
@@ -162,7 +162,7 @@ epilogue:
   return result;
 }
 
-udResult vcPolygonModel_CreateFromVSMFInMemory(vcPolygonModel **ppModel, char *pData, int dataLength, udWorkerPool* pWorkerPool)
+udResult vcPolygonModel_CreateFromVSMFInMemory(vcPolygonModel **ppModel, char *pData, int dataLength, udWorkerPool *pWorkerPool)
 {
   if (pData == nullptr || (size_t)dataLength < sizeof(VSMFHeader))
     return udR_InvalidParameter_;
@@ -228,9 +228,9 @@ udResult vcPolygonModel_CreateFromVSMFInMemory(vcPolygonModel **ppModel, char *p
 
     // override material id for now
     pNewModel->pMeshes[i].materialID = vcPMST_P3N3UV2_Opaque;
-
+    
     size_t vertArraySize = sizeof(vcP3N3UV2Vertex) * pNewModel->pMeshes[i].numVertices;
-    vcP3N3UV2Vertex* pVerts = (vcP3N3UV2Vertex*)pFilePos;
+    vcP3N3UV2Vertex *pVerts = (vcP3N3UV2Vertex*)pFilePos;
     pFilePos += vertArraySize;
 
     // TODO: Assume these all need flipping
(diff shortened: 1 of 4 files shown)


src/scene/vcPolyModelNode.cpp
```diff
@@ -46,8 +46,7 @@ vcPolyModelNode::vcPolyModelNode(vdkProject *pProject, vdkProjectNode *pNode, vc
     pLoadInfo->pNode = this;
     pLoadInfo->pProgramState = pProgramState;
 
-    // Queue for load
-    udWorkerPool_AddTask(pProgramState->pWorkerPool, vcPolyModelNode_LoadModel, pLoadInfo, true);
+    vcPolyModelNode_LoadModel(pLoadInfo);
```
samsinsane: I think this leaks memory now - the final parameter was `true` which is to `clearMemory` which means it will call `udFree` on the user data, `pLoadInfo`. You could have alternatively changed the call to `udWorkerPool_AddTask` to use the `postFunction` instead of the `func` parameter for `vcPolyModelNode_LoadModel`, this would keep the work done via the worker pool, but will perform the work on the main thread instead.

@OsiJr Is this a case where using the worker pool makes sense?

OsiJr: The memory is freed after the post- the model creation should definately be done in the post function. Some VSM files take >10s to load so can't happen on the main thread.

int-Frank: I see. It looks like the solution will be to break up `vcPolyModelNode_LoadModel` into worker-thread-work and main-thread-work, using both `func` and `postFunction`. Thoughts?

int-Frank: Righto I found the issue. See latest

src/rendering/vcPolygonModel.cpp
```diff
@@ -245,15 +253,34 @@ udResult vcPolygonModel_CreateFromVSMFInMemory(vcPolygonModel **ppModel, char *p
...
       pFilePos += textureFileSize;
     }
 
-    UD_ERROR_CHECK(vcMesh_Create(&pNewModel->pMeshes[i].pMesh, vcP3N3UV2VertexLayout, (int)udLengthOf(vcP3N3UV2VertexLayout), pVerts, pNewModel->pMeshes[i].numVertices, pIndices, pNewModel->pMeshes[i].numElements, vcMF_IndexShort));
+    AsyncPolygonModelLoadMeshInfo* pLoadInfo = udAllocType(AsyncPolygonModelLoadMeshInfo, 1, udAF_Zero);
+    UD_ERROR_NULL(pLoadInfo, udR_MemoryAllocationFailure);
+
+    pLoadInfo->pPolygonModel = pNewModel;
+    pLoadInfo->pMesh = &pNewModel->pMeshes[i];
+    pLoadInfo->pVerts = (vcP3N3UV2Vertex*)udAlloc(vertArraySize);
```
samsinsane: `udAllocType`? And below for the indices. Also, you could probably use `udMemdup` instead?

src/rendering/vcPolygonModel.cpp
```diff
@@ -84,6 +84,11 @@ struct AsyncPolygonModelLoadMeshInfo
   vcPolygonModel *pPolygonModel;
   vcPolygonModelMesh *pMesh;
   vcP3N3UV2Vertex *pVerts;
+  const vcVertexLayoutTypes* pMeshLayout;
```
samsinsane: The `*` goes on the right, this seems to be done throughout this PR, you might need to update your VS settings for this.

src/rendering/vcPolygonModel.cpp
int-Frank: Ok

src/rendering/vcPolygonModel.cpp
int-Frank: Sure, changes are in

int-Frank:
@peterAdams1014  @samsinsane  I have tested the fix on both opengl and directX builds

PR #347, opened 2020-03-02
VSM focus issue fix

int-Frank:
[AB#515](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/515)

Previously it seemed we required scene items to have a non-zero `WorldSpacePivot()` in order to be brought into focus. I'm not sure why this is. The world space pivot is built off the item's local space pivot, which can indeed be it's origin - the zero-vector. And if the the item's world space matrix has zero translation, the world space pivot will again remain at the origin.

I believe this check exists for the case where it does not make sense to focus on that item, such as a folder. If this check is important, we might need a `Is3DSceneItem()` method to determine this explicitly.

### DIFF
--- a/src/scene/vcFolder.cpp
+++ b/src/scene/vcFolder.cpp
@@ -315,7 +315,7 @@ void vcFolder::HandleImGui(vcState *pProgramState, size_t *pItemID)
         ImGui::EndPopup();
       }
 
-      if (pNode->itemtype != vdkPNT_Folder /*&& pSceneItem->GetWorldSpacePivot() != udDouble3::zero()*/ && ImGui::IsMouseDoubleClicked(0) && ImGui::IsItemHovered())
+      if (pSceneItem->Is3DSceneObject() && ImGui::IsMouseDoubleClicked(0) && ImGui::IsItemHovered())
       {
         if (pSceneItem->m_pPreferredProjection != nullptr)
           vcProject_UseProjectionFromItem(pProgramState, pSceneItem);
--- a/src/scene/vcFolder.cpp
+++ b/src/scene/vcFolder.cpp
@@ -315,7 +315,7 @@ void vcFolder::HandleImGui(vcState *pProgramState, size_t *pItemID)
         ImGui::EndPopup();
       }
 
-      if (pNode->itemtype != vdkPNT_Folder && pSceneItem->GetWorldSpacePivot() != udDouble3::zero() && ImGui::IsMouseDoubleClicked(0) && ImGui::IsItemHovered())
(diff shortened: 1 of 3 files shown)


src/scene/vcFolder.cpp
```diff
@@ -315,7 +315,7 @@ void vcFolder::HandleImGui(vcState *pProgramState, size_t *pItemID)
         ImGui::EndPopup();
       }
 
-      if (pNode->itemtype != vdkPNT_Folder && pSceneItem->GetWorldSpacePivot() != udDouble3::zero() && ImGui::IsMouseDoubleClicked(0) && ImGui::IsItemHovered())
+      if (pNode->itemtype != vdkPNT_Folder /*&& pSceneItem->GetWorldSpacePivot() != udDouble3::zero()*/ && ImGui::IsMouseDoubleClicked(0) && ImGui::IsItemHovered())
```
OsiJr: I think this should just check if the vcSLS is loaded?

int-Frank: Perhaps we need to check this anyway, although I think the issue comes from the fact that every item derived from vcSceneItem has a World matrix and a pivot; even items where this doesn't really make sense, folders for instance. That would suggest it's in the '3D' scene and can potentially interact with other objects in the scene.

DamianMadden: Possible fix is a deferred action... seems quite often my solutions come in the form of deferred actions, perhaps someone should come up with a framework for deferring actions by either frame count, time delta or a function pointer that you'd pass in. the vcSLS loaded check addresses the immediate issue but I expect QA will then log that their expectations still aren't met

DamianMadden: not sure, was this comment correct? 
```
Damian Madden
commented Nov 5, 2019
 
@Paul Fox To clarify, this is the case regardless of a UDS being in the scene or not, and because VSM files aren't located the position is equal to udDouble3::zero, which is a sentinel for the scene item being not loaded or otherwise in a state from which no world space pivot can be determined... I could fix by changing the default position to non-zero, possibly loading a VSM at the camera position? or immediately in front of the camera
```
A quick look turned up nothing that sets it for a new model

DamianMadden: Right so I think the solution will be choosing some non-zero origin for vcPolygonModel_CreateFromVSMFInMemory & vcPolygonModel_CreateFromRawVertexData

DamianMadden: or some other sentinel for unset polygonmodels

int-Frank: Ok added a more elegant solution.

PR #349, opened 2020-03-02
Added http code enums, and expanded http code support in vcWebFile.cpp

int-Frank:
[AB#187](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/187)

### DIFF
--- a/src/vcHTTP.cpp
+++ b/src/vcHTTP.cpp
@@ -12,6 +12,7 @@ udResult vcHTTP_StatusToudResult(vcHTTPStatus statusCode)
   case vcHTTP_ProxyAuthenticationRequired: return udR_AuthError;
   case vcHTTP_RequestTimeout: return udR_SessionExpired;
   case vcHTTP_ServiceUnavailable: return udR_Pending;
+  default: break; //Nothing
   }
 
   if (statusCode >= 500 && statusCode <= 599)
--- a/src/vcHTTP.cpp
+++ b/src/vcHTTP.cpp
@@ -1,0 +1,317 @@
+#include "vcHTTP.h"
+#include "udStringUtil.h"
+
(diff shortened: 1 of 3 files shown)


src/vcHTTP.cpp
```diff
@@ -0,0 +1,318 @@
...
+  case vcHTTP_ServerError:
+    return "Internal Server Error";
+  case vcHTTP_NotImplemented:
+    return "Not Implemented";
+  case vcHTTP_ServiceUnavailable:
+    return "Service Unavailable";
+  case vcHTTP_BadHTTPVersion:
+    return "HTTP Version not supported";
+  default:
+    return "Internal Server Error";
```
samsinsane: We shouldn't have a `default` (new entries in the `vcHTTP` enum should be handled correctly) and the fallback string shouldn't be an internal server error - only the 5xx range are server errors. A better fallback would be "unknown HTTP status code" or something.

src/vcHTTP.cpp
```diff
@@ -0,0 +1,318 @@
...
+    vcHTTP_SplitAndInject(&pHeaders->cookies, pCookieStr, ';', '=', true);
+
+    udFree(pCookieStr);
+  }
+
+  if (!pHeaders->tokens.Get("Cookie").IsObject())
+    pHeaders->tokens.Set("Cookie = {}");
+}
+
+const char* vcHTTP_StatusAsText(vcHTTPStatus status)
```
samsinsane: @OsiJr This function isn't used and I've made two comments against it, should vcHTTP contain so much unused functionality?

int-Frank:
@samsinsane Righto

OsiJr:
Closing this PR due to decay. It should be rebased and refreshed before reopening.

PR #351, opened 2020-03-02
To collapse scene explorer quickly by [F4]/new menu button.

asunawu:
Resolved [AB#322](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/322).

### DIFF
--- a/src/imgui_ex/vcMenuButtons.h
+++ b/src/imgui_ex/vcMenuButtons.h
@@ -39,8 +39,9 @@ enum vcMenuBarButtonIcon
 
   vcMBBI_FullScreen = 33,
   vcMBBI_MapMode = 34,
+  vcMBBI_ToggleScreenExplorer = 35,
 
-  //Reserved = 35 - 39
+  //Reserved = 36 - 39
 
   vcMBBI_Crosshair = 40,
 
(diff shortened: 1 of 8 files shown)


src/imgui_ex/vcMenuButtons.h
```diff
@@ -39,8 +39,9 @@ enum vcMenuBarButtonIcon
 
   vcMBBI_FullScreen = 33,
   vcMBBI_MapMode = 34,
+  vcMBBI_ToggleScreenExplorer = 35,
```
OsiJr: Is there actually a new button for this?

src/main.cpp
```diff
@@ -2320,18 +2327,24 @@ void vcRenderWindow(vcState *pProgramState)
       {
         ImGui::PopStyleVar();
 
+        int sceneExplorerSize = pProgramState->settings.presentation.sceneExplorerSize;
+        if (pProgramState->sceneExplorerCollapsed)
+        {
+          sceneExplorerSize = 0;
+        }
```
OsiJr: Unnecessary braces

asunawu: Braces Removed. And I think this is a quick UI operation, so the collapsed state is not saved. I added a blank button which still needs an icon.

PR #352, opened 2020-03-03
Improvements and Changes to Tile Renderer

peterAdams1014:
Fixed Tile Renderer to work with LogZ.
Preparing the Tile Renderer for globe rendering:
 - Removed stencilling, don't "always render root" etc.
 - Changed descend heuristics.
Tile tessellation is now independent of map colour (ancestors colour can be draped onto tiles).
Tiles are now 3x3 resolution.

### DIFF
--- a/src/rendering/vcTileRenderer.cpp
+++ b/src/rendering/vcTileRenderer.cpp
@@ -23,7 +23,7 @@
 
 enum
 {
-  TileVertexResolution = 2, // This should match GPU struct size
+  TileVertexResolution = 3, // This should match GPU struct size
   TileIndexResolution = (TileVertexResolution - 1),
 
   MaxTextureUploadsPerFrame = 3,
@@ -66,6 +66,7 @@ struct vcTileRenderer
       udFloat4x4 projectionMatrix;
       udFloat4 eyePositions[TileVertexResolution * TileVertexResolution];
       udFloat4 colour;
+      udFloat4 uvOffsetScale;
     } everyObject;
   } presentShader;
 };
@@ -161,13 +162,6 @@ uint32_t vcTileRenderer_LoadThread(void *pThreadData)
 
         tileCenter = udDouble3::create(pNode->renderInfo.center, pRenderer->pSettings->maptiles.mapHeight);
         double distanceToCameraSqr = udMagSq3(tileCenter - pRenderer->cameraPosition);
-
-        // root (special case)
-        if (pNode == &pRenderer->quadTree.nodes.pPool[pRenderer->quadTree.rootIndex])
-        {
-          best = i;
-          break;
-        }
 
         bool betterNode = true;
         if (best != -1)
@@ -431,8 +425,8 @@ bool vcTileRenderer_UpdateTileTexture(vcTileRenderer *pTileRenderer, vcQuadTreeN
     pNode->renderInfo.pTexture = nullptr;
     pNode->renderInfo.timeoutTime = pTileRenderer->totalTime;
 
-    udDouble2 min = udDouble2::create(pNode->worldBounds[0].x, pNode->worldBounds[2].y);
-    udDouble2 max = udDouble2::create(pNode->worldBounds[3].x, pNode->worldBounds[1].y);
+    udDouble2 min = udDouble2::create(pNode->worldBounds[0].x, pNode->worldBounds[6].y);
+    udDouble2 max = udDouble2::create(pNode->worldBounds[8].x, pNode->worldBounds[2].y);
     pNode->renderInfo.center = (max + min) * 0.5;
 
     pTileCache->tileLoadList.PushBack(pNode);
@@ -485,9 +479,6 @@ void vcTileRenderer_UpdateTextureQueues(vcTileRenderer *pTileRenderer)
 
   // update visible tiles textures
   vcTileRenderer_UpdateTextureQueuesRecursive(pTileRenderer, &pTileRenderer->quadTree.nodes.pPool[pTileRenderer->quadTree.rootIndex], tileUploadCount);
-
-  // always request root
-  vcTileRenderer_UpdateTileTexture(pTileRenderer, &pTileRenderer->quadTree.nodes.pPool[pTileRenderer->quadTree.rootIndex]);
 
   // remove from the queue any tiles that are invalid
   for (int i = 0; i < (int)pTileRenderer->cache.tileLoadList.length; ++i)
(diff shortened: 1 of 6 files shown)


src/rendering/vcTileRenderer.cpp
```diff
@@ -616,44 +585,59 @@ void vcTileRenderer_RecursiveSetRendered(vcTileRenderer *pTileRenderer, vcQuadTr
...
+    int slippyRange = (int)udPow(2.0f, (float)depthDiff);
+
+    // top-left, and bottom-right corners
+    udInt2 slippy0 = pNode->slippyPosition.toVector2();
+    udInt2 slippy1 = slippy0 + udInt2::one();
 
-    if (!vcTileRenderer_NodeHasValidBounds(pNode))
-      return false;
+    udInt2 ancestorSlippyLocal0 = udInt2::create(pBestTexturedAncestor->slippyPosition.x * slippyRange, pBestTexturedAncestor->slippyPosition.y * slippyRange);
+    udInt2 ancestorSlippyLocal1 = udInt2::create((pBestTexturedAncestor->slippyPosition.x + 1) * slippyRange, (pBestTexturedAncestor->slippyPosition.y + 1) * slippyRange);
```
samsinsane: Isn't this just `ancestorSlippyLocal0 + slippyRange`?

src/rendering/vcTileRenderer.cpp
```diff
@@ -616,44 +585,59 @@ void vcTileRenderer_RecursiveSetRendered(vcTileRenderer *pTileRenderer, vcQuadTr
...
+    udFloat2 boundsRange = udFloat2::create(ancestorSlippyLocal1 - ancestorSlippyLocal0);
```
samsinsane: Won't this just be `udFloat2::create(slippyRange)`?

src/rendering/vcTileRenderer.cpp
```diff
@@ -701,33 +680,29 @@ void vcTileRenderer_Render(vcTileRenderer *pTileRenderer, const udDouble4x4 &vie
...
 
   vcGLState_SetViewportDepthRange(0.0f, 1.0f);
   vcGLState_SetDepthStencilMode(vcGLSDM_LessOrEqual, true, nullptr);
   vcGLState_SetBlendMode(vcGLSBM_None);
   vcShader_Bind(nullptr);
 
-#if VISUALIZE_DEBUG_TILES
-  printf("touched=%d, visible=%d, rendered=%d, leaves=%d\n", pTileRenderer->quadTree.metaData.nodeTouchedCount, pTileRenderer->quadTree.metaData.visibleNodeCount, pTileRenderer->quadTree.metaData.nodeRenderCount, pTileRenderer->quadTree.metaData.leafNodeCount);
-#endif
+  // debugging
```
samsinsane: This could have been reverted? Except the changes to the `printf` itself, but this seems like it still fits the `VISUALIZE_DEBUG_TILES` define? Alternatively, it could be replaced with `#if 0` and turned on manually when required. I can appreciate that this exists out of convenience when debugging the system.

src/vcQuadTree.cpp
```diff
@@ -159,12 +160,13 @@ void vcQuadTree_InitNode(vcQuadTree *pQuadTree, uint32_t slotIndex, const udInt3
...
-  return -1 < vcQuadTree_FrustumTest(pQuadTree->frustumPlanes, udDouble3::create(pNode->tileCenter, 0.0), udDouble3::create(pNode->tileExtents, 0.0));
+  return -1 < vcQuadTree_FrustumTest(pQuadTree->frustumPlanes, udDouble3::create(pNode->tileCenter, pQuadTree->quadTreeHeightOffset), udDouble3::create(pNode->tileExtents, 0.0));
 }
 
-inline bool vcQuadTree_ShouldSubdivide(vcQuadTree *pQuadTree, double distanceMS, int depth)
+inline bool vcQuadTree_ShouldSubdivide(double distance, int depth)
 {
-  return distanceMS < (1.0 / (1 << (depth + vcTRMQToDepthModifiers[pQuadTree->pSettings->maptiles.mapQuality])));
+  // trial and error'd this heuristic
+  return distance < (10000000 >> depth);
```
samsinsane: Could this value be given a meaningful name? In the interest of progressing our coding standard forward, perhaps a `constexpr int name = 10000000;` locally to this function? @OsiJr @peterAdams1014 thoughts on using `constexpr` for one offs or local values instead of `enum`?

src/vcQuadTree.cpp
```diff
@@ -452,37 +439,30 @@ bool vcQuadTree_ShouldFreeBlock(vcQuadTree *pQuadTree, uint32_t blockIndex)
...
       {
         vcQuadTreeNode *pParentNode = &pQuadTree->nodes.pPool[parentIndex];
         if (pParentNode->touched)
         {
-          // We have an ancestor that has no texture, or is fading
+          // We have an ancestor that has no texture
           if (!pParentNode->renderInfo.pTexture)
             return false;
-
+
```
samsinsane: Unnecessary whitespace on this line and the line below.

src/vcQuadTree.h
```diff
@@ -28,6 +28,10 @@ struct vcNodeRenderInfo
   int32_t width, height;
   void *pData;
 
+  vcTexture *pDrawTexture; // may not own this
```
samsinsane: How do you know if you do own it or not? Does this comment still apply?

src/vcQuadTree.h
```diff
@@ -48,7 +52,7 @@ struct vcQuadTreeNode
   bool rendered;
 
   // cached
-  udDouble2 worldBounds[4]; // corners
+  udDouble2 worldBounds[9]; // corners
```
samsinsane: Does this comment still apply? Just want to confirm that they're still _corners_.

src/vcRender.cpp
```diff
@@ -687,13 +687,15 @@ void vcRenderTerrain(vcState *pProgramState, vcRenderContext *pRenderContext)
 
     int currentZoom = 21;
 
-    double farPlane = pProgramState->settings.camera.farPlane;
+    // This technique won't work with ECEF
+    // This 'value' was trial and errored. 
+    double visibleFarPlane = 1000.0 + udAbs(localCamPos.z) * 150.0;
```
samsinsane: Similar to above `constexpr` comment, if both `1000.0` and `150.0` can be given meaningful names, it would be nice to seem a `constexpr double name = ...;` for each. If it's decided to go down this path.

PR #356, opened 2020-03-05
Attribution Display

int-Frank:
[AB#671](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/671)

Attribution now displays at the bottom of the window with the following rules:
 - Displaying attribution overrides displaying of a watermark
 - Attribution is taken from every model in the scene and concatenated into a single string
 - The attribution text will automatically scroll if it does not fit in the render window.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1011,6 +1011,41 @@ epilogue:
   return 0;
 }
 
+void vcExtractAttributionText(vdkProjectNode *pNode, const char **ppCurrentText)
+{
+  if (pNode == nullptr)
+    return;
+
+  vcModel *pModel = nullptr;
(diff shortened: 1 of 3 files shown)


src/main.cpp
```diff
@@ -1011,6 +1013,42 @@ int main(int argc, char **args)
   return 0;
 }
 
+void vcExtractAttributionText(vdkProjectNode *pNode, const char **ppCurrentText)
```
OsiJr: This function should be a scene item function; its not only PointClouds that need this functionality long term

OsiJr:
There is some oddness here- I'm happy to merge it and continue to iterate on this functionality over time.

PR #366, opened 2020-03-10
Rendering labels (POI name etc.) directly to scene buffer

peterAdams1014:
Resolves [AB#1199](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1199)

### DIFF
--- a/src/rendering/vcLabelRenderer.cpp
+++ b/src/rendering/vcLabelRenderer.cpp
@@ -1,40 +1,142 @@
 #include "vcLabelRenderer.h"
 #include "udPlatformUtil.h"
+#include "gl/vcMesh.h"
+#include "gl/vcRenderShaders.h"
+#include "vcInternalModels.h"
 
 #include "imgui.h"
+#include <gl\vcGLState.h>
 
-bool vcLabelRenderer_Render(vcLabelInfo *pLabelRenderer, const udDouble4x4 &viewProjectionMatrix, const udUInt2 &screenSize)
+static struct vcLabelShader
 {
-  ImDrawList* drawList = ImGui::GetWindowDrawList();
+  vcShader *pShader;
+  vcShaderConstantBuffer *pEveryObjectConstantBuffer;
+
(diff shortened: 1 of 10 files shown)


src/rendering/vcLabelRenderer.cpp
```diff
@@ -1,40 +1,142 @@
 #include "vcLabelRenderer.h"
 #include "udPlatformUtil.h"
+#include "gl/vcMesh.h"
+#include "gl/vcRenderShaders.h"
+#include "vcInternalModels.h"
 
 #include "imgui.h"
+#include <gl\vcGLState.h>
```
samsinsane: Forward slash for compatibility with every non-Windows target we have.

src/rendering/vcLabelRenderer.cpp
```diff
@@ -1,40 +1,142 @@
...
+  } everyObject;
+} gShader;
+
+vcMesh *g_LabelMesh = nullptr;
+
+static int gRefCount = 0;
+udResult vcLabelRenderer_Init()
+{
+  udResult result;
+  gRefCount++;
```
samsinsane: This uses postfix, the decrement uses prefix. Additionally, is this intended to be thread safe?

src/vcRender.cpp
```diff
@@ -1203,8 +1218,9 @@ void vcRender_RenderScene(vcState *pProgramState, vcRenderContext *pRenderContex
 
   vcRenderSkybox(pProgramState, pRenderContext); // Drawing skybox after opaque geometry saves a bit on fill rate.
   vcRenderTerrain(pProgramState, pRenderContext);
-
+
```
samsinsane: Revert this.

src/rendering/vcLabelRenderer.cpp
peterAdams1014: No this is not mean to be thread safe. I have fixed the prefix issue.

PR #395, opened 2020-03-16
Fixes in preparation for using Shader Conductor

samsinsane:
- vcShader handles uniform and sampler names generated from Shader Conductor
- imguiVertexShader.hlsl uses the same layout as the vcLayout.h
- polygonP3N3UV2VertexShader.hlsl uses the same layout as the vcLayout.h
- Removed unused structures and predefined layouts in vcLayout.h

### DIFF
--- a/src/gl/opengl/vcShader.cpp
+++ b/src/gl/opengl/vcShader.cpp
@@ -156,8 +156,15 @@ bool vcShader_BindTexture(vcShader *pShader, vcTexture *pTexture, uint16_t sampl
   glBindTexture(vcTTToGL[pTexture->type], pTexture->id);
   VERIFY_GL();
 
-  if (pSampler != nullptr)
+  if (pSampler == nullptr)
+  {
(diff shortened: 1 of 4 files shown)


src/gl/opengl/vcShader.cpp
```diff
@@ -235,7 +245,7 @@ bool vcShader_GetSamplerIndex(vcShaderSampler **ppSampler, vcShader *pShader, co
   GLuint uID = glGetUniformLocation(pShader->programID, pSamplerName);
 
   if (uID == GL_INVALID_INDEX)
-    return false;
+    return true;
```
OsiJr: This seems wrong?

samsinsane: This is temporary - once we start using Shader Conductor for all of our shaders, this function won't be needed at all and can be changed to the DirectX implementation (it just returns `true`). We have something like `u_texture` in GLSL, but in HLSL it's `sampler0` and `texture0` and when this is converted using Shader Conductor it becomes `SPIRV_Cross_Combinedtexture0sampler0` -  we don't get this sampler index until we attempt to bind the texture, so we can't magic up this name in this function.

PR #399, opened 2020-03-17
Removed matrix equivalent check.

asunawu:
For [AB#1166](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1166). Always show export pointcloud menu.

### DIFF
--- a/src/scene/vcModel.cpp
+++ b/src/scene/vcModel.cpp
@@ -490,7 +490,16 @@ void vcModel::HandleContextMenu(vcState *pProgramState)
 #if VC_HASCONVERT
   if ((m_pPreferredProjection == nullptr && pProgramState->gis.SRID == 0) || (m_pPreferredProjection != nullptr && m_pPreferredProjection->srid == pProgramState->gis.SRID))
   {
-    if (ImGui::BeginMenu(vcString::Get("sceneExplorerExportPointCloud")))
+    bool matrixEqual = true;
+    for (int i = 0; i < m_defaultMatrix.ElementCount; i++)
+    {
+      if (abs(m_defaultMatrix.a[i] - m_sceneMatrix.a[i]) > UD_EPSILON)
+      {
+        matrixEqual = false;
+        break;
+      }
+    }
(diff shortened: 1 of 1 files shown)


src/scene/vcModel.cpp
```diff
@@ -488,7 +488,7 @@ void vcModel::HandleContextMenu(vcState *pProgramState)
   ImGui::Separator();
 
 #if VC_HASCONVERT
-  if (((m_pPreferredProjection == nullptr && pProgramState->gis.SRID == 0) || (m_pPreferredProjection != nullptr && m_pPreferredProjection->srid == pProgramState->gis.SRID)) && (m_defaultMatrix == m_sceneMatrix))
+  if ((m_pPreferredProjection == nullptr && pProgramState->gis.SRID == 0) || (m_pPreferredProjection != nullptr && m_pPreferredProjection->srid == pProgramState->gis.SRID))
```
OsiJr: That check is required, otherwise the model may be exported with a matrix that isn't the one used when rendering in the scene. Is this just wrong by a tiny amount and we might need to do a "equal within tolerance" check?

asunawu: Yes, t.x is slightly different. I can add a special check here.

asunawu: Restored the matrix equal check.

src/scene/vcModel.cpp
```diff
@@ -488,9 +488,18 @@ void vcModel::HandleContextMenu(vcState *pProgramState)
...
   {
-    if (ImGui::BeginMenu(vcString::Get("sceneExplorerExportPointCloud")))
+    bool matrixEqual = true;
+    for (int i = 0; i < m_defaultMatrix.ElementCount; i++)
+    {
+      if (abs(m_defaultMatrix.a[i] - m_sceneMatrix.a[i]) > UD_EPSILON)
```
samsinsane: This isn't compiling on Ubuntu 16.04 with Clang, you should use `udAbs` instead.

asunawu: I fixed it, and have moved the codes into udMath.

PR #403, opened 2020-03-18
Added ImGui as a submodule

OsiJr:
[AB#119](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/119)


src/vcConvert.cpp
```diff
@@ -1,6 +1,8 @@
 #include "vcConvert.h"
 
 #include "imgui.h"
+#include "imgui_internal.h"
```
samsinsane: No comment for what requires this?

src/vcModals.cpp
```diff
@@ -16,6 +16,8 @@
 #include "udStringUtil.h"
 
 #include "imgui.h"
+#include "imgui_internal.h"
```
samsinsane: No comment for what requires this?

src/vcConvert.cpp
OsiJr: `SeparatorEx` in case people look at this in future

src/vcModals.cpp
OsiJr: Lots of stuff in this file

PR #404, opened 2020-03-18
Saving convert window split proportion into account settings data

asunawu:
Fixed [AB#1029](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1029).

As the scene explorer did.

### DIFF
--- a/src/vcConvert.cpp
+++ b/src/vcConvert.cpp
@@ -326,7 +326,12 @@ void vcConvert_ShowUI(vcState *pProgramState)
   if (ImGui::IsWindowAppearing())
     ImGui::SetColumnWidth(0, size.x * pProgramState->settings.presentation.convertLeftPanelPercentage);
   else
+  {
     pProgramState->settings.presentation.convertLeftPanelPercentage = ImGui::GetColumnWidth(0) / size.x;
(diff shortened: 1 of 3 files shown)


src/vcConvert.cpp
```diff
@@ -322,8 +322,11 @@ void vcConvert_ShowUI(vcState *pProgramState)
   // Convert Jobs --------------------------------
   ImGui::Columns(2);
 
+  const ImVec2 &size = ImGui::GetWindowSize();
   if (ImGui::IsWindowAppearing())
-    ImGui::SetColumnWidth(0, 300);
+    ImGui::SetColumnWidth(0, size.x * pProgramState->settings.presentation.convertLeftPanelPercentage);
+  else
+    pProgramState->settings.presentation.convertLeftPanelPercentage = ImGui::GetColumnWidth(0) / size.x;
```
peterAdams1014: Is it possible to have a window size of `0`?

asunawu: Probably not for the value, but the left part could be dragged too small to recover, I'll set a minimum width for it.

asunawu: And setting maxinum width too.

PR #405, opened 2020-03-18
Added ability to display signed distances with model comparisons

int-Frank:
[AB#1197](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1197)

- Comparison models now export with signed distances
- The user now has a choice of rendering absolute or signed distances
- New UI elements to set up comparison viewing
- Gradient colour banding
- Possibility for users to set custom colours for colour banding (not implemented)

### DIFF
--- a/src/vcBPA.cpp
+++ b/src/vcBPA.cpp
@@ -130,7 +130,7 @@ void vcBPA_Init(vcBPAManifold **ppManifold, vdkContext *pContext)
   vcBPAManifold *pManifold = udAllocType(vcBPAManifold, 1, udAF_Zero);
   pManifold->grids.Init(1 << 10);
   pManifold->nodes.Init(1 << 13);
-  udWorkerPool_Create(&pManifold->pPool, 8, "vcBPAPool");
+  udWorkerPool_Create(&pManifold->pPool, (uint8_t)udGetHardwareThreadCount(), "vcBPAPool");
   pManifold->pContext = pContext;
   *ppManifold = pManifold;
 }
@@ -351,6 +351,13 @@ void vcBPA_GetNearbyPoints(vcBPAGrid *pGrid, udChunkedArray<uint64_t> *pPoints, 
   }
 }
 
+udDouble3 vcBPA_GetTriangleNormal(const udDouble3 & p0, const udDouble3 & p1, const udDouble3 & p2)
+{
+  udDouble3 a = p1 - p0;
+  udDouble3 b = p2 - p0;
+  return udNormalize(udCross(a, b));
+}
+
 vcBPATriangle vcBPA_FindSeedTriangle(vcBPAGrid *pGrid, double ballRadius)
 {
   // Pick any point not already in the triangle list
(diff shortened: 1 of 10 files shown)


src/vcVoxelShaders.h
```diff
@@ -22,8 +22,10 @@ struct vcUDRSData
     } classification;
     struct
     {
-      float minThreshold;
-      float maxThreshold;
+      float minBound;
+      float maxBound;
+      uint32_t errorColour;
+      uint32_t outOfBoundsColour;
```
OsiJr: This will likely conflict with @bwockner's work

int-Frank: I see there a lot of clashing with this [PR](https://github.com/Euclideon/vaultclient/pull/410). We have been working in the same area of the codebase.

bwockner: Mine is fairly small, I can fix it up if this one is approved first

src/vcBPA.cpp
```diff
@@ -398,16 +405,11 @@ vcBPATriangle vcBPA_FindSeedTriangle(vcBPAGrid *pGrid, double ballRadius)
         udULong3 triangle = udULong3::create((uint64_t)pointIndex, (uint64_t)nearbyPoints[i], (uint64_t)nearbyPoints[j]);
 
         // Check that the triangle normal is consistent with the vertex normals - i.e. pointing outward
-        udDouble3 a = p1 - p0;
-        udDouble3 b = p2 - p0;
-        udDouble3 triangleNormal = udNormalize(udCross(a, b));
+        udDouble3 triangleNormal = vcBPA_GetTriangleNormal(p0, p1, p2);
         if (triangleNormal.z < 0) // Normal is pointing down, that's bad!
```
samsinsane: In the interest of avoiding `triangleNormal` existing beyond this point with the wrong value, perhaps we should merge these two lines? `vcBPA_GetTriangleNormal(p0, p1, p2).z < 0`?

My main concern is that if we started using `triangleNormal` after this conditional, it would sometimes be wrong as the normal is no longer updated inside this block. I don't believe there's any reason for us to use the `triangleNormal` though.

src/vcBPA.cpp
```diff
@@ -1008,6 +1039,15 @@ vdkError vcBPA_ConvertReadPoints(vdkConvertCustomItem *pConvertInput, vdkPointBu
     float *pDisplacement = (float*)udAddBytes(pBuffer->pAttributes, pointAttrOffset + displacementOffset);
     *pDisplacement = (float)distance;
 
+    for (int elementIndex = 0; elementIndex < 3; ++elementIndex) //X,Y,Z
```
samsinsane: Could use `trianglePoint::ElementCount` or `udDouble3::ElementCount` instead of `3` - your call, with the comment it makes sense either way.

src/vcBPA.cpp
```diff
@@ -1120,6 +1167,24 @@ void vcBPA_CompareExport(vcState *pProgramState, const char *pOldModelPath, cons
     ++item.attributes.count;
   }
 
+  if (addDisplacementDirection)
+  {
+    item.attributes.pDescriptors[item.attributes.count].blendMode = vdkABM_SingleValue;
+    item.attributes.pDescriptors[item.attributes.count].typeInfo = vdkAttributeTypeInfo_float32;
```
samsinsane: Why not use `vdkAttributeTypeInfo_vec3f32`?

src/vcSettings.cpp
```diff
@@ -209,6 +209,12 @@ bool vcSettings_Load(vcSettings *pSettings, bool forceReset /*= false*/, vcSetti
     pSettings->camera.lensIndex = data.Get("camera.lensId").AsInt(vcLS_30mm);
     pSettings->camera.fieldOfView = data.Get("camera.fieldOfView").AsFloat(vcLens30mm);
 
+    pSettings->displacementShaderType = (vcDisplacementShaderType)data.Get("displacementShaderType").AsInt(0);
+    pSettings->visualization.displacement.bounds.x = data.Get("visualization.displacement.minBound").AsFloat(0.f);
+    pSettings->visualization.displacement.bounds.y = data.Get("visualization.displacement.maxBound").AsFloat(1.f);
+    pSettings->visualization.displacement.errorColour = data.Get("visualization.displacement.errorColour").AsInt(0);
+    pSettings->visualization.displacement.outOfBoundsColour = data.Get("visualization.displacement.outOfBoundsColour").AsInt(16777215);
```
samsinsane: I'd prefer to see colours like this specified in hex instead, `0xFFFFFF`. Makes it much clearer what kind of colour you're going for.

src/vcBPA.cpp
OsiJr: Can't- I forgot to take it out of VDK actually.

samsinsane: Ah ok - @int-Frank just ignore this comment! 😄

OsiJr:
I've handled this locally and merged with #410

PR #420, opened 2020-03-23
Fixed various issues with macOS and Metal builds

samsinsane:
- macOS: --force-vaultsdk now correctly builds libcurl on macOS
- Metal: Added missing post effects shader
- Metal: Fix crash when pCurrFramebuffer is nullptr and a framebuffer is being destroyed
- Metal: Shaders and textures now use 32bit float depth 8bit stencil, instead of 24bit int depth 8bit stencil
- Metal: ImGui shader doesn't bind on first call to bind like all other shaders
- Metal: Shader function names load off disk correctly

### DIFF
--- a/src/gl/metal/Shaders.metal
+++ b/src/gl/metal/Shaders.metal
@@ -825,3 +825,580 @@ highlightFragmentShader(HFSInput in [[stage_in]], constant HVSUniforms& uniforms
     result = max(uniforms.u_stepSizeThickness.w, result) * uniforms.u_color.w; // overlay color
     return float4(uniforms.u_color.xyz, result);
 }
+
+// g_postEffectsVertexShader
+
+struct PEVParams
+{
+  float4 u_screenParams;
+  float4 u_saturation;
+};
+
+struct PEVOutput
+{
+  float2 out_var_TEXCOORD0 [[user(locn0)]];
+  float2 out_var_TEXCOORD1 [[user(locn1)]];
+  float2 out_var_TEXCOORD2 [[user(locn2)]];
+  float2 out_var_TEXCOORD3 [[user(locn3)]];
+  float2 out_var_TEXCOORD4 [[user(locn4)]];
+  float2 out_var_TEXCOORD5 [[user(locn5)]];
+  float out_var_TEXCOORD6 [[user(locn6)]];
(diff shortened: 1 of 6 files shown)


src/gl/metal/Shaders.metal
```diff
@@ -825,3 +825,580 @@ highlightFragmentShader(HFSInput in [[stage_in]], constant HVSUniforms& uniforms
...
+  else
+  {
+    float4 _534;
+    switch (0u)
+    {
+      default:
+      {
+        float2 _123 = _67;
+        _123.x = in.in_var_TEXCOORD0.x;
+        float2 _125 = _123;
```
OsiJr: You didn't need this variable, `_123` isn't used other than this

src/gl/metal/vcGLState.mm
```diff
@@ -113,9 +113,8 @@ bool vcGLState_Init(SDL_Window *pWindow, vcFramebuffer **ppDefaultFramebuffer)
   _renderer = [vcRenderer alloc];
   [_renderer initWithView:sdlview];
 
-  vcTexture *defaultTexture, *defaultDepth;
+  vcTexture *defaultTexture, *defaultDepth = nullptr;
```
OsiJr: This fix won't set `defaultTexture` to NULL

samsinsane: It didn't need to, `defaultDepth` is no longer set.

src/gl/metal/Shaders.metal
samsinsane: This will have something to do with the way the HLSL shader is written. Lots of inline maths gets expanded into locals like this that do a whole bunch of nothing.

src/gl/metal/vcGLState.mm
samsinsane: In hindsight, I'm not sure if I needed to remove the depth texture in the end. I vaguely recall there being issues with it - lots of work is required for Metal, so I might loop back to this anyway.

samsinsane: Turns out this change wasn't necessary.

samsinsane: I fixed this up.

PR #429, opened 2020-03-24
Moved attribution text to vcProject

OsiJr:


src/vcProject.h
```diff
@@ -33,12 +33,14 @@ void vcProject_RemoveSelected(vcState *pProgramState);
 
 void vcProject_SelectItem(vcState *pProgramState, vdkProjectNode *pParent, vdkProjectNode *pNode);
 void vcProject_UnselectItem(vcState *pProgramState, vdkProjectNode *pParent, vdkProjectNode *pNode);
-void vcProject_ClearSelection(vcState *pProgramState);
+void vcProject_ClearSelection(vcState *pProgramState, bool clearToolState = true);
```
samsinsane: This doesn't appear to be part of this PR?

samsinsane: To confirm, this change breaks the builds.

PR #440, opened 2020-03-25
Texture Format and Vertex Textures

peterAdams1014:
Added R16 (signed int) texture format.
(WIP) Added ability to bind textures to vertex shading stage.


samsinsane:
You'll need to add this texture type into Metal - `MTLPixelFormatR16Snorm`, I think it just needs to go into `vcTexture_GetFormatAndPixelSize` but probably best to just find all places that use `MTLPixelFormat`.

PR #441, opened 2020-03-25
awu/export point cloud file name

asunawu:
Fixed [AB#1112](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1112).

Set a default file name to be exported if there isn't and added ".las" if no extension when user clicking export button.

### DIFF
--- a/src/imgui_ex/vcFileDialog.h
+++ b/src/imgui_ex/vcFileDialog.h
@@ -18,7 +18,6 @@ static const char *SupportedFileTypes_ProjectsImport[] = { ".json", ".udp" };
 
 static const char *SupportedTileTypes_ConvertExport[] = { ".uds" };
 static const char *SupportedTileTypes_QueryExport[] = { ".uds", ".las" };
-static const char *SupportedTileTypes_ExportDefault = ".las";
 static const char *SupportedFileTypes_ConvertImport[] = { ".uds", ".ssf", ".udg", ".las", ".laz", ".e57", ".txt", ".csv", ".asc", ".pts", ".ptx", ".xyz", ".obj"
 #ifdef FBXSDK_ON
   , ".fbx"
@@ -62,7 +61,6 @@ template <size_t N, size_t M> inline void vcFileDialog_Open(vcState *pProgramSta
   udUnused(SupportedTileTypes_ConvertExport);
   udUnused(SupportedTileTypes_QueryExport);
   udUnused(SupportedFileTypes_ConvertImport);
-  udUnused(SupportedTileTypes_ExportDefault);
 }
 
 void vcFileDialog_ShowModal(vcState *pProgramState);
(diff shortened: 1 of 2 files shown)


src/imgui_ex/vcFileDialog.h
```diff
@@ -18,6 +18,7 @@ static const char *SupportedFileTypes_ProjectsImport[] = { ".json", ".udp" };
 
 static const char *SupportedTileTypes_ConvertExport[] = { ".uds" };
 static const char *SupportedTileTypes_QueryExport[] = { ".uds", ".las" };
+static const char *SupportedTileTypes_ExportDefault = ".las";
```
OsiJr: I would recommend just using `SupportedTileTypes_QueryExport[0]` instead- the default should probably be `.uds` anyway

src/scene/vcModel.cpp
```diff
@@ -526,6 +540,13 @@ void vcModel::HandleContextMenu(vcState *pProgramState)
       {
         if (udFileExists(pProgramState->modelPath) != udR_Success || vcModals_OverwriteExistingFile(pProgramState->modelPath))
         {
+          if (!udStrchr(pProgramState->modelPath, "."))
+          {
+            // Adding a default externsion
+            size_t modelPathLength = udStrlen(pProgramState->modelPath);
+            udStrncpy(pProgramState->modelPath + modelPathLength, sizeof(pProgramState->modelPath), SupportedTileTypes_ExportDefault, udStrlen(SupportedTileTypes_ExportDefault));
+          }
```
OsiJr: This can be greatly simplified:

```suggestion
          if (!udStrchr(pProgramState->modelPath, ".")) // Adding a default extension
            udStrcat(pProgramState->modelPath, SupportedTileTypes_ExportDefault);
```

src/scene/vcModel.cpp
```diff
@@ -509,8 +509,22 @@ void vcModel::HandleContextMenu(vcState *pProgramState)
...
 
+        // Set a default file name to be exported if there isn't.
+        if (udStrlen(pProgramState->modelPath) == 0)
+        {
+          size_t fileNameLength = 0;
+          if (udStrchr(pProgramState->sceneExplorer.clickedItem.pItem->pName, ".", &fileNameLength))
+          {
+            memset(pProgramState->modelPath, 0, sizeof(pProgramState->modelPath));
+            udStrncpy(pProgramState->modelPath, pProgramState->sceneExplorer.clickedItem.pItem->pName, fileNameLength);
+            udStrncpy(pProgramState->modelPath + fileNameLength, sizeof(pProgramState->modelPath), SupportedTileTypes_ExportDefault, udStrlen(SupportedTileTypes_ExportDefault));
```
OsiJr: Would recommend `udStrcat` here as its much cleaner:

```suggestion
            udStrcat(pProgramState->modelPath, SupportedTileTypes_ExportDefault);
```

src/scene/vcModel.cpp
asunawu: Fixed.

src/scene/vcModel.cpp
asunawu: Fixed.

src/imgui_ex/vcFileDialog.h
asunawu: Fixed.

PR #442, opened 2020-03-25
awu/separate speedup hotkey info

asunawu:
Fixed [AB#836](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/836).

Removed speed up and slow down hotkey info from basic movements and showed them on top.
Fixed a small translate problem too.


OsiJr:
It's probably cleaner to just add the extra 4 bindings `Move Forward Fast` etc, that also removes lots of crazy complexity for supporting keys mods and things

asunawu:
ok

asunawu:
Keys mod was used to support user-defined key combinations (such as "shift + E"), I think we can't remove it unless we don't support them any more?

PR #444, opened 2020-03-25
Moved various things out of vcRenderer

samsinsane:
- Metal: Added missing camera plane uniform upload
- Metal: Moved pipelines out of vcRenderer into vcShader
- Metal: Moved textures out of vcRenderer into vcTexture
- Metal: Moved samplers out of vcRenderer into vcTexture
- Metal: Moved blit buffers out of vcRenderer into vcTexture
- Metal: Moved render passes out of vcRenderer into vcFramebuffer
- Metal: Moved command buffer out of vcRenderer into vcFramebuffer
- Metal: Moved command encoder out of vcRenderer into vcFramebuffer
- Metal: Moved index buffers out of vcRenderer into vcMesh
- Metal: Moved vertex buffers out of vcRenderer into vcMesh
- Metal: Moved depth states out of vcRenderer into vcGLState
- Metal: Removed command queue in vcRenderer in favour of unused queue in vcGLState
- Metal: Removed unused compute variables in vcRenderer

### DIFF
--- a/src/gl/metal/vcShader.mm
+++ b/src/gl/metal/vcShader.mm
@@ -1,4 +1,5 @@
 #import "vcMetal.h"
+#import "vcConstants.h"
 #import "gl/vcShader.h"
 #import <Metal/Metal.h>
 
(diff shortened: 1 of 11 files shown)


src/gl/metal/vcShader.mm
```diff
@@ -155,7 +156,7 @@ bool vcShader_CreateFromTextInternal(vcShader **ppShader, const char *pVertexSha
         break;
     }
 
-    [_renderer.pipelines addObject:[_device newRenderPipelineStateWithDescriptor:pDesc error:&err]];
+    pShader->pipelines[i] = [_device newRenderPipelineStateWithDescriptor:pDesc error:&err];
```
OsiJr: To confirm; this is still correct when line 122 looks like it does the same thing (at least when `i==0`)?

samsinsane: It's not quite the same thing - line 122 also generates the reflection information for the constant buffers. Also, I'm not entirely sure we need a unique pipeline for each blend mode, but that's how it worked before. Finally, since this also caught me out, this loop starts at `i==1` not `i==0` - insanely easy to gloss over.

OsiJr: Sorry; yes you're right. I totally missed that in the loop (I guess that's a negative for not magic numbers)

PR #452, opened 2020-03-27
Fixed profile loading crash of user defined classification labels and…

asunawu:
Resolve [AB#1009](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1009).

Also fixed:
1. An extra crash when loading the user defined name.
2. Reset cache labels when reloading profile.

### DIFF
--- a/src/vcSettings.cpp
+++ b/src/vcSettings.cpp
@@ -96,8 +96,10 @@ void vcSettings_Cleanup_CustomClassificationColorLabels(vcSettings *pSettings)
 {
   size_t lenth = udLengthOf(pSettings->visualization.customClassificationColorLabels);
   for (size_t i = 0; i < lenth; ++i)
+  {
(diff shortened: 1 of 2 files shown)


src/vcSettings.cpp
```diff
@@ -92,6 +92,14 @@ bool SDL_filewrite(const char* filename, const char *pStr, size_t bytes)
   return success;
 }
 
+void vcSettings_Cleanup_CustomClassificationColorLabels(vcSettings *pSettings)
+{
+  size_t lenth = udLengthOf(pSettings->visualization.customClassificationColorLabels);
+  for (size_t i = 0; i < lenth; ++i)
```
OsiJr: This is a multiline `for` so requires braces (there is another later as well)

asunawu: Fixed.

PR #477, opened 2020-04-02
Shader sampler uniform locations now grabbed on initialization. (OpenGL)

peterAdams1014:
Renamed some texture uniforms in shaders to be more meaningful.


src/gl/opengl/vcShader.cpp
```diff
@@ -243,7 +236,7 @@ bool vcShader_GetSamplerIndex(vcShaderSampler **ppSampler, vcShader *pShader, co
   if (pShader->numSamplerIndexes >= (int)udLengthOf(pShader->samplerIndexes))
     return false;
 
-  GLuint uID = glGetUniformLocation(pShader->programID, pSamplerName);
+  GLuint uID = glGetUniformLocation(pShader->programID, udTempStr("SPIRV_Cross_Combined%sTexture%sSampler", pSamplerName, pSamplerName));
 
   if (uID == GL_INVALID_INDEX)
     return true;
```
samsinsane, line 242: You need to return `false` here, this was a hack to do the lookup in the `vcTexture_BindTexture` function, which you've removed.

PR #484, opened 2020-04-03
awu/DoubleClickLeftMouse

asunawu:
[AB#1415](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1415).


src/vcCamera.cpp
```diff
@@ -498,7 +500,7 @@ void vcCamera_HandleSceneInput(vcState *pProgramState, udDouble3 oscMove, udFloa
   }
 
   // Double Clicking left mouse
-  if (isBtnDoubleClicked[0] && (pProgramState->pickingSuccess))
+  if (isBtnDoubleClicked[0] && (pProgramState->udModelPickedIndex != -1))
```
OsiJr, line 503: `pickingSuccess` should be false on the skybox so I don't think this fixes it- it will just break on non-UDS models

asunawu: Fixed pickingSuccess flag.

PR #486, opened 2020-04-06
Added Pin support for collapsing far off items

OsiJr:


src/scene/vcLiveFeed.cpp
```diff
@@ -270,6 +271,10 @@ void vcLiveFeed_UpdateFeed(void *pUserData)
 
               lodRef.pLabelInfo->pText = lodRef.pLabelText;
             }
+
+            const udJSON &pinObj = pLOD->Get("pin");
+            if (pinObj.IsObject())
+              lodRef.pPinIcon = udStrdup(pinObj.Get("image").AsString());
```
OsiJr, line 277: Free this first

src/vcRender.h
```diff
@@ -52,6 +52,8 @@ struct vcRenderPolyInstance
 
   // Helper to determine if flag is set
   bool HasFlag(const enum RenderFlags flag) { return (renderFlags & flag) == flag; }
+
+  const char *pPinIconURL; // null means not pinnable
```
OsiJr, line 56: This is no longer required

src/rendering/vcPinRenderer.cpp
```diff
@@ -0,0 +1,230 @@
...
+  *ppPinRenderer = pPinRenderer;
+  return udR_Success;
+}
+
+udResult vcPinRenderer_Destroy(vcPinRenderer **ppPinRenderer)
+{
+  vcPinRenderer *pPinRenderer = (*ppPinRenderer);
+  *ppPinRenderer = nullptr;
+
+
```
peterAdams1014, line 110: Excessive whitespace.

src/rendering/vcPinRenderer.cpp
```diff
@@ -0,0 +1,230 @@
...
+  return udR_Success;
+}
+
+bool vcPinRenderer_AddPin(vcPinRenderer *pPinRenderer, vcState *pProgramState, const char *pPinIconURL, udDouble3 position, int count)
+{
+  if (pPinIconURL == nullptr)
+    return false;
+
+  count = udMax(count, 1);
```
peterAdams1014, line 133: This seems odd. Intended?

peterAdams1014:
Few minor issues, but not blocking.

src/rendering/vcPinRenderer.cpp
OsiJr: Yeah, if count is `0` or negative I want it to at least count for `1`

src/rendering/vcPinRenderer.cpp
```diff
@@ -0,0 +1,233 @@
...
+    vcLoadState_InQueue,
+    vcLoadState_Downloaded,
+    vcLoadState_Loaded,
+    vcLoadState_Failed
+  } volatile loadState; // todo: proper thread safety
+  const char *pIconURL;
+
+  vcTexture *pTexture;
+  void *pTextureData;
+  int width, height;
```
samsinsane, line 25: These should be on different lines, perhaps we should create a `vcSize`?

src/vcRender.cpp
```diff
@@ -1053,13 +1060,14 @@ void vcRender_TransparentPass(vcState *pProgramState, vcRenderContext *pRenderCo
 
     for (size_t i = 0; i < renderData.images.length; ++i)
     {
-      double zScale = 1.0;
-      zScale -= udMag3(pProgramState->camera.position - renderData.images[i]->position) / pProgramState->settings.presentation.imageRescaleDistance;
+      vcImageRenderInfo *pImage = renderData.images[i];
+      double distToCamera = udMag3(pProgramState->camera.position - renderData.images[i]->position);
```
samsinsane, line 1064: This should be `pImage` otherwise it doesn't really make sense to have `pImage` grabbed here just for the last line, that might not be hit.

PR #488, opened 2020-04-06
Normalising camera axis before creating quaternion

int-Frank:
Otherwise it hits the assert n `udDoubleQuat::create()` when we 'move to' another location in the scene.


OsiJr:
This was already fixed by @asunawu in master.

PR #493, opened 2020-04-07
Minor improvements to camera display

OsiJr:


src/main.cpp
```diff
@@ -1170,32 +1176,32 @@ void vcRenderSceneUI(vcState *pProgramState, const ImVec2 &windowPos, const ImVe
...
+        if (pProgramState->gis.isProjected)
+        {
+          cameraLatLong = udGeoZone_CartesianToLatLong(pProgramState->gis.zone, pProgramState->camera.position);
+          if (ImGui::InputScalarN(vcString::Get("sceneCameraPositionGIS"), ImGuiDataType_Double, &cameraLatLong.x, 3))
+            pProgramState->camera.position = udGeoZone_LatLongToCartesian(pProgramState->gis.zone, cameraLatLong);
+        }
+
         udDouble2 headingPitch = UD_RAD2DEG(pProgramState->camera.headingPitch);
-        if (ImGui::InputScalarN(vcString::Get("sceneCameraRotation"), ImGuiDataType_Double, &headingPitch.x, 2))
+        if (ImGui::InputScalarN(vcString::Get("sceneCameraRotation"), ImGuiDataType_Double, &headingPitch.x, 2, nullptr, nullptr, "%.2f°"))
```
samsinsane, line 1187: If that degree character is not part of the ASCII set, it might be better to store it as the escaped byte sequence instead. From memory, one of the programs we use gets funny sometimes when dealing with files that contain non-ASCII and will insert a BOM at the start of the document and break builds.

OsiJr: Good point, removed it.

PR #499, opened 2020-04-08
Fixed camera orbit and pan when clicking outside the map. 

asunawu:
Fixed camera orbit and pan outside the map. Also fixed look around after orbit(may be [AB#1396](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1396)). And context menu outside the map.

### DIFF
--- a/src/vcCamera.cpp
+++ b/src/vcCamera.cpp
@@ -56,7 +56,7 @@ void vcCamera_UpdateSmoothing(vcCamera *pCamera, vcCameraInput *pCamInput, doubl
       pCamInput->smoothRotation -= step;
 
       pCamera->headingPitch.x = udMod(pCamera->headingPitch.x, UD_2PI);
-      pCamera->headingPitch.y = udClamp(pCamera->headingPitch.y, -UD_HALF_PI, UD_HALF_PI);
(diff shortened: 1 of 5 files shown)


src/vcCamera.cpp
```diff
@@ -56,7 +56,7 @@ void vcCamera_UpdateSmoothing(vcCamera *pCamera, vcCameraInput *pCamInput, doubl
       pCamInput->smoothRotation -= step;
 
       pCamera->headingPitch.x = udMod(pCamera->headingPitch.x, UD_2PI);
-      pCamera->headingPitch.y = udClamp(pCamera->headingPitch.y, -UD_HALF_PI, UD_HALF_PI);
+      pCamera->headingPitch.y = udMod(pCamera->headingPitch.y, UD_2PI);
```
OsiJr, line 59: Pitch should be UD_HALF_PI (straight up) to -UD_HALF_PI (straight down), otherwise the camera will flip over and be 'upside-down'

OsiJr: I was wrong, @asunawu did a demo and this is correct.

PR #508, opened 2020-04-14
Fixed issue with some polygons not rendering in Metal

samsinsane:


samsinsane:
I picked 30, it wasn't a shader conductor thing because shader conductor didn't allow us to offset the indexes for the buffers. Previously, I changed index draw calls but accidentally forgot non-indexed.

PR #511, opened 2020-04-14
Fixed POI bug and organised code

int-Frank:
[AB#1374](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1374)

Changes:

I have added a Tools section in settings, where the user can set line and node defaults
Also, In an effort to simplify the vcPOI class, I have pulled out the three different states: MeasureLine, MeasureArea and General, and added them to an internal state machine `m_pState`. I found the POI functions a little differently depending on what it was being used for. This seemed to be a source of bugs, as each POI method was trying to handle each mode, leading to excessive state checking and redundant code.

### DIFF
--- a/src/scene/vcPOI.cpp
+++ b/src/scene/vcPOI.cpp
@@ -31,10 +31,299 @@ static const char *vcFRIMStrings[] =
 };
 UDCOMPILEASSERT(udLengthOf(vcFRIMStrings) == vcRRIM_Count, "New enum values");
 
+//----------------------------------------------------------------------------------------------------
+// vcPOIState_General
+//----------------------------------------------------------------------------------------------------
+
+class vcPOIState_General
+{
+protected:
+  vcPOI *m_pParent;
+public:
+  vcPOIState_General(vcPOI *pParent)
+    : m_pParent(pParent)
+  {
+
+  }
+
+  virtual ~vcPOIState_General()
+  {
+
+  }
+
+  virtual void HandlePopupUI(vcState * /*pProgramState*/)
+  {
+
+  }
+
+  virtual void AddPoint(vcState * pProgramState, const udDouble3 &position, bool /*isPreview*/)
+  {
+    m_pParent->InsertPoint(position);
+
+    m_pParent->UpdatePoints();
+
+    vcProject_UpdateNodeGeometryFromCartesian( m_pParent->m_pProject,  m_pParent->m_pNode, pProgramState->gis.zone,  m_pParent->m_line.closed ? vdkPGT_Polygon : vdkPGT_LineString,  m_pParent->m_line.pPoints,  m_pParent->m_line.numPoints);
+    m_pParent->m_line.selectedPoint =  m_pParent->m_line.numPoints - 1;
+  }
+
+  virtual void AddToScene(vcState *pProgramState, vcRenderData *pRenderData)
+  {
+    if (!m_pParent->IsVisible(pProgramState))
+      return;
+
+    if (m_pParent->m_selected)
+    {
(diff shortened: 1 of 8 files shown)


src/scene/vcPOI.cpp
```diff
@@ -31,11 +31,300 @@ static const char *vcFRIMStrings[] =
...
+// vcPOI
+//----------------------------------------------------------------------------------------------------
+
 vcPOI::vcPOI(vdkProject *pProject, vdkProjectNode *pNode, vcState *pProgramState) :
   vcSceneItem(pProject, pNode, pProgramState)
+  , m_pState(nullptr)
 {
-  m_nameColour = 0xFFFFFFFF;
-  m_backColour = 0x7F000000;
+  m_nameColour = vcIGSW_ImGuiToBGRA(pProgramState->settings.tools.label.textColour);
```
OsiJr, line 326: The `vcPOI` constructor shouldn't be assigning anything non-const to the values- the `OnNodeUpdate` function can default to user settings if required. None of these values will be saved otherwise.

src/scene/vcPOI.cpp
```diff
@@ -69,11 +358,34 @@ vcPOI::vcPOI(vdkProject *pProject, vdkProjectNode *pNode, vcState *pProgramState
 
   memset(&m_flyThrough, 0, sizeof(m_flyThrough));
 
+  switch (pProgramState->activeTool)
```
OsiJr, line 361: Similiarly, this cannot be here- this should be a param or something otherwise if a update happens over the network while the user is in a tool this will break.

src/scene/vcPOI.cpp
```diff
@@ -771,3 +963,122 @@ bool vcPOI::LoadAttachedModel(const char *pNewPath)
 
   return false;
 }
+
+bool vcPOI::IsVisible(vcState *pProgramState)
+{
+  // if POI is invisible or if it exceeds maximum visible POI distance (unless selected)
+  bool visible = m_visible;
+  visible = visible && m_selected;
+  visible = visible && (udMag3(m_pLabelInfo->worldPosition - pProgramState->camera.position) < pProgramState->settings.presentation.POIFadeDistance);
```
OsiJr, line 972: Should these be `||`?

src/scene/vcPOI.cpp
int-Frank: Roger

src/scene/vcPOI.cpp
int-Frank: Roger

src/scene/vcPOI.cpp
int-Frank: I have moved this to OnNodeUpdate(), although I don't quite see the difference if OnNodeUpdate is called in the constructor?

src/scene/vcPOI.cpp
int-Frank: You're right, this is incorrect. A POI is visible if it is _not_ invisible _and_ it must be in the view frustum, _but_ if it is selected it should be visible.

src/scene/vcPOI.cpp
OsiJr: OnNodeUpdate is called whenever the node changes (this user or another user)

int-Frank: Right, although we already `UpdateState()` in OnNodeUpdate. This switch statement simply initialises state in the constructor. I've moved this to OnNodeUpdate, but it will only ever be hit once; during construction.

PR #518, opened 2020-04-16
Optimized handling of pins for extremely large pin counts.

peterAdams1014:


src/rendering/vcPinRenderer.cpp
```diff
@@ -159,10 +159,10 @@ bool vcPinRenderer_AddPin(vcPinRenderer *pPinRenderer, vcState *pProgramState, c
   for (; b < pPinRenderer->pinBins.length; ++b)
   {
     vcPinBin *pBin = &pPinRenderer->pinBins[b];
-    double distToBin = udMag3(position - pBin->position);
-    double cameraDistToBin = udMag3(pBin->position - pProgramState->camera.position);
+    double distToBinSqr = udMagSq3(position - pBin->position);
+    double cameraDistToBinSqr = udMagSq3(pBin->position - pProgramState->camera.position);
 
-    if (pBin->pIcon == pIcon && distToBin <= (cameraDistToBin * COLLAPSE_DISTANCE_RATIO))
+    if (pBin->pIcon == pIcon && distToBinSqr <= (cameraDistToBinSqr * COLLAPSE_DISTANCE_RATIO))
```
samsinsane, line 165: This doesn't work the same way as before unless you square the ratio too. Was it intentional to require the camera to be much closer to the bin than before?

PR #535, opened 2020-04-17
Fixed camera jumps after releasing mouse middle button.

asunawu:
[AB#1522](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1522)


src/vcCamera.cpp
```diff
@@ -211,6 +211,9 @@ void vcCamera_Apply(vcState *pProgramState, vcCamera *pCamera, vcCameraSettings
       // Save it back to the camera
       pCamera->position = pProgramState->worldAnchorPoint + rotation.apply(direction); // define new position
       pCamera->headingPitch = vcGIS_QuaternionToHeadingPitch(pProgramState->gis, pProgramState->camera.position, orientation);
+
+      if (pCamera->headingPitch.y > UD_PI)
```
samsinsane, line 215: Should this be `while` instead of `if`? Is it possible for the value to be greater than `2 * UD_PI`?

OsiJr: Shouldn't be possible in the current paths

PR #559, opened 2020-04-21
Depth now stored in alpha of colour attachments.

peterAdams1014:
Changed colour to rgba16f, and storing depth in alpha.
Fixed a bug where polygon models were not splatting their IDs correctly.
Fixed an issue with DirectX pixel readback assumed a pixel size of 4 bytes.
D3D: On passive picking (async readback), we're now only doing a 1x1 readback, instead of a full texture width*height readback (~1ms saved on desktop).

### DIFF
--- a/src/rendering/vcPolygonModel.cpp
+++ b/src/rendering/vcPolygonModel.cpp
@@ -508,8 +508,7 @@ udResult vcPolygonModel_Render(vcPolygonModel *pModel, const udDouble4x4 &modelM
       pPolygonShader = &gShaders[vcPMST_P3N3UV2_FlatColour];
     else if (passType == vcPMP_Shadows)
       pPolygonShader = &gShaders[vcPMST_P3N3UV2_DepthOnly];
-
-    if (pDiffuseOverride)
+    else if (pDiffuseOverride)
     {
       pDiffuseTexture = pDiffuseOverride;
       pPolygonShader = &gShaders[vcPMST_P3N3UV2_Opaque_Image]; // Assuming that models with textures could skip lighting.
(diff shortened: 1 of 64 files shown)


src/rendering/vcPolygonModel.cpp
```diff
@@ -508,8 +508,7 @@ udResult vcPolygonModel_Render(vcPolygonModel *pModel, const udDouble4x4 &modelM
       pPolygonShader = &gShaders[vcPMST_P3N3UV2_FlatColour];
     else if (passType == vcPMP_Shadows)
       pPolygonShader = &gShaders[vcPMST_P3N3UV2_DepthOnly];
-
-    if (pDiffuseOverride)
+    else if (pDiffuseOverride)
```
OsiJr, line 511: All of these conditions should be blocks now

src/vcRender.cpp
```diff
@@ -582,6 +582,44 @@ udResult vcRender_ResizeScene(vcState *pProgramState, vcRenderContext *pRenderCo
...
+//float d = Float16ToFloat32(0b0111101111111111); //65504
+//float e = Float16ToFloat32(0b0011101111111111); //0.99951
+//float f = Float16ToFloat32(0b0011110000000000); //1
+//float g = Float16ToFloat32(0b0011110000000001); //1.001
+//float h = Float16ToFloat32(0b0011010101010101); //0.333251953125
+//float i = Float16ToFloat32(0b1100000000000000); //-2
+//float j = Float16ToFloat32(0b0000000000000000); //0
+//float k = Float16ToFloat32(0b1000000000000000); //-0
+//float l = Float16ToFloat32(0b0111110000000000); //inf
+//float m = Float16ToFloat32(0b1111110000000000); //-inf
```
OsiJr, line 599: These could be tested in `vcTests`

OsiJr: I'd probably move the funcition to `vcMath` with the `ud` prefix as well

PR #565, opened 2020-04-23
DEM and Colour tiles now load into a proxy

int-Frank:
[AB#1577](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1577)

I'm loading the tiles into a proxy before inserting them back into the quadtree. This way we can release the cache mutex, allowing other work to be done on the nodes (clearing the tree for example) without having to wait for tiles to finish loading.

### DIFF
--- a/src/rendering/vcTileRenderer.cpp
+++ b/src/rendering/vcTileRenderer.cpp
@@ -119,6 +119,12 @@ struct vcTileRenderer
   } presentShader;
 };
 
+struct ProxyTexture
+{
+  int32_t width, height;
+  void *pData;
+};
+
 // This functionality here for now until the cache module is implemented
 bool vcTileRenderer_TryWriteTile(const char *filename, void *pFileData, size_t fileLen)
 {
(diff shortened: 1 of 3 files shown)


src/rendering/vcTileRenderer.cpp
```diff
@@ -304,53 +308,88 @@ uint32_t vcTileRenderer_LoadThread(void *pThreadData)
...
+      if (canDownloadDEM)
+        demResult = vcTileRenderer_HandleTileDownload(&textureDEM, serverURLDEM, localURLDEM);
+
+      if (canDownloadColour)
+        colourResult = vcTileRenderer_HandleTileDownload(&textureColour, serverURLColour, localURLColour);
+
+      udLockMutex(pCache->pMutex);
+
+      //Check if the node is still valid.
+      if (canDownloadDEM && pBestNode->demInfo.loadStatus.Get() == vcNodeRenderInfo::vcTLS_Downloading)
```
int-Frank, line 358: The thread should assign ownership to `pBestNode`. Otherwise while this thread is loading data, the node could be cleared and taken by another thread and subsequently flagged as `vcTLS_Downloading`. Now we have two threads trying to populate in this node.

peterAdams1014:
As per discussion with @int-Frank , we have decided on a better approach to solve this issue.

PR #568, opened 2020-04-24
Fixed orbit flickering when looking up/down.

asunawu:
Limit the camera orbit moving as the pitch angle near +- half PI.

### DIFF
--- a/src/vcCamera.cpp
+++ b/src/vcCamera.cpp
@@ -194,17 +194,24 @@ void vcCamera_Apply(vcState *pProgramState, vcCamera *pCamera, vcCameraSettings 
     if (distanceToPointSqr != 0.0 && (pCamInput->mouseInput.x != 0 || pCamInput->mouseInput.y != 0))
     {
       // Orbit Left/Right
-      udDoubleQuat rotation = udDoubleQuat::create(worldAnchorNormal, pCamInput->mouseInput.x);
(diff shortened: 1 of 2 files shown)


src/vcCamera.cpp
```diff
@@ -194,26 +194,31 @@ void vcCamera_Apply(vcState *pProgramState, vcCamera *pCamera, vcCameraSettings
...
 
-      pCamera->position = pProgramState->worldAnchorPoint + rotation.apply(direction); // define new position
-      pCamera->headingPitch = vcGIS_QuaternionToHeadingPitch(pProgramState->gis, pProgramState->camera.position, orientation);
+        pCamera->position = pProgramState->worldAnchorPoint + rotation.apply(direction); // define new position
+        pCamera->headingPitch = vcGIS_QuaternionToHeadingPitch(pProgramState->gis, pProgramState->camera.position, orientation);
+      }
+
+      //
+      if (pCamera->headingPitch.y < UD_DEG2RADf(-89.0) && pCamInput->mouseInput.y <= 0) break;
+      if (pCamera->headingPitch.y > UD_DEG2RADf(89.0) && pCamInput->mouseInput.y >= 0) break;
```
OsiJr, line 210: the `break` should be on the next line (even if it is less visually appealing)

asunawu: fixed.

PR #570, opened 2020-04-26
Fences now use world up vector

int-Frank:
[AB#1514](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1514)

### DIFF
--- a/src/rendering/vcFenceRenderer.cpp
+++ b/src/rendering/vcFenceRenderer.cpp
@@ -35,13 +35,11 @@ struct vcFenceRenderer
     {
       udFloat4 primaryColour;
       udFloat4 secondaryColour;
-      float orientation; // 0 == fence, 1 == flat
+      udFloat4 extendVector;
       float width;
       float textureRepeatScale;
       float textureScrollSpeed;
       float time;
-
-      char _padding[12]; // 16 byte alignment
     } everyFrameParams;
 
     struct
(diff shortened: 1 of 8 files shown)


src/rendering/vcFenceRenderer.cpp
```diff
@@ -341,9 +258,11 @@ udResult vcFenceRenderer_CreateSegmentVertexData(vcFenceRenderer *pFenceRenderer
...
   current = udFloat3::create(pSegment->pCachedPoints[0]);
   next = udFloat3::create(pSegment->pCachedPoints[1]);
-  expandVector = vcFenceRenderer_CreateStartJointExpandVector(current, next, pFenceRenderer->config.ribbonWidth, worldUp);
-  pVerts[vertIndex + 0].ribbonInfo = udFloat4::create(expandVector, 0.0f);
-  pVerts[vertIndex + 1].ribbonInfo = udFloat4::create(-expandVector, 1.0f);
+
+  //We don't use the full vec4, but leave it as is; we may want to use it laster to support
+  //individual extend vectors for each point in the line.
+  pVerts[vertIndex + 0].ribbonInfo = udFloat4::create(0.0, 0.0, 0.0, 0.0f);
+  pVerts[vertIndex + 1].ribbonInfo = udFloat4::create(0.0, 0.0, 0.0, 1.0f);
```
OsiJr, line 265: Removing the joint expansion (even if it is slightly broken) seems like a massive step backwards?

int-Frank: It didn't look like the vertical fence used the expand vector, and I thought we could get away with a similar idea for the horizontal fence. Long story short I've just come back to this PR and I agree we shouldn't the joint expansion.

PR #571, opened 2020-04-27
DEM and Map Height Fixes

peterAdams1014:
Fixed an issue where DEM on/off state was not correctly being applied to tile visibility.
Fixed an issue where global map tile height was not being applied correctly.

Resolves [AB#1512](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1512)

### DIFF
--- a/src/vcQuadTree.cpp
+++ b/src/vcQuadTree.cpp
@@ -125,17 +125,27 @@ void vcQuadTree_CleanupNode(vcQuadTreeNode *pNode)
   memset(pNode, 0, sizeof(vcQuadTreeNode));
 }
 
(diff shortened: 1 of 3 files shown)


src/vcQuadTree.cpp
```diff
@@ -476,7 +507,7 @@ void vcQuadTree_Update(vcQuadTree *pQuadTree, const vcQuadTreeViewInfo &viewInfo
       vcQuadTree_CalculateNodeBounds(pQuadTree, pNode);
     }
   }
-
+
```
samsinsane, line 510: Unnecessary whitespace here.

PR #579, opened 2020-04-28
updated blending for lines, fences and images

int-Frank:
[AB#1579](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1579)


src/gl/directx11/shaders/fenceFragmentShader.hlsl
```diff
@@ -36,6 +36,5 @@ PS_OUTPUT main(PS_INPUT input)
   float halfFcoef = 1.0 / log2(s_CameraFarPlane + 1.0);
   output.Depth0 = log2(input.fLogDepth.x) * halfFcoef;
 
-  output.Color0.a = output.Depth0; // depth packed here
```
OsiJr, line 39: I don't think removing these is correct- this takes as a step further from fences being selectable in the scene? @peterAdams1014 ?

PR #582, opened 2020-04-30
Using Paint 3D.

asunawu:
[AB#1543](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1543)


asunawu:
where is the original file? We could update it. If it won't cause a problem, why don't we merge it and make it better before we have a new texture.

samsinsane:
> where is the original file?

I'm not sure, @OsiJr would have a better idea, I hope!

> We could update it. If it won't cause a problem, why don't we merge it and make it better before we have a new texture.

I might be misunderstanding what you're saying here. If you're suggesting we merge this PR, I'm not against that :) just wanted to comment that we should make the same changes in the original before we forget about it. If you're suggesting we merge the original in, I disagree, those files tend to be several megabytes and we can't load them - I'm assuming that it's a PSD, but it could be any of the Adobe formats.

OsiJr:
The PNG is the primary version at the moment; I just looked at the source psd and its just each icon on its own layer as a raster image which can be recreated quickly from the PNG.

PR #585, opened 2020-05-01
Changed the axis's ends to orange and yellow as an optimization optio…

asunawu:
For [AB#1609](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1609).
I'd like to keep the original logic of axis's colour. So I changed the axis ends to yellow/orange to make it more visible. However, I submitted it as an optimazition option.


OsiJr:
I'm not sure this was worth fixing, @peterAdams1014 ? It was grey because it was disabled? Its full colour when its enabled and usable?

OsiJr:
I think the existing colours are fine. The bug relates to not being able to easily see the gizmo when the gizmo is disabled.

PR #597, opened 2020-05-06
Fixed fence orientation in ECEF

int-Frank:
[AB#1514](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1514)

EDIT: I have changed how horizontal paths behave slightly. If all points of the path lie close to the same horizontal plane, there is no real change compared to the previous implementation. If the points in the path have vastly different heights, the path will still look like a path when viewed from above; previously this was not the case. The difference is, in this case it may look a little warped when viewed from the side. 

My motivation was, we can use screen lines now for random points in a scene. Horizontal paths can now be tailored for a horizontal set of points, or at least render a horizontal path when viewed from above.


src/gl/directx11/shaders/fenceVertexShader.hlsl
```diff
@@ -25,15 +25,22 @@ cbuffer u_EveryFrame : register(b0)
 {
   float4 u_bottomColour;
   float4 u_topColour;
+  float4 u_worldUp; //can represent both vertical and horizontal orientations
```
OsiJr, line 28: Why not calculate expand vector on CPU and avoid all the work in the shader?

int-Frank: Oh this comment shouldn't be here. The expand vectors are still created cpu side. We have a vertex-specific expand vector and a global world up. The shader just switches between the two depending on whether we want to render fences or paths.

PR #598, opened 2020-05-06
Non-geolocated UDS models when dropped on the scene now get placed at mouse

OsiJr:
(or camera) position

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -443,9 +443,25 @@ void vcMain_MainLoop(vcState *pProgramState)
 
               continue;
             }
-            else if (firstLoad) // Was successful
+            else
             {
-              udStrcpy(pProgramState->sceneExplorer.movetoUUIDWhenPossible, pNode->UUID);
+              if (firstLoad) // Was successful
+                udStrcpy(pProgramState->sceneExplorer.movetoUUIDWhenPossible, pNode->UUID);
+
+              // Let is know about the mouse position- using bounding box currently
+              //TODO: Don't use the boundingBox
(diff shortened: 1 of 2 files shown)


src/scene/vcModel.cpp
```diff
@@ -127,6 +131,11 @@ vcModel::vcModel(vcProject *pProject, vdkProjectNode *pNode, vcState *pProgramSt
     pLoadInfo->pModel = this;
     pLoadInfo->pProgramState = pProgramState;
 
+    if (!pNode->hasBoundingBox)
```
int-Frank, line 134: Should this be `if (pNode->hasBoundingBox)` ?

OsiJr: This is correct- if the node _doesn't_ have a bounding box then I've thrown data in the `boundingBox` struct item; if it -does have one (from disk) then I don't want to use those values

src/main.cpp
```diff
@@ -443,9 +443,25 @@ void vcMain_MainLoop(vcState *pProgramState)
...
               continue;
             }
-            else if (firstLoad) // Was successful
+            else
             {
-              udStrcpy(pProgramState->sceneExplorer.movetoUUIDWhenPossible, pNode->UUID);
+              if (firstLoad) // Was successful
+                udStrcpy(pProgramState->sceneExplorer.movetoUUIDWhenPossible, pNode->UUID);
+
+              // Let is know about the mouse position- using bounding box currently
```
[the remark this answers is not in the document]
OsiJr: `it` I think

PR #599, opened 2020-05-06
Fixed dereferencing a null ptr

int-Frank:
[AB#521](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/521)

### DIFF
--- a/src/scene/vcModel.cpp
+++ b/src/scene/vcModel.cpp
@@ -312,7 +312,14 @@ void vcModel::HandleImGui(vcState *pProgramState, size_t * /*pItemID*/)
   if (repackMatrix)
   {
     m_sceneMatrix = udDouble4x4::translation(m_pivot) * udDouble4x4::rotationQuat(orientation, position) * udDouble4x4::scaleUniform(scale.x) * udDouble4x4::translation(-m_pivot);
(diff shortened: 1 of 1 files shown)


src/scene/vcModel.cpp
```diff
@@ -312,7 +312,14 @@ void vcModel::HandleImGui(vcState *pProgramState, size_t * /*pItemID*/)
   if (repackMatrix)
   {
     m_sceneMatrix = udDouble4x4::translation(m_pivot) * udDouble4x4::rotationQuat(orientation, position) * udDouble4x4::scaleUniform(scale.x) * udDouble4x4::translation(-m_pivot);
-    vcProject_UpdateNodeGeometryFromCartesian(&pProgramState->activeProject, m_pNode, *m_pCurrentZone, vdkPGT_Point, &position, 1);
+
+    if (m_pCurrentZone != nullptr)
+      vcProject_UpdateNodeGeometryFromCartesian(&pProgramState->activeProject, m_pNode, *m_pCurrentZone, vdkPGT_Point, &position, 1);
+    else if (m_pPreferredProjection != nullptr)
+      vcProject_UpdateNodeGeometryFromCartesian(&pProgramState->activeProject, m_pNode, *m_pPreferredProjection, vdkPGT_Point, &position, 1);
```
OsiJr, line 319: I think this will turn out to be not required as we should be in `pProgramState->geozone` at this point anyway

PR #603, opened 2020-05-06
OpenGL Terrain Picking

peterAdams1014:
Fixed an issue in OpenGL where the depth copy wasn't correctly being correctly scaled and biased.

### DIFF
--- a/src/gl/directx11/shaders/tileFragmentShader.hlsl
+++ b/src/gl/directx11/shaders/tileFragmentShader.hlsl
@@ -29,6 +29,8 @@ PS_OUTPUT main(PS_INPUT input)
 
   output.Color0 = float4(col.xyz * input.colour.xyz, input.colour.w);
   
-  output.Color0.a = input.depth.x / input.depth.y; // depth packed here
+  float scale = 1.0 / (u_clipZFar - u_clipZNear);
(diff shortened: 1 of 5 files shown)


src/gl/directx11/shaders/tileFragmentShader.hlsl
```diff
@@ -29,6 +29,8 @@ PS_OUTPUT main(PS_INPUT input)
 
   output.Color0 = float4(col.xyz * input.colour.xyz, input.colour.w);
   
-  output.Color0.a = input.depth.x / input.depth.y; // depth packed here
+  float scale = 1.0 / (u_clipZFar - u_clipZNear);
+  float bias = -(u_clipZNear * 0.5);
+  output.Color0.a = (input.depth.x / input.depth.y) * scale + bias; // depth packed here
```
samsinsane, line 34: Does this need to be done in the other shaders that pack the depth into the alpha channel?

peterAdams1014: Nope!

This is because tiles don't write their depth explicitly in the fragment shader. Tiles are 'tesselated enough' to be able to write the LogZ depth per vertex, then interpolate. This allows the hardware to perform normal depth optimizations, while still giving us a logarithmic distribution :).

PR #624, opened 2020-05-08
Add ECEF Toggle Button

Jossos:
- Added ECEF Button to Map Settings Panel
- Fixes [AB#1507](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1507)

### DIFF
--- a/src/vcSettingsUI.cpp
+++ b/src/vcSettingsUI.cpp
@@ -942,19 +942,16 @@ void vcSettingsUI_BasicMapSettings(vcState *pProgramState)
   }
 
   if (ImGui::Button(vcString::Get("settingsMapECEFMode")))
-  {
-    static const int32_t s_ECEFModeID = 4978;
-    static int32_t s_previousSRID = -1;
-    
-    if (pProgramState->geozone.srid == s_ECEFModeID)
+  {    
+    if (pProgramState->geozone.srid == vcPSZ_WGS84ECEF)
     {
-      if (s_previousSRID != -1)
-        pProgramState->geozone.srid = s_previousSRID;
+      if (pProgramState->previousSRID != -1)
+        pProgramState->geozone.srid = pProgramState->previousSRID;
     }
     else
     {
-      s_previousSRID = pProgramState->geozone.srid;
-      pProgramState->geozone.srid = s_ECEFModeID;
+      pProgramState->previousSRID = pProgramState->geozone.srid;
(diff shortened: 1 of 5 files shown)


src/vcSettingsUI.cpp
```diff
@@ -940,6 +940,23 @@ void vcSettingsUI_BasicMapSettings(vcState *pProgramState)
     vcSettings_ApplyMapChange(&pProgramState->settings);
     vcRender_ClearTiles(pProgramState->pRenderContext);
   }
+
+  if (ImGui::Button(vcString::Get("settingsMapECEFMode")))
+  {
+    static const int32_t s_ECEFModeID = 4978;
+    static int32_t s_previousSRID = -1;
```
OsiJr, line 947: This variable should live in `vcState` or `vcProject` rather than here, that way a project change can update it as well so that it isn't carried through logout or changing projects.

src/vcSettingsUI.cpp
```diff
@@ -940,6 +940,23 @@ void vcSettingsUI_BasicMapSettings(vcState *pProgramState)
     vcSettings_ApplyMapChange(&pProgramState->settings);
     vcRender_ClearTiles(pProgramState->pRenderContext);
   }
+
+  if (ImGui::Button(vcString::Get("settingsMapECEFMode")))
+  {
+    static const int32_t s_ECEFModeID = 4978;
```
OsiJr, line 946: This constant fits better as a value in the `vcProjectStandardZones` enum- `vcPSZ_WGS84ECEF`.

There are other places I've used `4978` as a magic number that will need to change eventually as well but that's outside the scope of this PR.

src/vcSettingsUI.cpp
```diff
@@ -940,6 +940,23 @@ void vcSettingsUI_BasicMapSettings(vcState *pProgramState)
...
+    static int32_t s_previousSRID = -1;
+    
+    if (pProgramState->geozone.srid == s_ECEFModeID)
+    {
+      if (s_previousSRID != -1)
+        pProgramState->geozone.srid = s_previousSRID;
```
OsiJr, line 952: Changing zones is trickier than just changing the `srid` value.

```
udGeoZone zone = {};
udGeoZone_SetFromSRID(&zone, s_previousSRID);
vcGIS_ChangeSpace(&pProgramState->geozone, zone, &pProgramState->camera.position);
pProgramState->activeProject.pFolder->ChangeProjection(zone);
```

PR #625, opened 2020-05-08
Added MRT support to vcGL

peterAdams1014:


peterAdams1014:
This doesn't add MRT for Metal - we will resolve this later.

src/gl/vcFramebuffer.h
```diff
@@ -16,7 +16,7 @@ enum vcFramebufferClearOperation
 
 struct vcFramebuffer;
 
-bool vcFramebuffer_Create(vcFramebuffer **ppFramebuffer, vcTexture *pTexture, vcTexture *pDepth = nullptr, uint32_t level = 0);
+bool vcFramebuffer_Create(vcFramebuffer **ppFramebuffer, vcTexture *pTexture, vcTexture *pDepth = nullptr, uint32_t level = 0, vcTexture *pAttachment2 = nullptr);
```
samsinsane, line 19: Should we perhaps consider creating a `vcFramebuffer_CreateAdv` which takes an array of attachments which indicate what kind of attachment they are and the `vcTexture*` to use?
```c++
enum vcFramebufferAttachmentType
{
  vcFAT_Colour,
  vcFAT_Depth
};

struct vcFramebufferAttachment
{
  vcTexture *pTexture;
  vcFramebufferAttachmentType type;
};
```

src/gl/opengl/vcFramebuffer.cpp
```diff
@@ -1,13 +1,13 @@
...
 
-bool vcFramebuffer_Create(vcFramebuffer **ppFramebuffer, vcTexture *pTexture, vcTexture *pDepth /*= nullptr*/, uint32_t level /*= 0*/)
+bool vcFramebuffer_Create(vcFramebuffer **ppFramebuffer, vcTexture *pTexture, vcTexture *pDepth /*= nullptr*/, uint32_t level /*= 0*/, vcTexture *pAttachment2 /*= nullptr*/)
 {
   if (ppFramebuffer == nullptr || pTexture == nullptr)
     return false;
 
   udResult result = udR_Success;
-  static const GLenum DrawBuffers[] = { GL_COLOR_ATTACHMENT0 };
+  static const GLenum DrawBuffers[] = { GL_COLOR_ATTACHMENT0, GL_COLOR_ATTACHMENT1 };
```
samsinsane, line 10: I vaguely recall it being suggested that these should be done as `GL_COLOR_ATTACHMENT0 + N` instead - I think it was due to the other values not always being defined or something? It's possible I'm thinking of a different OpenGL enum set?

peterAdams1014: You are correct. I don't mind either way to be honest.

src/gl/vcFramebuffer.h
peterAdams1014: Yeah sounds good - this change would be consistent with the vcTexture_CreateAdv() API. Probably won't make this change for this PR though.

samsinsane: > Probably won't make this change for this PR though.

Minor issue with going down this path is that `vcFramebuffer_Create` will need to have this change undone, and all usages moved to use `vcFramebuffer_CreateAdv`.

PR #631, opened 2020-05-11
Fix for Z not being factored in for culling purposes

OsiJr:
Fixes [AB#1517](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1517)


src/vcRender.cpp
```diff
@@ -804,29 +804,37 @@ void vcRenderTerrain(vcState *pProgramState, vcRenderContext *pRenderContext, co
     double visibleFarPlane = udMin((double)s_CameraFarPlane, BaseViewDistance + cameraHeightAboveEarthSurface * HeightViewDistanceScale);
 
     // Corners [nw, ne, sw, se]
```
samsinsane, line 806: This comment needs to be updated as there are now 8 "corners"

PR #635, opened 2020-05-12
Minor polygon fix and SceneLayerRenderer improvements.

peterAdams1014:
Better abstraction with vcSceneLayerRenderer.
Fixed a crash when loading a model with broken normal indices.

### DIFF
--- a/src/rendering/vcSceneLayerRenderer.cpp
+++ b/src/rendering/vcSceneLayerRenderer.cpp
@@ -7,8 +7,7 @@
 #include "gl/vcTexture.h"
 #include "gl/vcGLState.h"
 
-struct vcSceneLayerRendererInternal :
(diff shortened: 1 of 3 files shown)


src/rendering/vcSceneLayerRenderer.cpp
```diff
@@ -7,17 +7,33 @@
 #include "gl/vcTexture.h"
 #include "gl/vcGLState.h"
 
+struct vcSceneLayerRendererInternal :
+  public vcSceneLayerRenderer
```
samsinsane, line 11: One line, and is the `public` required?

peterAdams1014: It is not. Fixed.

PR #642, opened 2020-05-13
Change Password box

Jossos:
- Added box to allow user to change password
- Fixes [AB#1639](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1639)


src/vcModals.cpp
```diff
@@ -619,6 +619,83 @@ void vcModals_DrawProfile(vcState* pProgramState)
...
+      ImGui::InputText(vcString::Get("modalProfileReEnterNewPassword"), &pProgramState->changePassword.newPasswordConfirm[0], vcMaxPathLength, passwordFieldFlags);
+    }
+
+    ImGui::Separator();
+    
+    if (ImGui::Button(vcString::Get("modalProfileConfirmNewPassword"), ImVec2(130.0f, 0)) || vcHotkey::IsPressed(vcB_Cancel))
+    {
+      udJSON changePasswordData;
+      changePasswordData.Set("username = \"%s\"", pProgramState->settings.loginInfo.username);
+      changePasswordData.Set("oldpassword = \"%s\"", pProgramState->changePassword.currentPassword);
```
samsinsane, line 658: If these contain a `"` this will break, you need to use this pattern:
```c++
udJSON temp;
temp.SetString(string);
changePasswordData.Set(&tempNode, "key");
```

src/vcModals.cpp
```diff
@@ -619,6 +619,83 @@ void vcModals_DrawProfile(vcState* pProgramState)
...
+      changePasswordData.Set("password = \"%s\"", pProgramState->changePassword.newPassword);
+      changePasswordData.Set("passwordConfirm = \"%s\"", pProgramState->changePassword.newPasswordConfirm);
+
+      const char *pUpdatePasswordString = nullptr;
+      changePasswordData.Export(&pUpdatePasswordString, udJEO_JSON);
```
samsinsane, line 663: `udJEO_JSON` is the default value, usually you only need to specify it when you want the `udJEO_FormatWhiteSpace` option too.

src/vcModals.cpp
```diff
@@ -619,6 +619,83 @@ void vcModals_DrawProfile(vcState* pProgramState)
...
+        {
+          ImGui::CloseCurrentPopup();
+          memset(pProgramState->changePassword.currentPassword, 0, sizeof(pProgramState->changePassword.currentPassword));
+          memset(pProgramState->changePassword.newPassword, 0, sizeof(pProgramState->changePassword.newPassword));
+          memset(pProgramState->changePassword.newPasswordConfirm, 0, sizeof(pProgramState->changePassword.newPasswordConfirm));
+          memset(pProgramState->changePassword.message, 0, sizeof(pProgramState->changePassword.message));
+        }
+        else
+        {
+          udStrcpy(pProgramState->changePassword.message, resultData.Get("message").AsString());
```
samsinsane, line 682: These messages are going to need to be translated. I would recommend using `udStrEqual` to determine if they're the string you expect, and then substitute in a translation key. When it's none of the known strings, display unknown error.

src/vcModals.cpp
```diff
@@ -619,6 +619,83 @@ void vcModals_DrawProfile(vcState* pProgramState)
...
+
+    if (pProgramState->sessionInfo.isOffline)
+    {
+      ImGui::TextUnformatted(vcString::Get("modalProfileOffline"));
+    }
+    else
+    {
+      ImGuiInputTextFlags passwordFieldFlags = ImGuiInputTextFlags_Password;
+
+      ImGui::InputText(vcString::Get("modalProfileCurrentPassword"), &pProgramState->changePassword.currentPassword[0], vcMaxPathLength, passwordFieldFlags);
```
samsinsane, line 646: The usage of `&[0]` isn't necessary, arrays will decay to pointers.

src/vcModals.cpp
```diff
@@ -619,6 +619,83 @@ void vcModals_DrawProfile(vcState* pProgramState)
...
+  {
+    pProgramState->modalOpen = true;
+
+    if (pProgramState->sessionInfo.isOffline)
+    {
+      ImGui::TextUnformatted(vcString::Get("modalProfileOffline"));
+    }
+    else
+    {
+      ImGuiInputTextFlags passwordFieldFlags = ImGuiInputTextFlags_Password;
```
samsinsane, line 644: This seems kind of unnecessary to me, do we do this elsewhere?

src/vcModals.cpp
```diff
@@ -619,6 +619,83 @@ void vcModals_DrawProfile(vcState* pProgramState)
   }
 }
 
+void vcModals_DrawChangePassword(vcState *pProgramState)
+{
+  const char *profile = vcString::Get("modalChangePasswordTitle");
```
samsinsane, line 624: This is a pointer, `pProfile`.

src/vcModals.cpp
```diff
@@ -619,6 +619,83 @@ void vcModals_DrawProfile(vcState* pProgramState)
...
+      }
+    }
+    
+    ImGui::SameLine();
+
+    if (ImGui::Button(vcString::Get("popupClose"), ImVec2(buttonWidth, 0)) || vcHotkey::IsPressed(vcB_Cancel))
+      ImGui::CloseCurrentPopup();
+
+    ImGui::SameLine();
+    ImGui::TextColored(ImVec4(1, 0, 0, 1), pProgramState->changePassword.message);
```
samsinsane, line 693: This is the line failing the builds on Android and Linux, it needs to be `"%s", pProgramState->changePassword.message`

src/vcModals.cpp
Jossos: what's tempNode?

Jossos: This is fixed in latest push. not sure why there's no "outdated" tag on it.

src/vcModals.cpp
Jossos: so the string result that comes out of vdkServerAPI_Query will always be the english text? If so, I can have something that does a strcmp with the expected text, and if they match, just use the text from the language enAU.json file

src/vcModals.cpp
samsinsane: The `outdated` tag only applies to lines that are modified.

src/vcModals.cpp
Jossos: This is fixed in latest push. not sure why there's no "outdated" tag on it.

src/vcModals.cpp
```diff
@@ -619,6 +619,103 @@ void vcModals_DrawProfile(vcState* pProgramState)
...
+      //changePasswordData.Set()
+      changePasswordData.Set("username = \"%s\"", pProgramState->settings.loginInfo.username);
+      changePasswordData.Set("oldpassword = \"%s\"", pProgramState->changePassword.currentPassword);
+      changePasswordData.Set("password = \"%s\"", pProgramState->changePassword.newPassword);
+      changePasswordData.Set("passwordConfirm = \"%s\"", pProgramState->changePassword.newPasswordConfirm);
+
+      const char *pUpdatePasswordString = nullptr;
+      changePasswordData.Export(&pUpdatePasswordString);
+
+      const char *pResult = nullptr;
```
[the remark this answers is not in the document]
Jossos: this is freed in latest push

src/vcModals.cpp
```diff
@@ -619,6 +619,103 @@ void vcModals_DrawProfile(vcState* pProgramState)
...
+      temp.SetString(pProgramState->changePassword.newPasswordConfirm);
+      changePasswordData.Set(&temp, "passwordConfirm");
+
+      //changePasswordData.Set()
+      changePasswordData.Set("username = \"%s\"", pProgramState->settings.loginInfo.username);
+      changePasswordData.Set("oldpassword = \"%s\"", pProgramState->changePassword.currentPassword);
+      changePasswordData.Set("password = \"%s\"", pProgramState->changePassword.newPassword);
+      changePasswordData.Set("passwordConfirm = \"%s\"", pProgramState->changePassword.newPasswordConfirm);
+
+      const char *pUpdatePasswordString = nullptr;
```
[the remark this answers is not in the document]
Jossos: this is freed in latest push

src/vcModals.cpp
```diff
@@ -619,6 +619,103 @@ void vcModals_DrawProfile(vcState* pProgramState)
...
+      changePasswordData.Export(&pUpdatePasswordString);
+
+      const char *pResult = nullptr;
+      vdkError result = vdkServerAPI_Query(pProgramState->pVDKContext, "v1/user/updatepassword", pUpdatePasswordString, &pResult);
```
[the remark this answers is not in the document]
Jossos: this is freed in latest push

PR #643, opened 2020-05-13
Fixed screen line issues

int-Frank:
[AB#1626](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1626)

Notes:
- Line clipping added
- Line flickering fixed.
- Line end point 'dancing' fixed
- Line end points now proper thickness
- Line 'mitre' joins removed. Mitre joins don't really work for screen lines; too often we see joins at such small angles it looks weird. Perhaps we could use round or bevel joins, but this is outside the scope of this PR.

### DIFF
--- a/src/rendering/vcLineRenderer.cpp
+++ b/src/rendering/vcLineRenderer.cpp
@@ -7,7 +7,7 @@ struct vcLineInstance
 {
   udFloat4 colour;
   float width;
-  size_t pointCount;
+  size_t segmentCount;
   vcMesh *pMesh;
   udDouble4x4 originMatrix;
 };
@@ -24,6 +24,7 @@ struct vcLineRenderer
     {
       udFloat4 colour;
       udFloat4 thickness;
+      udFloat4 nearPlane;
       udFloat4x4 worldViewProjection;
     } everyObjectParams;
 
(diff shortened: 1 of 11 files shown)


src/rendering/vcLineRenderer.cpp
```diff
@@ -99,62 +100,87 @@ udResult vcLineRenderer_CreateLine(vcLineInstance **ppLine)
...
+    udDouble3 p0 = pPoints[i] - origin;
+    udDouble3 p1 = pPoints[(i + 1) % pointCount] - origin;
+    udDouble4 p0_neighbour = udDouble4::create(pPoints[(i + pointCount - 1) % pointCount] - origin, 1.0);
+    udDouble4 p1_neighbour = udDouble4::create(pPoints[(i + 2) % pointCount] - origin, 1.0);
+
+    if (i == 0 && closed)
+      p0_neighbour.w = 0.0; //Flag endpoint with w == 0
+
+    if (i == pointCount - 2 && closed)
+        p0_neighbour.w = 0.0; //Flag endpoint with w == 0
```
OsiJr, line 141: Can I get a confirm that these are supposed to be `closed`? This behaviour makes more intuitive sense as `!closed`?

int-Frank: Yes, these should be `!closed`

PR #644, opened 2020-05-13
Added vertical measure tool.

asunawu:
[AB#1638](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1638).

Done:

Cancel:


src/vcSettings.cpp
```diff
@@ -359,6 +359,7 @@ bool vcSettings_Load(vcSettings *pSettings, bool forceReset /*= false*/, vcSetti
     vcHotkey::Set(vcB_ToggleMeasureAreaTool, data.Get("keys.%s", vcHotkey::GetBindName(vcB_ToggleMeasureAreaTool)).AsInt(65));
     vcHotkey::Set(vcB_ToggleInspectionTool, data.Get("keys.%s", vcHotkey::GetBindName(vcB_ToggleInspectionTool)).AsInt(66));
     vcHotkey::Set(vcB_ToggleAnnotateTool, data.Get("keys.%s", vcHotkey::GetBindName(vcB_ToggleAnnotateTool)).AsInt(67));
+    vcHotkey::Set(vcB_ToggleMeasureHeightTool, data.Get("keys.%s", vcHotkey::GetBindName(vcB_ToggleMeasureHeightTool)).AsInt(68));
```
samsinsane, line 362: F11 is Fullscreen in Chrome, so this keybinding will need to change.

@OsiJr mentioned having an idea for not using the function keys? It would also be nice to see the `SDL_SCANCODE_` values used instead of the hardcoded numbers, and where appropriate the `vcMOD_` value ideally OR'd in.

OsiJr: The function keys are causing headaches everywhere but outside the scope of this MR to fix it.

src/scene/vcPOI.cpp
```diff
@@ -295,6 +295,90 @@ class vcPOIState_MeasureArea : public vcPOIState_General
...
+        m_pParent->InsertPoint(udDouble3::zero());
+        m_pParent->InsertPoint(udDouble3::zero());
+      }
+
+      if (startPoint.z > position.z)
+        m_pParent->m_line.pPoints[1] = udDouble3::create(position.x, position.y, startPoint.z);
+      else
+        m_pParent->m_line.pPoints[1] = udDouble3::create(startPoint.x, startPoint.y, position.z);
+
+      m_pParent->m_line.pPoints[2] = position;
```
OsiJr, line 343: This won't work in ECEF; it needs to be in the `Up` direction

PR #650, opened 2020-05-14
Added a proxy position to dump position values

int-Frank:
[AB#1613](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1613)

This root of this issue was floating point accuracy. Essentially, the real position of the model cannot be represented in the input UI (due to there's lots of packing/unpacking of matrices, conversions from Cartesian to lat long etc... ). Since the input buffer is different to resulting actual position which we init the input buffer with, it is flagged as 'changed' each frame, thus we see the numbers change each frame. So I added a proxy position object, and update the model position whenever the proxy changes.


OsiJr:
This fix unfortunately breaks the gizmo and means that changing zones places the model in whoop-whoop

PR #651, opened 2020-05-15
POI Hyperlinks

Jossos:
Added Hyperlinks to POI's
Fixes [AB#1642](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1642)

### DIFF
--- a/src/scene/vcPOI.cpp
+++ b/src/scene/vcPOI.cpp
@@ -879,8 +879,8 @@ void vcPOI::HandleImGui(vcState *pProgramState, size_t *pItemID)
   if (vcIGSW_InputText(vcString::Get("scenePOILabelHyperlink"), m_hyperlink, 512, ImGuiInputTextFlags_EnterReturnsTrue))
     vdkProjectNode_SetMetadataString(m_pNode, "hyperlink", m_hyperlink);
 
-  if (ImGui::Button(vcString::Get("scenePOILabelOpenHyperlink")))
-    vcWebFile_OpenBrowser((const char *)m_hyperlink);
+  if (m_hyperlink[0] != '\0' && ImGui::Button(vcString::Get("scenePOILabelOpenHyperlink")))
+    vcWebFile_OpenBrowser(m_hyperlink);
   
   if (m_attachment.pModel != nullptr)
   {
--- a/src/scene/vcPOI.cpp
+++ b/src/scene/vcPOI.cpp
@@ -876,12 +876,25 @@ void vcPOI::HandleImGui(vcState *pProgramState, size_t *pItemID)
     vdkProjectNode_SetMetadataString(m_pNode, "textSize", pTemp);
   }
 
-  if (vcIGSW_InputText(vcString::Get("scenePOILabelHyperlink"), m_hyperlink, 512, ImGuiInputTextFlags_EnterReturnsTrue))
+  if (vcIGSW_InputText(vcString::Get("scenePOILabelHyperlink"), m_hyperlink, vcMaxPathLength, ImGuiInputTextFlags_EnterReturnsTrue))
     vdkProjectNode_SetMetadataString(m_pNode, "hyperlink", m_hyperlink);
 
   if (m_hyperlink[0] != '\0' && ImGui::Button(vcString::Get("scenePOILabelOpenHyperlink")))
     vcWebFile_OpenBrowser(m_hyperlink);
-  
+
+  // Handle hyperlinks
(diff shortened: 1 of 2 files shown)


src/scene/vcPOI.cpp
```diff
@@ -873,19 +876,12 @@ void vcPOI::HandleImGui(vcState *pProgramState, size_t *pItemID)
...
-      ImGui::SameLine();
-      if (ImGui::Button(vcString::Get("scenePOILabelOpenHyperlink")))
-        pProgramState->pLoadImage = udStrdup(pHyperlink);
-    }
-  }
+  if (vcIGSW_InputText(vcString::Get("scenePOILabelHyperlink"), m_hyperlink, 512, ImGuiInputTextFlags_EnterReturnsTrue))
+    vdkProjectNode_SetMetadataString(m_pNode, "hyperlink", m_hyperlink);
 
+  if (ImGui::Button(vcString::Get("scenePOILabelOpenHyperlink")))
+    vcWebFile_OpenBrowser((const char *)m_hyperlink);
```
OsiJr, line 883: This cast shouldn't be required?

src/scene/vcPOI.h
```diff
@@ -46,7 +46,8 @@ class vcPOI : public vcSceneItem
   uint32_t m_nameColour;
   uint32_t m_backColour;
   vcLabelFontSize m_namePt;
-
+  char m_hyperlink[512]; // 512 needs to be eventually put into its own variable
```
samsinsane, line 49: We already have a variable for this, `vcMaxPathLength`.

src/scene/vcPOI.cpp
```diff
@@ -873,19 +876,12 @@ void vcPOI::HandleImGui(vcState *pProgramState, size_t *pItemID)
...
-  // Handle hyperlinks
-  const char *pHyperlink = m_metadata.Get("hyperlink").AsString();
-  if (pHyperlink != nullptr)
-  {
-    ImGui::TextWrapped("%s: %s", vcString::Get("scenePOILabelHyperlink"), pHyperlink);
-    if (udStrEndsWithi(pHyperlink, ".png") || udStrEndsWithi(pHyperlink, ".jpg"))
-    {
-      ImGui::SameLine();
-      if (ImGui::Button(vcString::Get("scenePOILabelOpenHyperlink")))
-        pProgramState->pLoadImage = udStrdup(pHyperlink);
```
samsinsane, line 885: @OsiJr should this continue to happen? I was under the impression that we were specifically requested to load any images in Client?

OsiJr: Images yes; Turns out that PDFs, movies and other miscellany were being linked as well

OsiJr: Actually; yes- you're correct. If we _can_ handle it we should. Ideally these things with Images should be `MediaNodes` though

samsinsane: I see, I believe these come from the UDP parser which predates `MediaNodes`, and it wasn't updated when they were added. Are they definitely the same feature? In Geoverse, they're a label that opens a picture - I think it might have opened through the scene viewer too by clicking on the label.

Jossos: am I good to remove this code or do you guys want it in?

Jossos: ok ive put it back in. just fyi the button never activates since the metadata never has "hyperlink" in it

samsinsane: @Jossos There seems to be a misunderstanding here - we weren't asking for the code to be put back, we were asking for the functionality to return. The functionality being that if the URI is for an image, load the image in Client using the code commented on here, otherwise call out to the OS as you've done in this PR.

src/scene/vcPOI.cpp
```diff
@@ -873,19 +876,17 @@ void vcPOI::HandleImGui(vcState *pProgramState, size_t *pItemID)
...
-    }
+    if (udStrEndsWithi(m_hyperlink, ".png") || udStrEndsWithi(m_hyperlink, ".jpg"))
+      pProgramState->pLoadImage = udStrdup(m_hyperlink);
+    else
+      vcWebFile_OpenBrowser(m_hyperlink);
   }
-
+
```
samsinsane, line 889: If you don't already have the VS extension installed, you need to install EditorConfig - this should sort out these whitespace issues you've had in this PR.

Jossos: fxied i installed codemaid

PR #657, opened 2020-05-18
Fixed accuracy issues with tile normals.

peterAdams1014:
Normals are now calculated for each control point of a tile.
(This significantly reduces the amount of cracking at tile boundaries at certain positions around the globe)


src/gl/directx11/shaders/tileVertexShader.hlsl
```diff
@@ -45,43 +45,43 @@ float CalcuteLogDepth(float4 clipPos)
   return (log2(max(1e-6, 1.0 + clipPos.w)) * Fcoef + u_clipZNear) * clipPos.w;
 }
 
+float4 BiliearSample(float4 samples[CONTROL_POINT_RES * CONTROL_POINT_RES], float2 uv)
```
OsiJr, line 48: `Bilinear`

PR #660, opened 2020-05-18
Fixed floating point drift with model positioning

int-Frank:
[AB#1613](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1613)

Removed m_sceneMatrix and storing position, orientation and scale separately. We just crunch these together when we need the scene matrix, which turns out isn't that often.

### DIFF
--- a/src/scene/vcModel.cpp
+++ b/src/scene/vcModel.cpp
@@ -66,9 +66,9 @@ void vcModel_LoadMetadata(vcState *pProgramState, vcModel *pModel, udDouble3 pos
 
   }
 
-  pModel->m_sceneMatrix = udDouble4x4::create(pModel->m_pointCloudHeader.storedMatrix);
-  pModel->m_defaultMatrix = pModel->m_sceneMatrix;
-  pModel->m_baseMatrix = pModel->m_defaultMatrix;
+  pModel->m_defaultMatrix = udDouble4x4::create(pModel->m_pointCloudHeader.storedMatrix);
+  udDouble4x4 matrix = udDouble4x4::translation(possibleLocation) * pModel->m_defaultMatrix;
+  pModel->SetTransforms(matrix);
 
   if (pModel->m_pPreferredProjection == nullptr && possibleLocation != udDouble3::zero())
     vcProject_UpdateNodeGeometryFromCartesian(pModel->m_pProject, pModel->m_pNode, pProgramState->geozone, vdkPGT_Point, &possibleLocation, 1);
@@ -116,8 +116,9 @@ vcModel::vcModel(vcProject *pProject, vdkProjectNode *pNode, vcState *pProgramSt
   m_pivot(udDouble3::zero()),
   m_defaultMatrix(udDouble4x4::identity()),
   m_pCurrentZone(nullptr),
-  m_sceneMatrix(udDouble4x4::identity()),
-  m_baseMatrix(udDouble4x4::identity()),
+  m_positionCurrent(udDouble3::zero()),
+  m_scaleCurrent(udDouble3::create(1.0, 1.0, 1.0)),
+  m_orientationCurrent(udDoubleQuat::create(0.0, 0.0, 0.0)),
   m_changeZones(false),
   m_meterScale(0.0),
   m_hasWatermark(false),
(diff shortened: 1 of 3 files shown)


src/scene/vcModel.cpp
```diff
@@ -758,10 +689,33 @@ udDouble3 vcModel::GetLocalSpacePivot()
...
 {
   return vcGAC_AllUniform;
 }
+
+void vcModel::SetTransforms(const udDouble4x4 &matrix)
+{
+  (udDouble4x4::translation(-m_pivot) * matrix * udDouble4x4::translation(m_pivot)).extractTransforms(m_positionCurrent, m_scaleCurrent, m_orientationCurrent);
+}
+
+void vcModel::SaveTransformsToNode(vcProject * pProject, const udGeoZone &defaultZone)
```
OsiJr, line 705: I think you have your visual studio settings wrong, your `*` always ends up in the middle instead of on the right

src/scene/vcModel.cpp
```diff
@@ -758,10 +689,33 @@ udDouble3 vcModel::GetLocalSpacePivot()
...
 }
 
 vcGizmoAllowedControls vcModel::GetAllowedControls()
 {
   return vcGAC_AllUniform;
 }
+
+void vcModel::SetTransforms(const udDouble4x4 &matrix)
+{
+  (udDouble4x4::translation(-m_pivot) * matrix * udDouble4x4::translation(m_pivot)).extractTransforms(m_positionCurrent, m_scaleCurrent, m_orientationCurrent);
```
OsiJr, line 702: Splitting this over 2 lines will make it more clear what it does.

src/scene/vcModel.cpp
```diff
@@ -758,10 +689,33 @@ udDouble3 vcModel::GetLocalSpacePivot()
...
+}
+
+void vcModel::SaveTransformsToNode(vcProject * pProject, const udGeoZone &defaultZone)
+{
+  if (m_pCurrentZone != nullptr)
+    vcProject_UpdateNodeGeometryFromCartesian(pProject, m_pNode, *m_pCurrentZone, vdkPGT_Point, &m_positionCurrent, 1);
+  else if (m_pPreferredProjection != nullptr)
+    vcProject_UpdateNodeGeometryFromCartesian(pProject, m_pNode, *m_pPreferredProjection, vdkPGT_Point, &m_positionCurrent, 1);
+  else
+    vcProject_UpdateNodeGeometryFromCartesian(pProject, m_pNode, defaultZone, vdkPGT_Point, &m_positionCurrent, 1);
```
samsinsane, line 712: I would recommend using something like:
```c++
const udGeoZone *pGeoZone = &defaultZone;
if (m_pCurrentZone != nullptr)
  pGeoZone = m_pCurrentZone;
else if (m_pPreferredProjection != nullptr)
  pGeoZone = m_pPreferredProjection;

vcProject_UpdateNodeGeometryFromCartesion(pProject, m_pNode, *pGeoZone, vdkPGT_Point, &m_positionCurrent, 1);
```
If any of the parameters need to change, or should change, this ensures that bugs aren't accidentally introduced because one of them wasn't updated. Additionally, it makes it abundantly clear to the reader exactly what changes.

src/scene/vcModel.cpp
```diff
@@ -195,37 +197,6 @@ void vcModel::AddToScene(vcState *pProgramState, vcRenderData *pRenderData)
...
-
-    vdkProjectNode_GetMetadataDouble(m_pNode, "transform.rotation.y", &euler.x, 0.0);
-    vdkProjectNode_GetMetadataDouble(m_pNode, "transform.rotation.p", &euler.y, 0.0);
-    vdkProjectNode_GetMetadataDouble(m_pNode, "transform.rotation.r", &euler.z, 0.0);
-    vdkProjectNode_GetMetadataDouble(m_pNode, "transform.scale", &scale, udMag3(m_baseMatrix.axis.x));
-
-    m_sceneMatrix = udDouble4x4::translation(m_pivot) * udDouble4x4::rotationYPR(UD_DEG2RAD(euler), *pPosition) * udDouble4x4::scaleUniform(scale) * udDouble4x4::translation(-m_pivot);
-  }
-
-  udFree(pPosition);
```
OsiJr, line 227: These aren't being loaded from disk? That will mean that models aren't saved/loaded correctly.

src/scene/vcModel.cpp
int-Frank: Gah, old habits die hard...

src/scene/vcModel.cpp
int-Frank: Ok

src/scene/vcModel.cpp
int-Frank: I like it

PR #670, opened 2020-05-19
Added POI descriptions

Jossos:
Added Descriptions for POI's. (Doesn't display the description, only exists in the Scene Explorer)

### DIFF
--- a/src/scene/vcPOI.cpp
+++ b/src/scene/vcPOI.cpp
@@ -1152,18 +1152,6 @@ void vcPOI::AddLabelsToScene(vcRenderData *pRenderData)
       m_pLabelInfo->pText = m_pNode->pName;
 
     pRenderData->labels.PushBack(m_pLabelInfo);
-    
-    // Add Description
-    if (m_description[0] != '\0')
-    {
-      m_descriptionLabel.pText = m_description;
-      m_descriptionLabel.textSize = m_pLabelInfo->textSize;
-      m_descriptionLabel.textColourRGBA = m_pLabelInfo->textColourRGBA;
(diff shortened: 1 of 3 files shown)


src/scene/vcPOI.cpp
```diff
@@ -892,6 +895,9 @@ void vcPOI::HandleImGui(vcState *pProgramState, size_t *pItemID)
       vcWebFile_OpenBrowser(m_hyperlink);
   }
 
+  if (vcIGSW_InputText(vcString::Get("scenePOILabelDescription"), m_description, 512, ImGuiInputTextFlags_EnterReturnsTrue))
```
samsinsane, line 898: `udLengthOf(m_description)` instead of `512`.

src/scene/vcPOI.cpp
```diff
@@ -776,6 +779,9 @@ void vcPOI::HandleImGui(vcState *pProgramState, size_t *pItemID)
       vcWebFile_OpenBrowser(m_hyperlink);
   }
 
+  if (vcIGSW_InputText(vcString::Get("scenePOILabelDescription"), m_description, sizeof(m_description), ImGuiInputTextFlags_EnterReturnsTrue))
```
samsinsane, line 782: For future reference, the reason I said `udLengthOf` and not `sizeof` is that `sizeof` will work on every variable. If `m_description` changes to a pointer instead, this code will continue to compile. `udLengthOf` only works on arrays, so this line would fail to compile.

Jossos: sounds good

PR #672, opened 2020-05-19
Fixed crash moving non-geolocated models

Jossos:
- Fixed Moving of non-geolocated models causing a crash
- Fixes [AB#1655](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1655)


samsinsane:
While your changes might stop the crash from occurring, the crash was highlighting a problem with non-geolocated models not being handled correctly. Your solution is wrong. You've ensured that exporting and re-importing a non-geolocated project won't work if a model is moved with the gizmos. I'm going to close this PR as the correct solution has no overlap with the change you've made.

PR #684, opened 2020-05-25
Added function to calculate a heading between two lat/lon positions

peterAdams1014:


peterAdams1014:
Oops. Updated udCore.

PR #689, opened 2020-05-26
Awu/vertical measurement tool

asunawu:
Spliting class to solve point/label display and gizmo tools problems.


src/main.cpp
```diff
@@ -1276,6 +1282,17 @@ void vcRenderSceneUI(vcState *pProgramState, const ImVec2 &windowPos, const ImVe
...
         }
+        break;
+        case vcActiveTool_MeasureHeight:
+        {
+          ImGui::Separator();
+          ImGui::TextUnformatted(vcString::Get("toolMeasureStart"));
+          ImGui::Separator();
+        }
+        break;
+        default:
```
samsinsane, line 1293: I think it would be better to avoid using `default`, when a new value is added to the enum it will trigger a warning on GCC and Clang (all non-Windows builds). This means everywhere we use the enum values, should fail to build prompting the developer to handle that new value in all the same locations - this has previously been a source of our bugs, especially with the tools/camera.

asunawu: fixed.

PR #690, opened 2020-05-26
Fhart/lidar point attribs

int-Frank:
[AB#1643](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1643)

The basic structure is complete. TODO:
 - init settings
 - add strings to schema
 - tidy up UI
 - test shaders


src/vcSettings.h
```diff
@@ -139,6 +146,29 @@ struct vcVisualizationSettings
...
+  static const uint32_t s_nSegments = 3;
+  struct
+  {
+    uint32_t colours[s_nSegments]; //left, center, right
+    uint32_t errorColor;
+  } scanAngle;
+
+  struct
+  {
+    std::map<uint16_t, uint32_t> colourMap; // id/colour
```
samsinsane, line 164: I believe @OsiJr found that `udChunkedArray` was faster than `std::map`, so it might be worth considering using that instead of `std::map` - from memory, people recommend `std::unordered_map` over `std::map` and nobody recommends `std::unordered_map`, I don't think any of the maps in the STL are actually recommended due to their hashing choices being very slow.

You're also going to leak memory with this because the destructor is never called - there may also be issues unless the STL spec dictates that `std::map` must work without a constructor being called, because this also doesn't call the constructor. You can inplace construct it when the `vcSettings` object is created/initialised, something like `new (&state.visualisation.pointSourceID.colourMap) std::map<uint16_t, uint32_t>();` but I'm not entirely sure, we don't do it very often.

PR #694, opened 2020-05-28
New voxel shaders

int-Frank:
[AB#1643](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1643)

Added new visualisation shaders for
 - GPS time
 - scan angle
 - point source ID
 - return number
 - number of returns

Additional notes: 

vdkSA_ScanDirection, vdkSA_ScanAngle and vdkSA_ScanAngleRank seem to be the same thing. Maybe only have one of these? 

`vdkAttributeSet_GetOffsetOfStandardAttribute()` seems to sometimes fail for the same attribute in different models. I'm not sure why this is but didn't look too much into it as it was really outside the scope of this PR.


src/vcSettings.h
```diff
@@ -124,6 +129,14 @@ struct vcLanguageOption
 
 struct vcVisualizationSettings
 {
+  //These are defined in the LAS 1.4 spec
+  static const int16_t s_scanAngleMin = -30'000; //Represents -180 deg
+  static const int16_t s_scanAngleMax =  30'000; //Represents  180 deg
+  static const uint32_t s_scanAngleRange = s_scanAngleMax - s_scanAngleMin;
```
OsiJr, line 135: Normally static const are treated with the same rules as enums; ProperCase with no prefix.

PR #700, opened 2020-06-03
Updated new visualisations to be per-model

int-Frank:
Fixes #AB1700

### DIFF
--- a/src/scene/vcModel.cpp
+++ b/src/scene/vcModel.cpp
@@ -124,6 +124,8 @@ vcModel::vcModel(vcProject *pProject, vdkProjectNode *pNode, vcState *pProgramSt
   m_pWatermark(nullptr),
   m_visualization()
 {
+  Init(pProgramState);
+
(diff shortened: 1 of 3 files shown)


src/scene/vcModel.cpp
```diff
@@ -281,6 +292,50 @@ void vcModel::OnNodeUpdate(vcState *pProgramState)
     }
   }
 
+  if (vdkProjectNode_GetMetadataDouble(m_pNode, "visualization.GPSTime.minTime", &m_visualization.GPSTime.minTime, pProgramState->settings.visualization.GPSTime.minTime) != vE_Success && vdkProjectNode_GetMetadataDouble(m_pNode, "visualization.GPSTime.maxTime", &m_visualization.GPSTime.maxTime, pProgramState->settings.visualization.GPSTime.maxTime) != vE_Success)
+  {
+    const char *pGPSTime = m_metadata.Get("AttrMinMax_udGPSTime").AsString();
```
OsiJr, line 297: `m_metadata` is deprecated to be removed, you'll need to use the project node versions.

PR #703, opened 2020-06-04
Files are no longer 'checked' for being valid being creating a node/c…

OsiJr:
…onvert job (fixes a stall waiting for OneDrive)

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -385,26 +385,6 @@ void vcMain_MainLoop(vcState *pProgramState)
         }
 #endif //VC_HASCONVERT
 
-        // test to see if specified filepath is valid
-        udFile *pTestFile = nullptr;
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -385,26 +385,6 @@ void vcMain_MainLoop(vcState *pProgramState)
...
-        else
-        {
-          vcState::ErrorItem status;
-          status.source = vcES_File;
-          status.pData = pNextLoad; // this takes ownership so we don't need to dup or free
-          status.resultCode = result;
-
-          pNextLoad = nullptr;
-
-          pProgramState->errorItems.PushBack(status);
```
samsinsane, line 404: Without this code block, do load failures still end up in the error window?

OsiJr: Only as scene item errors

PR #708, opened 2020-06-05
Filters oriented correctly

int-Frank:
[AB#1515](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1515)
Ok so filters now orientate correctly, but there is an issue with queries not playing well with the octree. ~~This will be the next fix in this PR.~~

EDIT: This PR should be standalone. The fact that queries are broken in ECEF should be another PR.


peterAdams1014:
Does this still work when using the gizmos to rotate?

int-Frank:
@peterAdams1014 Oh I see... It does not. Let me fix this

PR #717, opened 2020-06-11
Added description field for vertical measurement tool.

asunawu:
Fixed [AB#1737](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1737).

### DIFF
--- a/src/scene/vcVerticalMeasureTool.cpp
+++ b/src/scene/vcVerticalMeasureTool.cpp
@@ -173,12 +173,15 @@ void vcVerticalMeasureTool::HandleImGui(vcState *pProgramState, size_t *pItemID)
       m_labelInfo.textSize = (vcLabelFontSize)size;
       vdkProjectNode_SetMetadataInt(m_pNode, "textSize", size);
     }
+
+    if (vcIGSW_InputText(vcString::Get("scenePOILabelDescription"), m_description, sizeof(m_description), ImGuiInputTextFlags_EnterReturnsTrue))
+      vdkProjectNode_SetMetadataString(m_pNode, "description", m_description);
+
+    ImGui::Text("%s: %.3f", vcString::Get("sceneVerticalDistance"), udAbs(m_points[0].z - m_points[2].z));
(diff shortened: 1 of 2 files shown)


src/scene/vcVerticalMeasureTool.cpp
```diff
@@ -173,12 +173,15 @@ void vcVerticalMeasureTool::HandleImGui(vcState *pProgramState, size_t *pItemID)
...
 
-  ImGui::Text("%s: %.3f", vcString::Get("sceneVerticalDistance"), udAbs(m_points[0].z - m_points[2].z));
-  ImGui::Text("%s: %.3f", vcString::Get("sceneHorizontalDistance"), udMag2(m_points[0] - m_points[2]));
-  ImGui::Text("%s: %.3f", vcString::Get("sceneStraightDistance"), udMag3(m_points[0] - m_points[2]));
+    if (vcIGSW_InputText(vcString::Get("scenePOILabelDescription"), m_description, sizeof(m_description), ImGuiInputTextFlags_EnterReturnsTrue))
+      vdkProjectNode_SetMetadataString(m_pNode, "description", m_description);
+
+    ImGui::Text("%s: %.3f", vcString::Get("sceneVerticalDistance"), udAbs(m_points[0].z - m_points[2].z));
+    ImGui::Text("%s: %.3f", vcString::Get("sceneHorizontalDistance"), udMag2(m_points[0] - m_points[2]));
+    ImGui::Text("%s: %.3f", vcString::Get("sceneStraightDistance"), udMag3(m_points[0] - m_points[2]));
```
OsiJr, line 182: This isn't actually correct. Vertical is more like `udDot(worldUp, m_points[0] - m_points[2])` @int-Frank?

int-Frank: Yes I think you are right. This makes the horizontal distance incorrect as well; it should be projected on the ground plane (ground surface if you want to include earth curvature, but I don't think we are doing that yet).

PR #718, opened 2020-06-15
Deleted vertical measurement tool node from scene explorer if action was canceled.

asunawu:
Fixed [AB#1740](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1740).

### DIFF
--- a/src/scene/vcFolder.cpp
+++ b/src/scene/vcFolder.cpp
@@ -73,7 +73,9 @@ void vcFolder::AddToScene(vcState *pProgramState, vcRenderData *pRenderData)
       if (!pSceneItem->IsValid())
       {
         pSceneItem->Cleanup(pProgramState);
-        pNode = RemoveCurrentNode(pProgramState, pNode);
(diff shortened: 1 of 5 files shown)


src/scene/vcFolder.cpp
```diff
@@ -391,3 +397,13 @@ void vcFolder::Cleanup(vcState *pProgramState)
     pNode = pNode->pNextSibling;
   }
 }
+
+vdkProjectNode *vcFolder::RemoveCurrentNode(vcState *pProgramState, vdkProjectNode *pNode)
+{
+  if (m_pNode->pFirstChild == pNode)
+    m_pNode->pFirstChild = pNode->pNextSibling;
```
samsinsane, line 404: @OsiJr can you do this? If not, should these pointers be `const`?

asunawu: Deleted.

PR #721, opened 2020-06-16
Re-organized normal packing

peterAdams1014:
Normal-Z direction is now stored (in the depth channel).

### DIFF
--- a/src/gl/directx11/shaders/atmosphereFragmentShader.hlsl
+++ b/src/gl/directx11/shaders/atmosphereFragmentShader.hlsl
@@ -1364,9 +1364,8 @@ float linearizeDepth(float depth)
 
 float3 unpackNormal(float4 normalPacked)
 {
-  return float3((2.0 * normalPacked.x) / (1 + normalPacked.x*normalPacked.x + normalPacked.y*normalPacked.y),
-                (2.0 * normalPacked.y) / (1 + normalPacked.x*normalPacked.x + normalPacked.y*normalPacked.y),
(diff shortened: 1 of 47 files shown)


src/gl/directx11/shaders/atmosphereFragmentShader.hlsl
```diff
@@ -1364,9 +1364,8 @@ float linearizeDepth(float depth)
 
 float3 unpackNormal(float4 normalPacked)
 {
-  return float3((2.0 * normalPacked.x) / (1 + normalPacked.x*normalPacked.x + normalPacked.y*normalPacked.y),
-                (2.0 * normalPacked.y) / (1 + normalPacked.x*normalPacked.x + normalPacked.y*normalPacked.y),
-		        (-1.0 + normalPacked.x*normalPacked.x + normalPacked.y*normalPacked.y) / (1.0 + (normalPacked.x*normalPacked.x) + (normalPacked.y*normalPacked.y)));
+  return float3(normalPacked.x, normalPacked.y,
+                sign(normalPacked.w) * sqrt(1 - dot(normalPacked.xy, normalPacked.xy)));
```
samsinsane, line 1368: The code might be simpler, but wouldn't it be significantly slower?

peterAdams1014: The performance difference will be negligible.

peterAdams1014:
If that will work with ShaderConductor, then yeah I'm very on board with the `common.hlsl` idea!

PR #726, opened 2020-06-18
Add labels for vertical measurement tool

asunawu:
Fixed [AB#1693](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1693).

Replaced the original label with a horizontal label and a vertical label, to show more distance information to the users. Auto reference grammar was used to simplify variable names.

### DIFF
--- a/src/scene/vcVerticalMeasureTool.cpp
+++ b/src/scene/vcVerticalMeasureTool.cpp
@@ -115,13 +115,11 @@ void vcVerticalMeasureTool::AddToScene(vcState *pProgramState, vcRenderData *pRe
     m_labelList[1].worldPosition = (m_points[1] + m_points[2]) / 2;
 
     char labelBuf[128] = {};
-    udStrcat(labelBuf, m_pNode->pName);
-    udStrcat(labelBuf, udTempStr("\n%s: %.3f", vcString::Get("sceneStraightDistance"), udMag3(m_points[0] - m_points[2])));
-    udStrcat(labelBuf, udTempStr("\n%s: %.3f", vcString::Get("sceneHorizontalDistance"), udMag2(m_points[0] - m_points[2])));
+    udSprintf(labelBuf, "%s\n%s: %.3f\n%s: %.3f", m_pNode->pName, vcString::Get("sceneStraightDistance"), udMag3(m_points[0] - m_points[2]), vcString::Get("sceneHorizontalDistance"), udMag2(m_points[0] - m_points[2]));
     m_labelList[0].pText = udStrdup(labelBuf);
 
     char labelBufVertical[128] = {};
-    udStrcat(labelBufVertical, udTempStr("%s: %.3f", vcString::Get("sceneVerticalDistance"), udAbs(m_points[0].z - m_points[2].z)));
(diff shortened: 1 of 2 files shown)


src/scene/vcVerticalMeasureTool.cpp
```diff
@@ -183,6 +198,18 @@ void vcVerticalMeasureTool::HandleImGui(vcState *pProgramState, size_t *pItemID)
...
 
+void vcVerticalMeasureTool::ResetLabelText()
+{
+  for (auto &label : m_labelList)
+  {
+    if (label.pText)
+    {
+      udFree(label.pText);
+      label.pText = nullptr;
+    }
```
samsinsane, line 209: All of this is done by `udFree`, it checks if the provided pointer is null to avoid crashes, it also sets the pointer to null before it frees the pointer. So you can collapse this down to just `udFree(label.pText);`.

src/scene/vcVerticalMeasureTool.cpp
```diff
@@ -104,14 +108,22 @@ void vcVerticalMeasureTool::AddToScene(vcState *pProgramState, vcRenderData *pRe
...
-    if(m_labelInfo.pText)
-      udFree(m_labelInfo.pText);
-    m_labelInfo.pText = udStrdup(labelBuf);
-    pRenderData->labels.PushBack(&m_labelInfo);
+    udStrcat(labelBuf, udTempStr("\n%s: %.3f", vcString::Get("sceneStraightDistance"), udMag3(m_points[0] - m_points[2])));
+    udStrcat(labelBuf, udTempStr("\n%s: %.3f", vcString::Get("sceneHorizontalDistance"), udMag2(m_points[0] - m_points[2])));
+    m_labelList[0].pText = udStrdup(labelBuf);
+
+    char labelBufVertical[128] = {};
+    udStrcat(labelBufVertical, udTempStr("%s: %.3f", vcString::Get("sceneVerticalDistance"), udAbs(m_points[0].z - m_points[2].z)));
```
samsinsane, line 122: You could do this with `udSprintf` instead of both `udStrcat` and `udTempStr`.

src/scene/vcVerticalMeasureTool.cpp
```diff
@@ -248,21 +271,25 @@ bool vcVerticalMeasureTool::HasLine()
 
 void vcVerticalMeasureTool::ClearPoints()
 {
-  for (size_t i = 0; i < POINTSIZE; i++)
+  for (size_t i = 0; i < udLengthOf(m_points); i++)
```
samsinsane, line 274: You can also do the range-for loop here, `for (auto &point : m_points)`

src/scene/vcVerticalMeasureTool.cpp
asunawu: Fixed.

src/scene/vcVerticalMeasureTool.cpp
asunawu: Fixed.

src/scene/vcVerticalMeasureTool.cpp
asunawu: This is old code, I prefer remain it.

PR #737, opened 2020-06-24
Shaders are now asynchronously loaded

peterAdams1014:
Removed API specific shader loading functions.
Moved vcPinRenderer Init() location.

### DIFF
--- a/src/rendering/vcAtmosphereRenderer.cpp
+++ b/src/rendering/vcAtmosphereRenderer.cpp
@@ -72,7 +72,7 @@ struct vcAtmosphereRenderer
 constexpr double kSunAngularRadius = 0.00935 / 2.0;
 constexpr double kLengthUnitInMeters = 1.0;
 
-udResult vcAtmosphereRenderer_Create(vcAtmosphereRenderer **ppAtmosphereRenderer)
+udResult vcAtmosphereRenderer_Create(vcAtmosphereRenderer **ppAtmosphereRenderer, udWorkerPool *pWorkerPool)
 {
   udResult result;
   vcAtmosphereRenderer *pAtmosphereRenderer = nullptr;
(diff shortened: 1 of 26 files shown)


src/rendering/vcAtmosphereRenderer.cpp
```diff
@@ -240,23 +240,26 @@ udResult vcAtmosphereRenderer_Destroy(vcAtmosphereRenderer **ppAtmosphereRendere
...
+      vcShader_GetSamplerIndex(&pAtmosphereRenderer->renderShader.uniform_scattering, pAtmosphereRenderer->renderShader.pProgram, "scattering");
+      vcShader_GetSamplerIndex(&pAtmosphereRenderer->renderShader.uniform_irradiance, pAtmosphereRenderer->renderShader.pProgram, "irradiance");
+      vcShader_GetSamplerIndex(&pAtmosphereRenderer->renderShader.uniform_sceneColour, pAtmosphereRenderer->renderShader.pProgram, "sceneColour");
+      vcShader_GetSamplerIndex(&pAtmosphereRenderer->renderShader.uniform_sceneNormal, pAtmosphereRenderer->renderShader.pProgram, "sceneNormal");
+      vcShader_GetSamplerIndex(&pAtmosphereRenderer->renderShader.uniform_sceneDepth, pAtmosphereRenderer->renderShader.pProgram, "sceneDepth");
+      vcShader_GetConstantBuffer(&pAtmosphereRenderer->renderShader.uniform_vertParams, pAtmosphereRenderer->renderShader.pProgram, "u_vertParams", sizeof(pAtmosphereRenderer->renderShader.vertParams));
+      vcShader_GetConstantBuffer(&pAtmosphereRenderer->renderShader.uniform_fragParams, pAtmosphereRenderer->renderShader.pProgram, "u_fragParams", sizeof(pAtmosphereRenderer->renderShader.fragParams));
+    }
+  ), udR_InternalError);
```
samsinsane, line 262: I didn't really notice this before, but you should avoid inlining huge lambdas like this. You could do something like:
```c++
udCallback<?> callback = [...](...) {
  ...
};
```
and then use the `callback` variable instead. Or you can move the guts of the lambda to a function, then have the lambda call that with the captured variable as a parameter.

PR #749, opened 2020-06-26
Images, labels, pins and transparent geometry can now be selected

peterAdams1014:
Picking information is now stamped into final buffer, and then asynchronously read from there.
Images are now drawn with a new shader - one which stamps picking information.
Transparent geometry gets tagged in the shader.

### DIFF
--- a/src/vcRender.cpp
+++ b/src/vcRender.cpp
@@ -37,8 +37,14 @@ enum vcObjectId
   vcObjectId_Null = 0,
   vcObjectId_Terrain = 1,
 
-  // These are offsets
-  vcObjectId_PolygonModels = 2,
+  // This is an offset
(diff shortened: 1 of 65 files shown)


src/vcRender.cpp
```diff
@@ -1475,19 +1497,20 @@ void vcRender_RenderScene(vcState *pProgramState, vcRenderContext *pRenderContex
...
+  
   vcRender_TransparentPass(pProgramState, pRenderContext, renderData);
-
+  
   vcRender_PostProcessPass(pProgramState, pRenderContext);
-
+  
   vcRender_RenderAndApplyViewSheds(pProgramState, pRenderContext, renderData);
-
+
```
OsiJr, line 1509: Lots of strange indentation

PR #755, opened 2020-06-29
Added LRU history of projects

peterAdams1014:


src/vcSettings.h
```diff
@@ -205,6 +206,13 @@ struct vcToolSettings
   } label;
 };
 
+struct HistoryProjectInfo
+{
+  const char *pName;
+  const char *pPath;
+  double date; // TODO:
```
samsinsane, line 213: Why is this a `double`? Probably best to just not have this at all until it's required.

PR #758, opened 2020-06-30
Map tiles on/off Shortcut

Jossos:
- Added Shortcut for Turning Map tiles on/off
- Fixes [AB#1765](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1765)

### DIFF
--- a/src/vcSettingsUI.cpp
+++ b/src/vcSettingsUI.cpp
@@ -965,6 +965,8 @@ void vcSettingsUI_BasicMapSettings(vcState *pProgramState)
       pProgramState->activeProject.pFolder->ChangeProjection(zone);
     }
   }
+
+  ImGui::Checkbox(vcString::Get("settingsMapsMapTiles"), &pProgramState->settings.maptiles.mapEnabled);
 }
 
 bool vcSettingsUI_VisualizationSettings(vcVisualizationSettings *pVisualizationSettings, bool isGlobal /*= false*/)


src/vcSettingsUI.cpp
```diff
@@ -965,6 +965,8 @@ void vcSettingsUI_BasicMapSettings(vcState *pProgramState)
       pProgramState->activeProject.pFolder->ChangeProjection(zone);
     }
   }
+
+  ImGui::Checkbox(vcString::Get("settingsMapsMapTiles"), &pProgramState->settings.maptiles.mapEnabled);
```
samsinsane, line 969: This function is also called by the settings UI, is there now two checkboxes to enable/disable map tiles?

Jossos: yes. someone just wanted a checkbox in the shortcut window for convenience

samsinsane: I don't think it makes sense to have two checkboxes for the same thing in the settings UI though. Do you?

samsinsane:
@OsiJr has instructed me to close this as he has other changes coming soon that will fix this.

src/vcSettingsUI.cpp
Jossos: No so onew of these boxes is in the settings window, and the other checkbox is in the little Map shortcuts window in the bottom left

samsinsane: Are you sure? This function is called by both of those systems.

Jossos: yes there's another little window called "map settings" but it isn't actually part of the settings UI. It's in the bottom right of the screen. This is where I've added the checkbox

samsinsane: image
This is what I'm talking about.

Jossos: oh what i didnt see that when i tested it

Jossos: OK i get what ur saying now. I thought they were 2 separate functions

PR #763, opened 2020-06-30
Added more options to the select UI

int-Frank:
Added more support for the select UI. Specifically:

- No item or multiple items selected: '[x] items selected'
- vcFolder displays name
- vcI3S displays name
- vcLiveFeed displays name
- vcMedia displays dropdown to change image type
- vcModel: dropdown to select attribute to view
- vcPlaceLayer: name
- vcQueryNode: Invert checkbox
- vcVerticalMeasureTool: distances
- vcViewPoint: button to move to viewpoint
- vcViewShed: Distance & colour selectors
- vcWaterNode: name

### DIFF
--- a/builds/assets/lang/enAU.json
+++ b/builds/assets/lang/enAU.json
@@ -363,6 +363,7 @@
   "menuBarWaitingForRenderLicense": "Waiting for Render license...",
   "menuBarNoRenderLicense": "No Render licenses are available for this user",
   "menuBarInactive": "Inactive",
+  "selectedItemInfoPanelitemsSelected": "items selected",
   "popupMenuCopy": "Copy",
   "popupMenuPaste": "Paste",
   "settingsConnection": "Connection",
--- a/src/scene/vcMedia.cpp
+++ b/src/scene/vcMedia.cpp
@@ -310,6 +310,14 @@ void vcMedia::HandleSceneExplorerUI(vcState *pProgramState, size_t *pItemID)
   }
 }
 
+void vcMedia::HandleSceneEmbeddedUI(vcState * /*pProgramState*/)
+{
+  const char *imageTypeNames[] = {vcString::Get("scenePOILabelImageTypeStandard"), vcString::Get("scenePOILabelImageTypeOriented"), vcString::Get("scenePOILabelImageTypePanorama"), vcString::Get("scenePOILabelImageTypePhotosphere"), vcString::Get("scenePOILabelImageTypeScreen")};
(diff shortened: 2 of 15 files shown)


builds/assets/lang/enAU.json
```diff
@@ -363,6 +363,7 @@
   "menuBarWaitingForRenderLicense": "Waiting for Render license...",
   "menuBarNoRenderLicense": "No Render licenses are available for this user",
   "menuBarInactive": "Inactive",
+  "selectedItemInfoPanelitemsSelected": "items selected",
```
OsiJr, line 366: This probably should use the `{0}` style stuff in case order needs to change

src/scene/vcMedia.cpp
```diff
@@ -310,6 +310,14 @@ void vcMedia::HandleSceneExplorerUI(vcState *pProgramState, size_t *pItemID)
   }
 }
 
+void vcMedia::HandleSceneEmbeddedUI(vcState * /*pProgramState*/)
+{
+  const char *imageTypeNames[] = {vcString::Get("scenePOILabelImageTypeStandard"), vcString::Get("scenePOILabelImageTypeOriented"), vcString::Get("scenePOILabelImageTypePanorama"), vcString::Get("scenePOILabelImageTypePhotosphere"), vcString::Get("scenePOILabelImageTypeScreen")};
```
samsinsane, line 315: We normally break these over multiple lines - otherwise, there should be a space after `{` and before `}`.

src/scene/vcMedia.cpp
```diff
@@ -310,6 +310,14 @@ void vcMedia::HandleSceneExplorerUI(vcState *pProgramState, size_t *pItemID)
...
+  UDCOMPILEASSERT(udLengthOf(imageTypeNames) == vcIT_Count, "Update image names");
+  if (ImGui::Combo(udTempStr("%s##ImageEmbeddedUI", vcString::Get("scenePOILabelImageType")), (int *)&m_image.type, imageTypeNames, (int)udLengthOf(imageTypeNames)))
```
samsinsane, line 317: What is `m_image.type`? You're blindly casting between pointer types, this is dangerous. At the very least you should add a compile assert that these two types are the same length - if they're not things are going to break. Otherwise, create a local `int` assign to it, use the address of that for the combo, and then assign it back into `m_image.type`, this avoids all type issues:
```c++
int temp = (int)m_image.type; // Correctly casts to int
if (ImGui::Combo(..., &temp, ...))
  ...
m_image.type = (the type...)temp; // Correctly casts back to the right type
```

PR #775, opened 2020-07-01
Fixed a memory leak with tile server preview images

peterAdams1014:


src/vcSettingsUI.cpp
```diff
@@ -49,6 +49,13 @@ enum vcLASClassifications
   vcLASClassifications_LastClassification = 255 // Because they are always stored in a uint8
 };
 
+void vcSettingsUI_Cleanup(vcState *pProgramState)
+{
+  for (size_t i = 0; i < udLengthOf(s_mapTiles); i++)
```
samsinsane, line 54: For what it's worth you could do a range-for loop here. `for (auto &mapTile : s_mapTiles)` or with the correct type.

PR #782, opened 2020-07-01
Added system of measurement option: Metric or Imperial

int-Frank:
Fixes [AB#220](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/220)

- Users can now select to use imperial or metric measurement system and have this display in the Embedded UI
- Measurements can be promoted or demoted, for example a length of `1000m` will be promoted to '1km'
- Framework to set all significant figures for each unit type, separately
- Framework to set all measurement unit types separately


src/vcSettingsUI.cpp
```diff
@@ -368,6 +369,22 @@ void vcSettingsUI_Show(vcState *pProgramState)
...
+            vcUnitConversion_SetMetric(&pProgramState->settings.unitConversionData);
+            udStrcpy(unitTextBuffer, vcString::Get("settingsUnitsOfMeasurementUsingMetric")); //Temp feedback for the user until we implement per unit options
+          }
+          if (ImGui::Button(vcString::Get("settingsSetMeasurementSystemImperial")))
+          {
+            vcUnitConversion_SetImperial(&pProgramState->settings.unitConversionData);
+            udStrcpy(unitTextBuffer, vcString::Get("settingsUnitsOfMeasurementUsingImperial")); //Temp feedback for the user until we implement per unit options
+          }
+          ImGui::Text("%s", unitTextBuffer);
+        }
```
OsiJr, line 386: I know this is for future expansion but right now with just these 2 options, is should be a combo box on the "Appearance" tab

src/vcUnitConversion.cpp
```diff
@@ -286,3 +300,293 @@ int vcUnitConversion_ConvertTemperatureToString(char *pBuffer, size_t bufferSize
...
+  {
+    finalUnit = (vcDistanceUnit)pData->distanceUnit[i].unit;
+    finalValue = vcUnitConversion_ConvertDistance(value, unit, finalUnit);
+
+    if (udAbs(finalValue) < pData->distanceUnit[i].upperLimit)
+      break;
+  }
+
+  sigFigs = pData->distanceSigFigs;
+  if (sigFigs > 10) sigFigs = 10;
```
OsiJr, line 405: Either newline or just do in in one line using clamp

src/vcUnitConversion.cpp
```diff
@@ -286,3 +300,293 @@ int vcUnitConversion_ConvertTemperatureToString(char *pBuffer, size_t bufferSize
...
+
+  convertResult = vcUnitConversion_ConvertTimeToString(pBuffer, bufferSize, finalValue, pData->timeReference, SigFigsFormats[sigFigs]);
+
+  result = convertResult < 0 ? udR_WriteFailure : udR_Success;
+
+epilogue:
+  return result;
+}
+
+//These should be translatable
```
OsiJr, line 539: Lots of dead code here?

PR #785, opened 2020-07-01
Media image nodes now don't 'flicker' on update.

peterAdams1014:


src/vcRender.cpp
```diff
@@ -1398,7 +1398,7 @@ bool vcRender_DrawSelectedGeometry(vcState *pProgramState, vcRenderContext *pRen
     }
 
   }
-
+
```
samsinsane, line 1401: This should probably be reverted?

PR #809, opened 2020-07-06
Set the srid to 0 for non geolocated startup

int-Frank:
Fixes [AB#1839](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1839)


OsiJr:
I don't think this is correct- @peterAdams1014  to confirm?

PR #849, opened 2020-07-09
show hints if the creation of new project fails.

asunawu:
The user experience of [AB#1877](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1877) has been improved, but it still needs to be fixed in the VDK. And fixed the empty string comparison when creating a new project.

### DIFF
--- a/src/vcModals.cpp
+++ b/src/vcModals.cpp
@@ -434,6 +434,15 @@ void vcModals_DrawNewProject(vcState *pProgramState)
       //TODO: Additional export settings
     }
 
+    static vdkError result = vE_Success;
+    if (result != vE_Success)
+    {
+      ImGui::NextColumn();
+      ImGui::Spacing();
+      ImGui::Spacing();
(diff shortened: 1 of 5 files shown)


src/vcModals.cpp
```diff
@@ -443,40 +452,39 @@ void vcModals_DrawNewProject(vcState *pProgramState)
...
+          result = vcProject_CreateFileScene(pProgramState, pProjectPath, pProgramState->modelPath, zoneCustomSRID);
         }
         else // server
         {
-          createSucceeded = vcProject_CreateServerScene(pProgramState, pProgramState->modelPath, udUUID_GetAsString(selectedGroup), zoneCustomSRID);
+          result = vcProject_CreateServerScene(pProgramState, pProgramState->modelPath, udUUID_GetAsString(selectedGroup), zoneCustomSRID);
         }
 
-        if (createSucceeded)
+        if (vE_Success == result)
```
OsiJr, line 481:
```suggestion
        if (result == vE_Success )
```

PR #858, opened 2020-07-09
Convert button on getting started

OsiJr:
Fixes [AB#1913](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1913)


src/vcModals.cpp
```diff
@@ -376,34 +368,45 @@ void vcModals_DrawWelcome(vcState *pProgramState)
...
+            udStrcpy(pProgramState->modelPath, vcString::Get("modalProjectNewTitle"));
 
-          if (i == 0) // Geolocated
-            zoneCustomSRID = 84;
-          else if (i == 1)
-            zoneCustomSRID = 0; // Non Geolocated
+            if (items[i].icon == vcMBBI_Geospatial)
+              zoneCustomSRID = 84; // Geolocated
+            else if (items[i].icon == vcMBBI_Grid)
+              zoneCustomSRID = 0; // Non Geolocated
```
int-Frank, line 384: Is `0` standard for non-geolocated? I wonder if we should use an enum for common srid values.

PR #861, opened 2020-07-10
Automatically complete the saving name of new project and also check …

asunawu:
Fixed [AB#1916](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1916).

 Automatically complete the new project name and also check whether the file exists (will return an error hint).


src/vcProject.cpp
```diff
@@ -16,8 +16,8 @@ const char *vcProject_ErrorToString(vdkError error)
     return vcString::Get("errorInvalidParameter");
   case vE_OpenFailure:
     return vcString::Get("errorOpenFailure");
-  case vE_NotSupported:
-    return vcString::Get("errorUnsupported");
+  case vE_WriteFailure:
+    return vcString::Get("errorFileExists");
```
samsinsane, line 20: You probably shouldn't remove the `vE_NotSupported` message?

src/vcModals.cpp
```diff
@@ -548,9 +548,9 @@ void vcModals_DrawWelcome(vcState *pProgramState)
       {
         if (localOrServerProject == 0) // local
         {
-          // TODO: CONFIRM FILE OVERRIDE IF EXISTS
-
-          result = vcProject_CreateFileScene(pProgramState, pProjectPath, pProgramState->modelPath, zoneCustomSRID);
+          udFilename exportFilename(pProjectPath);
+          vcProject_AutoCompletedName(&exportFilename, pProjectPath, pProgramState->modelPath);
+          result = vcProject_CreateFileScene(pProgramState, exportFilename, pProgramState->modelPath, zoneCustomSRID);
```
samsinsane, line 553: I think the intent here was to call `vcModals_OverwriteExistingFile` instead of failing, so the `errorFileExists` error message isn't required now.

asunawu: I saw in vcProject_CreateFileScene that it called the function vdkProject_CreateInFile, which will create an empty, only local vdkProject instance and return false if file exists. I could change the current behaviour of vcProject_CreateFileScene if you sure. And nowhere in vcProject_CreateFileScene will return vE_NotSupported now, however I can add it back too.

src/vcProject.cpp
asunawu: It is not used now.

src/vcModals.cpp
asunawu: Overwriting existing files is currently not supported,

src/vcProject.cpp
```diff
@@ -16,8 +16,8 @@ const char *vcProject_ErrorToString(vdkError error)
     return vcString::Get("errorInvalidParameter");
   case vE_OpenFailure:
     return vcString::Get("errorOpenFailure");
-  case vE_NotSupported:
-    return vcString::Get("errorUnsupported");
+  case vE_WriteFailure:
```
OsiJr, line 19:
```suggestion
  case vE_NotSupported:	  case vE_WriteFailure:
    return vcString::Get("errorUnsupported");
  case vE_WriteFailure:
```

PR #866, opened 2020-07-10
Added UI to import server projects.

peterAdams1014:


src/vcModals.cpp
```diff
@@ -729,37 +729,121 @@ void vcModals_DrawImportProject(vcState *pProgramState)
...
-      pProgramState->modelPath[0] = '\0';
-      ImGui::CloseCurrentPopup();
+      selectedGroupIndex = 0;
+      selectedProjectIndex = 0;
     }
 
-    ImGui::SameLine();
+    struct
+    {
+      const char *pName;          vcMenuBarButtonIcon icon;
```
samsinsane, line 751: Not a huge fan of this, but we can fix this later.

PR #912, opened 2020-07-17
Awu/bpa

asunawu:
Improved BPA algorithm.

[AB#1768](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1768)


OsiJr (changes requested):
I started reviewing it but finally got results back. This changeset results in much less accurate calculations. The original implementation has issues where there are gaps but this version is off by 0.2-0.5m most of the time.

PR #913, opened 2020-07-20
Area Measurement Area Calculation

Jossos:
- Calculate Area Measurement area based on polymodel generated
- Fixes [AB#1973](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1973)

### DIFF
--- a/src/scene/vcPOI.cpp
+++ b/src/scene/vcPOI.cpp
@@ -1332,17 +1332,17 @@ void vcPOI::GenerateLineFillPolygon()
       int index2 = pIndices[i + 1];
       int index3 = pIndices[i + 2];
 
-      udFloat3 e1 = {
-        pVerts[index2].position.x - pVerts[index1].position.x,
-        pVerts[index2].position.y - pVerts[index1].position.y,
-        pVerts[index2].position.z - pVerts[index1].position.z };
-      udFloat3 e2 = {
-        pVerts[index3].position.x - pVerts[index1].position.x,
-        pVerts[index3].position.y - pVerts[index1].position.y,
-        pVerts[index3].position.z - pVerts[index1].position.z };
-      udFloat3 e3 = udCross3<float>(e1, e2);
+      udDouble3 e1 = {
+        (double)(pVerts[index2].position.x - pVerts[index1].position.x),
+        (double)(pVerts[index2].position.y - pVerts[index1].position.y),
+        (double)(pVerts[index2].position.z - pVerts[index1].position.z) };
+      udDouble3 e2 = {
+        (double)(pVerts[index3].position.x - pVerts[index1].position.x),
+        (double)(pVerts[index3].position.y - pVerts[index1].position.y),
+        (double)(pVerts[index3].position.z - pVerts[index1].position.z) };
+      udDouble3 e3 = udCross3<double>(e1, e2);
 
-      m_meshArea += (double)(udMag3(e3) / 2.0f);
+      m_meshArea += udMag3(e3) / 2.0;
     }
 
     udFree(pVerts);
--- a/src/scene/vcPOI.cpp
+++ b/src/scene/vcPOI.cpp
@@ -318,6 +318,7 @@ public:
     m_pParent->m_showAllLengths = false;
     m_pParent->m_showLength = false;
     m_pParent->m_showFill = true;
+    m_pParent->m_meshArea = 0.0;
   }
 
   ~vcPOIState_MeasureArea()
@@ -1090,7 +1091,10 @@ void vcPOI::CalculateArea(const udDouble4 &projectionPlane)
   if (m_line.numPoints < 3)
     return;
 
-  m_area = udProjectedArea(projectionPlane, m_line.pPoints, (size_t)m_line.numPoints);
+  if (m_pPolyModel)
+    m_area = m_meshArea;
+  else
+    m_area = udProjectedArea(projectionPlane, m_line.pPoints, (size_t)m_line.numPoints);
 }
 
 void vcPOI::CalculateTotalLength()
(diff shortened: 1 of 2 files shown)


int-Frank:
> * Calculate Area Measurement area based on polymodel generated
> * Fixes AB#1973

Does this format work, or does 'Fixes AB#1973' need to be the first line of text in the description?

src/scene/vcPOI.cpp
```diff
@@ -1337,6 +1341,23 @@ void vcPOI::GenerateLineFillPolygon(vcState *pProgramState)
     vcPolygonModel_CreateFromRawVertexData(&m_pPolyModel, pVerts, numPoints, vcP3N3UV2VertexLayout, (int)(udLengthOf(vcP3N3UV2VertexLayout)));
 
     m_pPolyModel->modelOffset = udDouble4x4::translation(m_line.pPoints[0]);
+    
+    // Calculate area
+    m_meshArea = 0.0;
+    for (int i = 0; i < trianglePointList.size(); i += 3)
```
int-Frank, line 1347: signed / unsigned comparison mismatch. This will cause some builds to fail.

src/scene/vcPOI.cpp
```diff
@@ -319,6 +319,7 @@ class vcPOIState_MeasureArea : public vcPOIState_General
     m_pParent->m_showAllLengths = false;
     m_pParent->m_showLength = false;
     m_pParent->m_showFill = true;
+    m_pParent->m_meshArea = 0.0f;
```
int-Frank, line 322: This should be `0.0` (no f, it is a double)

src/scene/vcPOI.cpp
```diff
@@ -1337,6 +1341,23 @@ void vcPOI::GenerateLineFillPolygon(vcState *pProgramState)
     vcPolygonModel_CreateFromRawVertexData(&m_pPolyModel, pVerts, numPoints, vcP3N3UV2VertexLayout, (int)(udLengthOf(vcP3N3UV2VertexLayout)));
 
     m_pPolyModel->modelOffset = udDouble4x4::translation(m_line.pPoints[0]);
+    
+    // Calculate area
+    m_meshArea = 0.0;
+    for (int i = 0; i < trianglePointList.size(); i += 3)
+    {
+      udFloat3 e1 = {
+        pVerts[i + 1].position.x - pVerts[i].position.x,
```
int-Frank, line 1350: You should use doubles to calculate area.

src/scene/vcPOI.cpp
```diff
@@ -1337,6 +1341,23 @@ void vcPOI::GenerateLineFillPolygon(vcState *pProgramState)
...
+        pVerts[i + 1].position.y - pVerts[i].position.y,
+        pVerts[i + 1].position.z - pVerts[i].position.z };
+      udFloat3 e2 = {
+        pVerts[i + 2].position.x - pVerts[i].position.x,
+        pVerts[i + 2].position.y - pVerts[i].position.y,
+        pVerts[i + 2].position.z - pVerts[i].position.z };
+      udFloat3 e3 = udCross3<float>(e1, e2);
+
+      m_meshArea += (double)(udMag3(e3) / 2.0f);
```
int-Frank, line 1359: Again, you should do all calculations in doubles. Then convert to floats if the renderer needs floats.

src/scene/vcPOI.cpp
```diff
@@ -1344,19 +1344,14 @@ void vcPOI::GenerateLineFillPolygon(vcState *pProgramState)
...
-        pVerts[i + 1].position.y - pVerts[i].position.y,
-        pVerts[i + 1].position.z - pVerts[i].position.z };
-      udFloat3 e2 = {
-        pVerts[i + 2].position.x - pVerts[i].position.x,
-        pVerts[i + 2].position.y - pVerts[i].position.y,
-        pVerts[i + 2].position.z - pVerts[i].position.z };
-      udFloat3 e3 = udCross3<float>(e1, e2);
-
-      m_meshArea += (double)(udMag3(e3) / 2.0f);
+      udDouble3 triPoint1 = udDouble3::create(pVerts[i].position);
```
int-Frank, line 1349: I guess I mean to say; it looks like `pVerts` is created from `trianglePointList`. `pVerts` looks to be floats, so we lose precision when we convert from double to float. Can we calculate the area using `trianglePointList` instead? Then we can keep double precision all the way through.

Jossos: Ok i've fixed it so that it does all of the math in doubles until the end where it converts it to float for the final render

src/scene/vcPOI.cpp
```diff
@@ -1304,19 +1304,16 @@ void vcPOI::GenerateLineFillPolygon(vcState *pProgramState)
        
     int numPoints = (int)trianglePointList.size();
 
-    vcP3N3UV2Vertex *pVerts = udAllocType(vcP3N3UV2Vertex, numPoints, udAF_Zero);
+    udDouble3 *pPositions = udAllocType(udDouble3, numPoints, udAF_Zero);
```
samsinsane, line 1307: When is this freed?

Jossos: never. good catch.

Jossos: fixed

src/scene/vcPOI.cpp
```diff
@@ -1320,6 +1324,27 @@ void vcPOI::GenerateLineFillPolygon()
...
+    for (int i = 0; i < numIndices; i += 3)
+    {
+      int index1 = pIndices[i];
+      int index2 = pIndices[i + 1];
+      int index3 = pIndices[i + 2];
+
+      udDouble3 e1 = {
+        (double)(pVerts[index2].position.x - pVerts[index1].position.x),
+        (double)(pVerts[index2].position.y - pVerts[index1].position.y),
+        (double)(pVerts[index2].position.z - pVerts[index1].position.z) };
```
samsinsane, line 1338: This and the line below should like this:
```suggestion
      udDouble3 e1 = udDouble3::create(pVerts[index2].position - pVerts[index1].position);
```

samsinsane (changes requested):

samsinsane (approved):

PR #916, opened 2020-07-21
Area Measurement Point Placement Crash Fix

Jossos:
- Fixed crash occurring when placing points of an area measurement
- Crash is still possible when crossing lines, otherwise it should be impossible.

### DIFF
--- a/src/scene/vcPOI.cpp
+++ b/src/scene/vcPOI.cpp
@@ -1276,19 +1276,22 @@ void vcPOI::GenerateLineFillPolygon(vcState *pProgramState)
 {
   if (m_line.numPoints >= 3)
   {
+    if (m_line.pPoints[m_line.numPoints - 1] == m_line.pPoints[m_line.numPoints - 2])
+      return;
+
     udFloat3 defaultNormal = udFloat3::create(0.0f, 0.0f, 1.0f);
     udFloat2 defaultUV = udFloat2::create(0.0f, 0.0f);
     udDouble2 min = {};
     udDouble2 max = {};
     std::vector<udDouble2> trianglePointList;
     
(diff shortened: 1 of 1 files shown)


src/scene/vcPOI.cpp
```diff
@@ -1276,19 +1276,22 @@ void vcPOI::GenerateLineFillPolygon(vcState *pProgramState)
 {
   if (m_line.numPoints >= 3)
   {
+    if (m_line.pPoints[m_line.numPoints - 1] == m_line.pPoints[m_line.numPoints - 2])
```
samsinsane, line 1279: Why are the last points the same? I assume this crashes in the triangle generation, it might be better to explicitly check for degenerate triangles instead?

int-Frank: > * Crash is still possible when crossing lines, otherwise it should be impossible.

So, are these crashes the result of 2 different problems? Or do they stem from the same issue?

Jossos: So the last 2 points will be the same while you're creating the area measurement. you click, it places a point while also having the next point ready on the mousecursor, placing both points at the exact same spot.

There is another crash due to generating the poly model, but with this fix, it should never crash unless you cross lines.

samsinsane: I was able to reproduce this crash by:
- Not moving the mouse and clicking again.
- Placing two points, and moving the mouse back to the first and clicking again.

Whatever is actually crashing needs to be fixed, don't just workaround it.

Jossos: ok i know what the issue is. Ill fix it in another PR

PR #917, opened 2020-07-21
Changed start/end selecting to slider bar and fixed label position.

asunawu:
Fixed [AB#1757](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1757).

Changed start/end selecting to slider bar and fixed label position.

### DIFF
--- a/src/scene/vcVerticalMeasureTool.cpp
+++ b/src/scene/vcVerticalMeasureTool.cpp
@@ -20,8 +20,7 @@
 vcVerticalMeasureTool::vcVerticalMeasureTool(vcProject *pProject, vdkProjectNode *pNode, vcState *pProgramState) :
   vcSceneItem(pProject, pNode, pProgramState)
   , m_done(false)
-  , m_pickStart(true)
-  , m_pickEnd(true)
+  , m_selectedPoint(-1)
   , m_markDelete(false)
   , m_showAllDistances(false)
   , m_pLineInstance(nullptr)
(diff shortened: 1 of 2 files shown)


src/scene/vcVerticalMeasureTool.cpp
```diff
@@ -64,8 +63,7 @@ void vcVerticalMeasureTool::EndMeasure(vcState *pProgramState, const udDouble3 &
 
 void vcVerticalMeasureTool::OnNodeUpdate(vcState *pProgramState)
 {
-  vdkProjectNode_GetMetadataBool(m_pNode, "pickStart", &m_pickStart, true);
-  vdkProjectNode_GetMetadataBool(m_pNode, "pickEnd", &m_pickEnd, true);
+  vdkProjectNode_GetMetadataInt(m_pNode, "selectedPoint", &m_selectedPoint, -1);
```
samsinsane, line 66: @OsiJr Should we be handling the old metadata going forward?

OsiJr: Yes but why are these attributes even being saved to the project?

asunawu: Sorry I thought we needed to save all attributes, will remove it.

asunawu: Removed.

PR #918, opened 2020-07-21
Hide map tiles when off

Jossos:
- Disabled Map tile selection when map tiles are turned off
- Fixes [AB#1936](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1936)

### DIFF
--- a/src/vcSettingsUI.cpp
+++ b/src/vcSettingsUI.cpp
@@ -727,113 +727,120 @@ void vcSettingsUI_BasicMapSettings(vcState *pProgramState, bool alwaysShowOption
       }
     }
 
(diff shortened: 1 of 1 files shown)


Jossos:
@samsinsane I've purposefully had the window not shrink from the removal of the tile icons. positioning is based on the window size

samsinsane:
I can't imagine this being an acceptable solution to whoever reported this issue in the first place. It looks quite awful.

Jossos:
what should it do then?

samsinsane:
Resize and reposition?

samsinsane (approved):

PR #920, opened 2020-07-27
Fixed an issue with getting the rotations for non projections

int-Frank:
`vcGIS_GetQuaternion()` now returns the unit quaternion for all projections but ECEF. Indeed this function was specifically written for just ECEF.

### DIFF
--- a/src/vcGIS.cpp
+++ b/src/vcGIS.cpp
@@ -98,6 +98,10 @@ void vcGIS_GetOrthonormalBasis(const udGeoZone &zone, udDouble3 localPosition, u
 
 udDoubleQuat vcGIS_GetQuaternion(const udGeoZone &zone, udDouble3 localPosition)
 {
+  //Flat projections just return the unit quaternion.
+  if (zone.srid != 4978)
+    return udDoubleQuat{0.0, 0.0, 0.0, 1.0};
+
   udDouble3 up, north, east;
   vcGIS_GetOrthonormalBasis(zone, localPosition, &up, &north, &east);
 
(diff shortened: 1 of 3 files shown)


src/vcGIS.cpp
```diff
@@ -98,6 +98,10 @@ void vcGIS_GetOrthonormalBasis(const udGeoZone &zone, udDouble3 localPosition, u
 
 udDoubleQuat vcGIS_GetQuaternion(const udGeoZone &zone, udDouble3 localPosition)
 {
+  //Flat projections just return the unit quaternion.
+  if (zone.srid != 4978)
```
samsinsane, line 102: Your comment suggests this is an ECEF feature, so shouldn't this be `zone.projection != udGZPT_ECEF`? Otherwise, it's only working for a very specific ECEF SRID.

samsinsane (changes requested):
@OsiJr should probably approve of this too.

src/vcGIS.cpp
int-Frank: You are correct. I've added support for other projections.

src/vcGIS.cpp
```diff
@@ -98,15 +98,25 @@ void vcGIS_GetOrthonormalBasis(const udGeoZone &zone, udDouble3 localPosition, u
...
 
-  udDouble4x4 m = udDouble4x4::create(-north.x, -north.y, -north.z, 0,
-                                       east.x,   east.y,   east.z,  0,
-                                       up.x,     up.y,     up.z,    0,
-                                       0,        0,        0,       1);
+  if (zone.projection == udGZPT_LatLong)
+  {
+    q = udDoubleQuat::create(-UD_HALF_PI, 0.0, 0.0);
+  }
+  if (zone.projection == udGZPT_ECEF)
```
samsinsane, line 107: `else if`, only one can be true.

Perhaps a switch statement should be used, this way a new projection being added triggers a warning here (requires that the `default` label isn't specified obviously). This warning will be treated as an error, and we'll be forced to handle new projections as they're supported instead of having random bugs appear.

samsinsane (changes requested):

PR #921, opened 2020-07-29
Moved add filter from right click menu to scene tool panel

asunawu:
Add more information in SceneEmbeddedUI, including the position and extents of the filter, the mouse endpoint (the point where the mouse ray intersects the ground), and the width/length warning message if the user drags filter to a negative size.


src/scene/vcQueryNode.h
```diff
@@ -47,4 +48,21 @@ class vcQueryNode : public vcSceneItem
   vdkQueryFilter *m_pFilter;
 };
 
+struct vdkProjectNode;
+struct vcQueryNodeFilterInput
+{
+  vcQueryNodeFilterShape shape = vcQNFS_None;
```
samsinsane, line 54: You should try to avoid this, it requires the constructor is called to set this. This will implicitly happen in most cases, but a `udChunkedArray` of this struct won't call the constructor.

src/vcProject.cpp
```diff
@@ -709,10 +710,13 @@ void vcProject_ClearSelection(vcState *pProgramState, bool clearToolState /*= tr
 {
   vcProject_ClearSelection(pProgramState->activeProject.pRoot);
   pProgramState->sceneExplorer.selectedItems.clear();
-  pProgramState->sceneExplorer.clickedItem = {};
+  pProgramState->sceneExplorer.clickedItem = {};
```
samsinsane, line 713: White space at the end of the line.

src/vcSettings.cpp
```diff
@@ -475,6 +475,9 @@ bool vcSettings_Load(vcSettings *pSettings, bool forceReset /*= false*/, vcSetti
     vcHotkey::Set(vcB_ToggleAnnotateTool, data.Get("keys.%s", vcHotkey::GetBindName(vcB_ToggleAnnotateTool)).AsInt(67));
     vcHotkey::Set(vcB_OpenSettingsMenu, data.Get("keys.%s", vcHotkey::GetBindName(vcB_OpenSettingsMenu)).AsInt(vcMOD_Ctrl + SDL_SCANCODE_L));
     vcHotkey::Set(vcB_ToggleMeasureHeightTool, data.Get("keys.%s", vcHotkey::GetBindName(vcB_ToggleMeasureHeightTool)).AsInt(11));
+    vcHotkey::Set(vcB_AddBoxFilter, data.Get("keys.%s", vcHotkey::GetBindName(vcB_AddBoxFilter)).AsInt(12));
+    vcHotkey::Set(vcB_AddSphereFilter, data.Get("keys.%s", vcHotkey::GetBindName(vcB_AddSphereFilter)).AsInt(13));
+    vcHotkey::Set(vcB_AddCylinderFilter, data.Get("keys.%s", vcHotkey::GetBindName(vcB_AddCylinderFilter)).AsInt(14));
```
samsinsane, line 480: What keys are these?

asunawu: I J K

src/scene/vcQueryNode.h
asunawu: Deleted.

src/vcProject.cpp
asunawu: Deleted.

PR #923, opened 2020-07-29
Added layers to maps system.

peterAdams1014:


src/vcQuadTree.cpp
```diff
@@ -543,7 +548,13 @@ bool vcQuadTree_IsBlockUsed(vcQuadTree *pQuadTree, uint32_t blockIndex)
...
 bool vcQuadTree_RecursiveHasDrawData(vcQuadTree *pQuadTree, vcQuadTreeNode *pNode)
 {
-  if (pNode->colourInfo.data.pTexture || pNode->demInfo.data.pTexture)
+  for (int i = 0; i < pQuadTree->pSettings->maptiles.activeLayerCount; ++i)
+  {
+    if (pNode->pPayloads[i].data.pTexture != nullptr)
+      return true;
+  }
+
+  if (pNode->demInfo.data.pTexture != nullptr)
```
samsinsane, line 557: This should probably be first, it's the cheapest check.

src/vcQuadTree.cpp
```diff
@@ -563,7 +574,14 @@ bool vcQuadTree_ShouldFreeBlock(vcQuadTree *pQuadTree, uint32_t blockIndex)
...
     vcQuadTreeNode *pChildNode = &pQuadTree->nodes.pPool[blockIndex + c];
-    if (pChildNode->colourInfo.loadStatus.Get() == vcNodeRenderInfo::vcTLS_Downloading || pChildNode->demInfo.loadStatus.Get() == vcNodeRenderInfo::vcTLS_Downloading)
+    for (int i = 0; i < pQuadTree->pSettings->maptiles.activeLayerCount; ++i)
+    {
+      if (pChildNode->pPayloads[i].loadStatus.Get() == vcNodeRenderInfo::vcTLS_Downloading)
+        return false;
+
+    }
+
+    if (pChildNode->demInfo.loadStatus.Get() == vcNodeRenderInfo::vcTLS_Downloading)
```
samsinsane, line 584: Same as above.

src/vcSettings.cpp
```diff
@@ -250,14 +250,48 @@ bool vcSettings_Load(vcSettings *pSettings, bool forceReset /*= false*/, vcSetti
     pSettings->maptiles.mapEnabled = data.Get("maptiles.enabled").AsBool(true);
     pSettings->maptiles.demEnabled = data.Get("maptiles.demEnabled").AsBool(true);
 
-    udStrcpy(pSettings->maptiles.mapType, data.Get("maptiles.type").AsString("euc-az-aerial"));
+    pSettings->maptiles.activeLayerCount = data.Get("maptiles.layerCount").AsInt(0);
```
samsinsane, line 253: `maptiles.layerCount` isn't really necessary, you can do this with `data.Get("maptiles.layer").ArrayLength()`?

src/vcSettings.cpp
```diff
@@ -250,14 +250,48 @@ bool vcSettings_Load(vcSettings *pSettings, bool forceReset /*= false*/, vcSetti
...
+      pSettings->maptiles.activeLayerCount = 1;
+      pSettings->maptiles.layers[0].enabled = true;
+      for (int mapLayer = 1; mapLayer < vcMaxTileLayerCount; ++mapLayer)
+      {
+        pSettings->maptiles.layers[mapLayer].mapHeight = 0.0f;
+        pSettings->maptiles.layers[mapLayer].blendMode = vcMTBM_Hybrid;
+        pSettings->maptiles.layers[mapLayer].transparency = 1.0f;
+      }
+
+      udStrcpy(pSettings->maptiles.layers[0].mapType, data.Get("maptiles.layer[0].type").AsString("euc-az-aerial"));
```
samsinsane, line 268: This is the backwards compatibility path, these `data.Get()` calls shouldn't contain `.layer[0]`.

src/vcSettings.h
```diff
@@ -366,14 +367,19 @@ struct vcSettings
...
+      char mapType[32]; // 'custom', 'euc-osm-base', 'euc-az-aerial', 'euc-az-roads'
+      vcMapServer customServer;
+      vcMapServer activeServer; // The server settings actually in use
 
-    vcMapTileBlendMode blendMode;
-    float transparency;
+      vcMapTileBlendMode blendMode;
+      float transparency;
+    } layers[vcMaxTileLayerCount];
+    int activeLayerCount;
```
samsinsane, line 382: What does this provide over checking if the layer is `enabled`?

src/vcSettingsUI.cpp
```diff
@@ -768,63 +724,107 @@ void vcSettingsUI_BasicMapSettings(vcState *pProgramState, bool alwaysShowOption
...
+    for (int mapLayer = 0; mapLayer < vcMaxTileLayerCount; ++mapLayer)
+    {
+      if (ImGui::Button(udTempStr(pProgramState->settings.maptiles.layers[mapLayer].enabled ? "-##mapLayer%d" : "+##mapLayer%d", mapLayer), ImVec2(24, 24)))
+        pProgramState->settings.maptiles.layers[mapLayer].enabled = !pProgramState->settings.maptiles.layers[mapLayer].enabled;
 
-    ImGuiStyle &style = ImGui::GetStyle();
-    ImVec2 button_sz(128, 128);
-    float window_visible_x2 = ImGui::GetWindowPos().x + ImGui::GetWindowContentRegionMax().x;
+      ImGui::SameLine();
+      ImGui::Text("%s %d", vcString::Get("settingsMapLayer"), mapLayer);
```
samsinsane, line 736: The number should be part of the string, I believe we've done this for other counts (search for `{0}` in the JSON file), also this should be `mapLayer + 1` otherwise it's `Layer 0` instead of `Layer 1`.

src/vcSettings.h
peterAdams1014: `activeLayerCount`? Just simplifies some logic in parts.

src/main.cpp
```diff
@@ -1492,8 +1492,8 @@ void vcRenderSceneUI(vcState *pProgramState, const ImVec2 &windowPos, const ImVe
 
     if (pProgramState->settings.maptiles.mapEnabled && pProgramState->geozone.projection != udGZPT_Unknown)
     {
-      if (pProgramState->settings.maptiles.activeServer.attribution[0] != '\0')
-        udSprintf(&pBuffer, "%s, %s", pBuffer, pProgramState->settings.maptiles.activeServer.attribution);
+      if (pProgramState->settings.maptiles.layers[0].activeServer.attribution[0] != '\0')
+        udSprintf(&pBuffer, "%s, %s", pBuffer, pProgramState->settings.maptiles.layers[0].activeServer.attribution);
```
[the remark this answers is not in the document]
peterAdams1014: Yes. Nice catch.

src/vcSettings.cpp
peterAdams1014: Good catch. Yes this was very broken.

src/vcSettingsUI.cpp
peterAdams1014: Done and done.

PR #936, opened 2020-08-12
Project Info Images

Jossos:
- Implemented Image loading for Project Info image tags (Something here)
- Fixes [AB#1983](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1983)

### DIFF
--- a/src/vcModals.cpp
+++ b/src/vcModals.cpp
@@ -619,12 +619,11 @@ void vcModals_DrawExportProject(vcState *pProgramState)
   }
 }
 
-#include <string>
 void vcModals_DrawProjectInfo(vcState *pProgramState)
 {
   if (pProgramState->openModals & (1 << vcMT_ProjectInfo))
     ImGui::OpenPopup(udTempStr("%s###projectInfo", pProgramState->activeProject.pRoot->pName));
-  
+
   ImGui::SetNextWindowSize(ImVec2(600, 600), ImGuiCond_Appearing);
   if (ImGui::BeginPopupModal(udTempStr("%s###projectInfo", pProgramState->activeProject.pRoot->pName), nullptr, ImGuiWindowFlags_NoSavedSettings))
   {
@@ -1235,9 +1234,6 @@ void vcModals_DrawProjectSettings(vcState *pProgramState)
 
       ImGui::CloseCurrentPopup();
     }
-
-    if (ImGui::Button(vcString::Get("menuProjectInfoPreview"), ImVec2(-1, 0)))
-      pProgramState->openModals |= 1 << vcMT_ProjectInfo;
 
     ImGui::EndPopup();
   }
--- a/src/vcModals.cpp
+++ b/src/vcModals.cpp
@@ -26,7 +26,6 @@
 #include "udServerAPI.h"
 
 #include "stb_image.h"
-#include "vcTexture.h"
 
 bool gShowInputControlsNextHack = false;
 
@@ -632,28 +631,31 @@ void vcModals_DrawProjectInfo(vcState *pProgramState)
     else
       pProgramState->modalOpen = true;
 
-    const char *pInfo = nullptr;
-
     if (ImGui::BeginChild("##infoModal", ImVec2(-1, -30), true))
     {
-      if (udProjectNode_GetMetadataString(pProgramState->activeProject.pRoot, "information", &pInfo, "") == udE_Success)
+      for (size_t i = 0; i < pProgramState->projectInfoTextures.infoStrings.length; ++i)
       {
-        for (size_t i = 0; i < pProgramState->projectInfoTextures.infoStrings.length; ++i)
+        ImGui::TextWrapped("%s", pProgramState->projectInfoTextures.infoStrings[i]);
+        if (i < pProgramState->projectInfoTextures.infoStrings.length - 1)
         {
-          ImGui::TextWrapped("%s", pProgramState->projectInfoTextures.infoStrings[i]);
-          if (i < pProgramState->projectInfoTextures.infoStrings.length - 1)
+          if (*pProgramState->projectInfoTextures.textures[i] == nullptr)
           {
-            if (pProgramState->projectInfoTextures.textures[i] == nullptr)
-              ImGui::TextWrapped("%s", pProgramState->projectInfoTextures.textureAltStrings[i]);
+            ImGui::TextWrapped("%s", pProgramState->projectInfoTextures.textureAltStrings[i]);
+          }
+          else
+          {
+            int w, h;
+            if (vcTexture_GetSize(*pProgramState->projectInfoTextures.textures[i], &w, &h) == udR_Success && (w > 1 || h > 1))
+              ImGui::Image(*pProgramState->projectInfoTextures.textures[i], ImVec2((float)w, (float)h));
             else
-              ImGui::Image(pProgramState->projectInfoTextures.textures[i], pProgramState->projectInfoTextures.textureSizes[i]);
+              ImGui::TextWrapped("%s", pProgramState->projectInfoTextures.textureAltStrings[i]);
           }
         }
       }
     }
     ImGui::EndChild();
 
-    if (ImGui::Button(vcString::Get("popupOK"), ImVec2(-1, 0)) || vcHotkey::IsPressed(vcB_Cancel) || pInfo == nullptr || pInfo[0] == '\0')
+    if (ImGui::Button(vcString::Get("popupOK"), ImVec2(-1, 0)) || vcHotkey::IsPressed(vcB_Cancel) || pProgramState->projectInfoTextures.infoStrings.length == 0)
       ImGui::CloseCurrentPopup();
 
     ImGui::EndPopup();
--- a/src/vcModals.cpp
+++ b/src/vcModals.cpp
@@ -26,6 +26,7 @@
 #include "udServerAPI.h"
 
 #include "stb_image.h"
+#include "vcTexture.h"
 
 bool gShowInputControlsNextHack = false;
 
@@ -618,12 +619,12 @@ void vcModals_DrawExportProject(vcState *pProgramState)
   }
 }
 
-
+#include <string>
 void vcModals_DrawProjectInfo(vcState *pProgramState)
 {
   if (pProgramState->openModals & (1 << vcMT_ProjectInfo))
     ImGui::OpenPopup(udTempStr("%s###projectInfo", pProgramState->activeProject.pRoot->pName));
-
+  
   ImGui::SetNextWindowSize(ImVec2(600, 600), ImGuiCond_Appearing);
   if (ImGui::BeginPopupModal(udTempStr("%s###projectInfo", pProgramState->activeProject.pRoot->pName), nullptr, ImGuiWindowFlags_NoSavedSettings))
   {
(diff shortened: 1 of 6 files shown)


src/vcModals.cpp
```diff
@@ -656,10 +656,89 @@ void vcModals_DrawProjectInfo(vcState *pProgramState)
...
+            udStrchr(pInfo + startImagePos, "[", &openBracketPos);
+            openBracketPos += startImagePos;
+            udStrchr(pInfo + openBracketPos, "]", &closeBracketPos);
+            closeBracketPos += openBracketPos;
+            udStrchr(pInfo + closeBracketPos, "(", &openParanthesisPos);
+            openParanthesisPos += closeBracketPos;
+            udStrchr(pInfo + openParanthesisPos, ")", &closeParanthesisPos);
+            closeParanthesisPos += openParanthesisPos;
+            
+            const int64_t invalidLen = udStrlen(pInfo);
```
samsinsane, line 693: This operation isn't free, you execute this N+1 times, where N is the number of valid `alt` patterns. However, it won't change. Additionally, `udStrchr` returns `nullptr` when the character isn't found which would allow you to exit early more often.

src/vcModals.cpp
```diff
@@ -656,10 +656,89 @@ void vcModals_DrawProjectInfo(vcState *pProgramState)
...
+              closeBracketPos == invalidLen ||
+              openParanthesisPos == invalidLen ||
+              closeParanthesisPos == invalidLen)
+              break;
+
+            size_t lastCharPos = closeParanthesisPos;
+            if (openBracketPos == startImagePos + 1 &&
+              closeBracketPos > openBracketPos &&
+              openParanthesisPos == closeBracketPos + 1 &&
+              closeParanthesisPos > openParanthesisPos)
```
samsinsane, line 706: I'm not sure that `openBracketPos` is necessary, it should always be `startImagePos + 1`, so instead of using `udStrchr` to find it, just check that the character after `startImagePos` is `[` instead. Same goes for `openParenthesisPos`. Additionally, `closeBracketPos` is calculated from `pInfo + openBracketPos`, so `closeBracketPos > openBracketPos` will always be true by this point. Same goes for `closeParenthesisPos > openParenthesisPos`.

src/vcModals.cpp
```diff
@@ -656,10 +656,89 @@ void vcModals_DrawProjectInfo(vcState *pProgramState)
...
+          // Discover Images
+          int firstIndex = 0;
+          int startStrIndex = 0;
+          while (true)
+          {
+            size_t startImagePos = 0;
+            size_t openBracketPos = 0;
+            size_t closeBracketPos = 0;
+            size_t openParanthesisPos = 0;
+            size_t closeParanthesisPos = 0;
```
samsinsane, line 680: These should be `Parenthesis`.

src/vcModals.cpp
```diff
@@ -656,10 +656,89 @@ void vcModals_DrawProjectInfo(vcState *pProgramState)
...
+              pProgramState->projectInfoTextures.infoStrings.PushBack(pSubStr);
+
+              const char *pTexPath = udStrndup(pInfo + (openParanthesisPos + 1), closeParanthesisPos - openParanthesisPos - 1);
+              uint32_t w, h;
+              vcTexture *pNewTexture = nullptr;
+              vcTexture_CreateFromFilename(&pNewTexture, pTexPath, &w, &h);
+              pProgramState->projectInfoTextures.textures.PushBack(pNewTexture);
+              pProgramState->projectInfoTextures.textureSizes.PushBack(ImVec2((float)w, (float)h));
+
+              firstIndex = (int)(lastCharPos + 1);
```
samsinsane, line 718: This happens in both cases, and isn't modified - this should be set after `lastCharPos` is instead.

src/vcModals.cpp
```diff
@@ -656,10 +656,89 @@ void vcModals_DrawProjectInfo(vcState *pProgramState)
...
+              startStrIndex = firstIndex;
+            }
+            else
+            {
+              firstIndex = (int)(lastCharPos + 1);
+            }
+          }
+          
+          const char *pSubStr = udStrndup(pInfo + firstIndex, udStrlen(pInfo) - firstIndex);
```
samsinsane, line 727: This is just `udStrdup(pInfo + firstIndex)`

src/vcModals.cpp
```diff
@@ -656,10 +656,89 @@ void vcModals_DrawProjectInfo(vcState *pProgramState)
...
+          pProgramState->projectInfoTextures.infoStrings.PushBack(pSubStr);
+        }
+
+        for (size_t i = 0; i < pProgramState->projectInfoTextures.infoStrings.length; ++i)
+        {
+          ImGui::TextWrapped("%s", pProgramState->projectInfoTextures.infoStrings[i]);
+          if (i < pProgramState->projectInfoTextures.infoStrings.length - 1 && pProgramState->projectInfoTextures.textures[i] != nullptr)
+            ImGui::Image(pProgramState->projectInfoTextures.textures[i], pProgramState->projectInfoTextures.textureSizes[i]);
```
samsinsane, line 735: The task explicitly states that if the image isn't available to render the alt text.

src/vcModals.cpp
```diff
@@ -656,10 +656,89 @@ void vcModals_DrawProjectInfo(vcState *pProgramState)
...
+        }
+      }
+
+      ImGui::EndChild();
     }
-    ImGui::EndChild();
```
samsinsane, line 661: Just noticed your "Rebase Fix" commit and didn't see this during my original review, this needs to be called here.

Jossos:
> You haven't linked the issue in the description. Also, based on what I can see so far, this is non-functional in a real world use case.
> 
> ```
> First Paragraph
> First Image
> Second Paragraph
> Second Image
> Final Paragraph
> ```
> 
> Should look like:
> 
> > First Paragraph
> > First Image
> > Second Paragraph
> > Second Image
> > Final Paragraph
> 
> I'm confident that your implementation will not look like this, and will instead look more like:
> 
> > First Image
> > Second Paragraph
> > Second Image
> > Final Paragraph
> > Second Image
> > Final Paragraph

No, it looks like the first example. i tested it with 2 images and text in between. it doesn't paste the final text twice or anything like that.

src/vcModals.cpp
Jossos: I need the position for splitting up the string so i think it's necessary. this function doesn't get called as much now, so the performance hit i think it acceptable

samsinsane: > I need the position for splitting up the string so i think it's necessary

You don't, you only need to know where `!`, `]` and `)` are, the characters that follow `!` and `]` are `[` and `(`, respectively. You can remove `openBracketPos` and `openParenthesisPos` and use `startImagePos + 1` and `closeBracketPos + 1`, respectively. Instead of checking `openBracketPos == startImagePos + 1` check `pInfo[startImagePos + 1] == '['`, similarly for `openParenthesis`.

> this function doesn't get called as much now, so the performance hit i think it acceptable

If this was about performance, I would be explaining how you should be doing this without allocating so much memory, a single `udStrdup` is sufficient. This is about readability and making your intentions clear.

src/vcModals.cpp
```diff
@@ -26,6 +26,7 @@
 #include "udServerAPI.h"
 
 #include "stb_image.h"
+#include "vcTexture.h"
```
samsinsane, line 29: This is already included on line 5.

src/vcModals.cpp
```diff
@@ -656,10 +656,22 @@ void vcModals_DrawProjectInfo(vcState *pProgramState)
     if (ImGui::BeginChild("##infoModal", ImVec2(-1, -30), true))
     {
       if (udProjectNode_GetMetadataString(pProgramState->activeProject.pRoot, "information", &pInfo, "") == udE_Success)
```
samsinsane, line 658: Is this still required? Won't `pProgramState->projectInfoTextures.infoStrings.length` be 0 when this returns false?

src/vcProject.cpp
```diff
@@ -300,7 +302,7 @@ bool vcProject_LoadFromServer(vcState *pProgramState, const char *pProjectID)
 
     vcModals_OpenModal(pProgramState, vcMT_ProjectChange);
   }
-
+
```
samsinsane, line 305: Revert.

src/vcProject.cpp
```diff
@@ -356,7 +360,7 @@ bool vcProject_LoadFromURI(vcState *pProgramState, const char *pFilename)
...
```
samsinsane, line 363: Revert.

src/vcProject.cpp
```diff
@@ -833,3 +837,101 @@ void vcProject_ExtractAttributionText(udProjectNode *pFolderNode, const char **p
 
   
 }
+
+void vcProject_UpdateProjectInformationDisplayTextures(vcState *pProgramState)
+{
+  const char *pInfo = nullptr;
+  if (udProjectNode_GetMetadataString(pProgramState->activeProject.pRoot, "information", &pInfo, "") == udE_Success)
+  {
+    if (pProgramState->projectInfoTextures.pLastInfoText != pInfo)
```
samsinsane, line 846: These will never be equal - you duplicate it two lines down.

src/vcProject.cpp
```diff
@@ -833,3 +837,101 @@ void vcProject_ExtractAttributionText(udProjectNode *pFolderNode, const char **p
...
+        if (udStrchr(pInfo + openParenthesisPos, ")", &closeParenthesisPos) == nullptr)
+          break;
+        closeParenthesisPos += openParenthesisPos;
+
+        if (startImagePos == invalidLen ||
+          openBracketPos == invalidLen ||
+          closeBracketPos == invalidLen ||
+          openParenthesisPos == invalidLen ||
+          closeParenthesisPos == invalidLen)
+          break;
```
samsinsane, line 892: This can never be hit - when any of these are `invalidLen`, `udStrchr` will return `nullptr`.

src/vcProject.cpp
```diff
@@ -833,3 +837,101 @@ void vcProject_ExtractAttributionText(udProjectNode *pFolderNode, const char **p
...
+
+        size_t lastCharPos = closeParenthesisPos;
+        firstIndex = (int)(lastCharPos + 1);
+
+        if (openBracketPos == startImagePos + 1 &&
+          closeBracketPos > openBracketPos &&
+          openParenthesisPos == closeBracketPos + 1 &&
+          closeParenthesisPos > openParenthesisPos)
```
samsinsane, line 900: This should be on a single line. As previously discussed, the 2nd and 4th conditions can never be false.

src/vcProject.cpp
```diff
@@ -833,3 +837,101 @@ void vcProject_ExtractAttributionText(udProjectNode *pFolderNode, const char **p
...
+        {
+          // Text
+          size_t prevStrLen = startImagePos - startStrIndex;
+          if (prevStrLen > 0)
```
samsinsane, line 904: When can `prevStrLen` be 0?

src/vcProject.cpp
```diff
@@ -833,3 +837,101 @@ void vcProject_ExtractAttributionText(udProjectNode *pFolderNode, const char **p
...
+          {
+            const char *pSubStr = udStrndup(pInfo + startStrIndex, prevStrLen);
+            pProgramState->projectInfoTextures.infoStrings.PushBack(pSubStr);
+          }
+
+          // Texture Alt Text
+          if (closeBracketPos == openBracketPos + 1)
+          {
+            pProgramState->projectInfoTextures.textureAltStrings.PushBack((const char *)nullptr);
```
samsinsane, line 913: Why do you do this? Consider commenting the code.

src/vcProject.cpp
```diff
@@ -833,3 +837,101 @@ void vcProject_ExtractAttributionText(udProjectNode *pFolderNode, const char **p
...
+          {
+            const char *pSubStr = udStrndup(pInfo + openBracketPos + 1, closeBracketPos - openBracketPos - 1);
+            pProgramState->projectInfoTextures.textureAltStrings.PushBack(pSubStr);
+          }
+
+          // Texture
+          const char *pTexPath = udStrndup(pInfo + (openParenthesisPos + 1), closeParenthesisPos - openParenthesisPos - 1);
+          uint32_t w, h;
+          vcTexture *pNewTexture = nullptr;
+          vcTexture_CreateFromFilename(&pNewTexture, pTexPath, &w, &h);
```
samsinsane, line 925: This will probably need to end up being one of the async functions for two reasons:
- If this is on the main thread, it's going to stall performing the HTTP request for the image
- If this isn't on the main thread, the GPU texture handle will be created outside of the main thread

src/vcModals.cpp
```diff
@@ -656,10 +656,89 @@ void vcModals_DrawProjectInfo(vcState *pProgramState)
...
-        ImGui::TextWrapped("%s", pInfo);
+      {
+        if (udStrcmp(pProgramState->projectInfoTextures.pLastInfoText, pInfo) != 0)
```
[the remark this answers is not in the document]
Jossos: fixed (compares pointers now)

src/vcModals.cpp
```diff
@@ -656,10 +656,89 @@ void vcModals_DrawProjectInfo(vcState *pProgramState)
...
+        {
+          pProgramState->projectInfoTextures.pLastInfoText = udStrdup(pInfo);
```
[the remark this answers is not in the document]
Jossos: fixed

src/vcModals.cpp
Jossos: moved outside while loop

src/vcModals.cpp
Jossos: Fixed

src/vcModals.cpp
Jossos: Fixed

src/vcModals.cpp
Jossos: Fixed

src/vcModals.cpp
Jossos: Fixed

src/vcModals.cpp
Jossos: actually i can remove this if completely. the for loop will do nothing if the array is empty

src/vcProject.cpp
Jossos: the alt text and texture arrays need to match

samsinsane: This doesn't explain why you specifically push back nullptr instead of what should be an empty string in the else case.

src/vcModals.cpp
```diff
@@ -632,16 +631,31 @@ void vcModals_DrawProjectInfo(vcState *pProgramState)
...
     if (ImGui::BeginChild("##infoModal", ImVec2(-1, -30), true))
     {
-      if (udProjectNode_GetMetadataString(pProgramState->activeProject.pRoot, "information", &pInfo, "") == udE_Success)
-        ImGui::TextWrapped("%s", pInfo);
+      for (size_t i = 0; i < pProgramState->projectInfoTextures.infoStrings.length; ++i)
+      {
+        ImGui::TextWrapped("%s", pProgramState->projectInfoTextures.infoStrings[i]);
+        if (i < pProgramState->projectInfoTextures.infoStrings.length - 1)
+        {
+          if (*pProgramState->projectInfoTextures.textures[i] == nullptr)
```
samsinsane, line 641: This check is unnecessary, `vcTexture_GetSize` will return `udR_InvalidParameter` when the texture is `nullptr`.

src/vcProject.cpp
```diff
@@ -844,3 +848,66 @@ void vcProject_ExtractAttributionText(udProjectNode *pFolderNode, const char **p
...
+        const char *pSubStr = udStrndup(pInfo + startStrIndex, prevStrLen);
+        pProgramState->projectInfoTextures.infoStrings.PushBack(pSubStr);
+
+        // Texture Alt Text
+        const char *pAltStr = udStrndup(pInfo + startImagePos + 2, closeBracketPos - startImagePos - 2);
+        pProgramState->projectInfoTextures.textureAltStrings.PushBack(pAltStr);
+
+        // Texture
+        const char *pTexPath = udStrndup(pInfo + (closeBracketPos + 2), closeParenthesisPos - closeBracketPos - 2);
+        vcTexture **ppNewTexture = udAllocType(vcTexture*, 1, udAF_Zero);
```
samsinsane, line 901: This is not a great way of doing this. If you keep the `textures` chunked array as `udChunkedArray<vcTexture*>` you can just push back with no parameters and it will return a `vcTexture **`. The function is marked as deprecated, but you can ignore this as we use it often and it's deprecation was not discussed properly, and breaks the pattern you want right here due to argument ambiguity with the other two overloads.

src/vcState.h
```diff
@@ -260,6 +260,13 @@ struct vcState
     vcImageType imageType;
   } image, screenshot;
 
+  struct 
+  {
+    udChunkedArray<const char *> infoStrings;
+    udChunkedArray<const char *> textureAltStrings;
+    udChunkedArray<vcTexture**> textures;
```
samsinsane, line 267: As mentioned above, this should be `udChunkedArray<vcTexture*> textures;`

src/vcProject.cpp
```diff
@@ -844,3 +848,66 @@ void vcProject_ExtractAttributionText(udProjectNode *pFolderNode, const char **p
...
+        break;
+      closeParenthesisPos += closeBracketPos + 1;
+        
+      size_t lastCharPos = closeParenthesisPos;
+      firstIndex = (int)(lastCharPos + 1);
+
+      if (pInfo[startImagePos + 1] == '[' && pInfo[closeBracketPos + 1] == '(')
+      {
+        // Text
+        size_t prevStrLen = startImagePos - startStrIndex;
```
samsinsane, line 891: This variable is kind of unnecessary, and inconsistent with the `udStrndup` calls below.

samsinsane (approved):

PR #942, opened 2020-08-17
Enabled multi-viewport rendering

peterAdams1014:


src/main.cpp
```diff
@@ -2594,201 +2595,255 @@ void vcMain_RenderSceneWindow(vcState *pProgramState)
...
+      ImGui::Image(renderData.pSceneTexture, ImVec2((float)pProgramState->pActiveViewport->resolution.x, (float)pProgramState->pActiveViewport->resolution.y), ImVec2(0, 0), ImVec2(renderData.sceneScaling.x, renderData.sceneScaling.y));
+
+      // TODO: Screenshot which viewport?
+      if (viewportIndex == 0 && pProgramState->settings.screenshot.taking)
+        pProgramState->screenshot.pImage = renderData.pSceneTexture;
+
+      static bool wasContextMenuOpenLastFrame = false;
+      bool useTool = (io.MouseDragMaxDistanceSqr[0] < (io.MouseDragThreshold * io.MouseDragThreshold)) && ImGui::IsMouseReleased(0) && ImGui::IsItemHovered();
+
+      // TODO: TODO: global state in here makes it not work for multiple viewports
```
samsinsane, line 2651: Is this a TODO to do a TODO?

src/main.cpp
```diff
@@ -2594,201 +2595,255 @@ void vcMain_RenderSceneWindow(vcState *pProgramState)
...
 
-    if ((io.MouseDragMaxDistanceSqr[1] < (io.MouseDragThreshold * io.MouseDragThreshold) && ImGui::BeginPopupContextItem("SceneContext")))
-    {
-      static bool hadMouse = false;
-      static udDouble3 mousePosCartesian;
-      static udDouble3 mousePosLongLat;
+      pProgramState->pActiveViewport = &pProgramState->pViewports[viewportIndex];
 
-      if (!wasContextMenuOpenLastFrame || ImGui::IsMouseClicked(1))
+      // TODO: Screenshot which viewport?
```
samsinsane, line 2628: Both as a single image - if the user wants a screenshot of just one, they should close the other viewport.

src/main.cpp
peterAdams1014: :P

src/main.cpp
peterAdams1014: Sounds reasonable.

PR #991, opened 2020-08-30
Allow users to override existing project with a new porject

jproust:
Trigger a popup window when using ImGui to select the file to override. (Windows when not using native UI and all other systems)
Fixes [AB#2078](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/2078)

### DIFF
--- a/src/vcProject.cpp
+++ b/src/vcProject.cpp
@@ -149,7 +149,7 @@ udError vcProject_CreateFileScene(vcState *pProgramState, const char *pFileName,
 
   if (overrideProject)
   {
-      if (!vcProject_SaveAs(pProgramState, pFileName, true))
+      if (!vcProject_SaveAs(pProgramState, pFileName, pProgramState->settings.window.useNativeUI))
         return udE_WriteFailure;
   }
 
--- a/src/vcProject.cpp
+++ b/src/vcProject.cpp
@@ -126,32 +126,35 @@ udError vcProject_CreateFileScene(vcState *pProgramState, const char *pFileName,
 
   udError error = udE_Success;
   udProject* pNewProject = nullptr;
-  bool overrideProject = false;
 
   if (udFileExists(pFileName) == udR_Success)
   {
-    overrideProject = true;
-    error = udProject_CreateInMemory(&pNewProject, pProjectName);
-    if (error != udE_Success)
-      return error;
+    if (!pProgramState->settings.window.useNativeUI)
+    {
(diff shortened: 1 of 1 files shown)


src/vcProject.cpp
```diff
@@ -124,20 +124,39 @@ udError   vcProject_CreateFileScene(vcState *pProgramState, const char *pFileNam
...
-  if (error != udE_Success)
-    return error;
+  udError error = udE_Success;
+  udProject* pNewProject = nullptr;
+  bool overrideProject = false;
 
+  if (udFileExists(pFileName) == udR_Success)
+  {
+    overrideProject = true;
+    error = udProject_CreateInMemory(&pNewProject, pProjectName);
```
samsinsane, line 134: This seems wrong? @OsiJr does `udProject_CreateInFile` continue to write to disk as changes are made?

src/vcProject.cpp
```diff
@@ -124,20 +124,39 @@ udError   vcProject_CreateFileScene(vcState *pProgramState, const char *pFileNam
...
+  }
   if (pProgramState->activeProject.pProject != nullptr)
     vcProject_Deinit(pProgramState, &pProgramState->activeProject);
 
   pProgramState->activeProject.pProject = pNewProject;
   vcProject_InitScene(pProgramState, srid);
 
+  if (overrideProject)
+  {
+#if UDPLATFORM_WINDOWS
```
samsinsane, line 152: This check isn't necessary, `pProgramState->settings.window.useNativeUI` will be `false` on all platforms that aren't supported. By putting in this explicit check, you're creating future bugs as we expand the native UI support.

samsinsane (changes requested):

src/vcProject.cpp
jproust: Fixed

src/vcProject.cpp
jproust: @samsinsane Would you prefer that I use udProject_CreateInFile with a temp file and once saved to the file provided, delete the temp file ?

samsinsane: If my understanding is correct, `udProject_CreateInFile` will write any future changes to the file provided. The correct thing would be to do the detection immediately, and handle it before we ever get to a `udProject_*` function - assuming my understanding is correct.

samsinsane (approved):

PR #994, opened 2020-08-31
use 2 clicks instead of drag for creation of filters

jproust:
Fixes [AB#2094](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/2094)

### DIFF
--- a/src/scene/vcQueryNode.cpp
+++ b/src/scene/vcQueryNode.cpp
@@ -292,7 +292,7 @@ vcGizmoAllowedControls vcQueryNode::GetAllowedControls()
     return vcGAC_All;
 }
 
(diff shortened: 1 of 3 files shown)


samsinsane (changes requested):
While I think this change is the right direction, I would like to see the clicking code move to `src/scene/tools/vcAdd*Filter`. Look at how `vcAnnotate.cpp` and the measuring tools work to see how the mouse input works (handle picking I think). You should be able to remove code from `vcCamera.cpp` since we already handle the mouse clicks and what not.

samsinsane (approved):
As discussed on Teams, the other changes will come through in another PR as it makes more sense to do this in pieces.

PR #997, opened 2020-09-01
Disabled Filter tool if click happens on New or Open Project

jproust:
Fixes [AB#2098](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/2098)

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1667,10 +1667,16 @@ void vcRenderSceneUI(vcState *pProgramState, const ImVec2 &windowPos, const ImVe
       vcMain_ProfileMenu(pProgramState);
 
       if (vcMenuBarButton(pProgramState->pUITexture, vcString::Get("menuNewScene"), vcB_Invalid, vcMBBI_NewProject, vcMBBG_SameGroup))
+      {
+        pProgramState->activeTool = vcActiveTool_Select;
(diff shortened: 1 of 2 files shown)


src/main.cpp
```diff
@@ -1667,10 +1667,16 @@ void vcRenderSceneUI(vcState *pProgramState, const ImVec2 &windowPos, const ImVe
       vcMain_ProfileMenu(pProgramState);
 
       if (vcMenuBarButton(pProgramState->pUITexture, vcString::Get("menuNewScene"), vcB_Invalid, vcMBBI_NewProject, vcMBBG_SameGroup))
+      {
+        pProgramState->activeTool = vcActiveTool_Select;
```
samsinsane, line 1671: From memory, there's a common code path when create and load project are successful, could this be put there instead? That way, any time a project changes (dropping a JSON file into the scene for example) this will always happen, instead of in the cases we've handled.

samsinsane (changes requested):

src/main.cpp
jproust: Moved the deactivation of the Active Tool in vcProject_Deinit()

samsinsane (dismissed):

samsinsane (approved):

PR #1005, opened 2020-09-02
Fixed an issue where screenshot viewing (opengl) was upside down

peterAdams1014:
Fixed an issue where while a screenshot was in progress, the viewport display would show the incorrect image.
Fixes [AB#1237](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1237)

### DIFF
--- a/src/vcModals.cpp
+++ b/src/vcModals.cpp
@@ -1386,12 +1386,7 @@ void vcModals_DrawImageViewer(vcState *pProgramState)
       ImVec2 window = ImGui::GetWindowSize();
       ImVec2 windowPos = ImGui::GetWindowPos();
 
-#if GRAPHICS_API_OPENGL
-      ImVec2 uvs[2] = { {0,1}, {1,0} };
-#else
-      ImVec2 uvs[2] = { {0,0}, {1,1} };
-#endif
-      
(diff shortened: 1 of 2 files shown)


src/vcModals.cpp
```diff
@@ -1386,12 +1386,7 @@ void vcModals_DrawImageViewer(vcState *pProgramState)
       ImVec2 window = ImGui::GetWindowSize();
       ImVec2 windowPos = ImGui::GetWindowPos();
 
-#if GRAPHICS_API_OPENGL
-      ImVec2 uvs[2] = { {0,1}, {1,0} };
-#else
-      ImVec2 uvs[2] = { {0,0}, {1,1} };
-#endif
-      
+      ImVec2 uvs[2] = { {0,0}, {1,1} };
```
samsinsane, line 1389: Does this still work for images that are loaded in?

samsinsane (changes requested):

src/vcModals.cpp
peterAdams1014: Yes.

samsinsane (approved):

PR #1009, opened 2020-09-02
Fixed Enter Key state being released

jproust:
Fixes AB#2086

### DIFF
--- a/src/imgui_ex/imgui_impl_sdl.cpp
+++ b/src/imgui_ex/imgui_impl_sdl.cpp
@@ -231,7 +231,7 @@ bool ImGui_ImplSDL2_ProcessEvent(const SDL_Event* event)
             // TODO: Come back when https://github.com/ocornut/imgui/pull/2005 is merged
             //       and add ImGuiKey_NumericEnter to our io.KeyMap below
             if (key == SDL_SCANCODE_KP_ENTER)
-              io.KeysDown[SDL_SCANCODE_RETURN] = io.KeysDown[SDL_SCANCODE_RETURN] || (event->type == SDL_KEYDOWN);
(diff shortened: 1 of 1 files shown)


src/imgui_ex/imgui_impl_sdl.cpp
```diff
@@ -231,7 +231,7 @@ bool ImGui_ImplSDL2_ProcessEvent(const SDL_Event* event)
             // TODO: Come back when https://github.com/ocornut/imgui/pull/2005 is merged
             //       and add ImGuiKey_NumericEnter to our io.KeyMap below
             if (key == SDL_SCANCODE_KP_ENTER)
-              io.KeysDown[SDL_SCANCODE_RETURN] = io.KeysDown[SDL_SCANCODE_RETURN] || (event->type == SDL_KEYDOWN);
+              io.KeysDown[SDL_SCANCODE_RETURN] = !io.KeysDown[SDL_SCANCODE_RETURN] || (event->type == SDL_KEYDOWN);
```
samsinsane, line 234: This was supposed to be the same as the normal case below, looks like I copied the wrong one. Does it work if it's just `= (event->type == SDL_KEYDOWN);`?

samsinsane (changes requested):

src/imgui_ex/imgui_impl_sdl.cpp
jproust: Looks like it does.
Changed

samsinsane: Oops, my bad!

jproust: don't merge ...

samsinsane (approved):

PR #1020, opened 2020-09-07
Added show angles option to line and area tools

int-Frank:
Fixes [AB#1837](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1837)

### DIFF
--- a/src/scene/vcPOI.cpp
+++ b/src/scene/vcPOI.cpp
@@ -1142,7 +1142,6 @@ void vcPOI::AddLengths(const vcUnitConversionData *pConversionData)
   // j = previous, i = current
   int i = m_line.closed ? 0 : 1;
   int j = m_line.closed ? m_line.numPoints - 1 : 0;
-  //int iMax = m_line.closed ? m_line.numPoints : m_line.numPoints - 1;
   for (; i < m_line.numPoints; i++)
   {
     double lineLength = udMag3(m_line.pPoints[j] - m_line.pPoints[i]);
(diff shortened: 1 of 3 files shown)


jproust:
Angles_issue
What side is the 177.27 deg angle

int-Frank:
> Angles_issue
> What side is the 177.22 deg angle

I use the  _smallest_ angle between the lines. Or in other words, the inverse cosine of the dot product between the vectors of the two lines.

jproust:
> I use the _smallest_ angle between the lines. Or in other words, the inverse cosine of the dot product between the vectors of the two lines.

I know I have read your code, I just find the display of the angle a bit confusing when it's close to 180° as the user can't really know which side the angle is. It can be confusing if for instance the smallest angle is an angle in the z plane and the user is watching from above.
Anyway we can display the measured angle ?
angle_issue

src/scene/vcPOI.cpp
```diff
@@ -68,6 +68,9 @@ class vcPOIState_General
       if (ImGui::Checkbox(udTempStr("%s##POIShowAllLengths%zu", vcString::Get("scenePOILineShowAllLengths"), itemID), &m_pParent->m_showAllLengths))
         udProjectNode_SetMetadataBool(m_pParent->m_pNode, "showAllLengths", m_pParent->m_showAllLengths);
 
+      if (ImGui::Checkbox(udTempStr("%s##POIShowAngles%zu", vcString::Get("scenePOILineShowAngles"), itemID), &m_pParent->m_showAngles))
+        udProjectNode_SetMetadataBool(m_pParent->m_pNode, "showAngles", m_pParent->m_showAngles);
```
OsiJr, line 72: `OnNodeUpdate` doesn't load this key (there is a helper in vcProject for bools)

OsiJr (changes requested):
Will need to load the show angles setting

int-Frank:
@jproust Ah I see. Perhaps I could render a little angle symbol between the lines. Something like 
image
 @OsiJr What do you think?

OsiJr (approved):

OsiJr (approved):

PR #1029, opened 2020-09-08
Changed where the tool updates

int-Frank:
Fixes [AB#1741](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/1741)

Ok, from what I could gather, saving and loading data to and from the project node each frame (in `ApplyToScene()`) was introducing floating point drift. I don't think there should be a need to save out every frame anyway. I changed this to only save out when the preview point has changed, and when preview has finished, in addition to `ApplyDelta()` and `OnNodeUpdate()`

This has been tested changing GeoZones.


samsinsane (changes requested):
Is this the only scene item that continuously updates the points? Happy to merge this otherwise.

int-Frank:
@samsinsane I think so. I checked the POI item, and it does not update every frame.

samsinsane:
This should probably be updated to merge into the next release instead.

PR #1059, opened 2020-09-15
Added basic 12dxml support

int-Frank:
First block of work for [AB#489](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/489). This PR currently implements:
 - load UTF-8 encoded 12dxml file
 - loading 12dxml models (converts to folder in udStream)
 - loading 3D point and line data from 12dxml superstrings
 - `<null></null>` and `<colour></colour>` 'commands' from the xml
 - a few essential tags to display data (name, colours, weights, ...)

Not currently implemented
 - non UTF-8 encoded files
 - 2D and 3D varients of arcs, circles, transitions, offset transitions
 - different project structures (there are at least 2 different structures I have seen)

EDIT: The idea behind the `12dxml` loader being multiple files is to separate the 'loading' of the `12dxml` file from the creation of a udStream project. If we eventually want to support the full `12dxml` spec, both of these might become very large things, and it may be easier to handle if they were separate modules. Otherwise I'm happy to re-submit with everything in one file.

### DIFF
--- a/src/parsers/vc12DXMLLoad.cpp
+++ b/src/parsers/vc12DXMLLoad.cpp
@@ -15,9 +15,9 @@
 *                not be obeyed!
 */
 
-#define LOG_ERROR(...)
-#define LOG_WARNING(...)
-#define LOG_INFO(...)
+#define LOG_ERROR(...) {}
+#define LOG_WARNING(...) {}
+#define LOG_INFO(...) {}
 
 // TODO support 'light' and 'dark' variants of all colours
 // TODO support capitalisation
--- a/src/parsers/vc12DXMLLoad.cpp
+++ b/src/parsers/vc12DXMLLoad.cpp
@@ -164,8 +164,8 @@ udResult vc12DXML_SuperString::Build(udJSON const *pNode, vc12DXML_ProjectGlobal
 {
   udResult result;
   const udJSON *pItem = nullptr;
-  bool has_z = false;
-  double z = 0.0;
+  //bool has_z = false;
+  //double z = 0.0;
 
   UD_ERROR_IF(pNode == nullptr, udR_InvalidParameter_);
 
@@ -175,14 +175,14 @@ udResult vc12DXML_SuperString::Build(udJSON const *pNode, vc12DXML_ProjectGlobal
   // From the spec:
   //   - There is exactly 0 or 1 <z> tag per string.
   //   - Its position is irrelevant, it is global to the string
-  pItem = &pNode->Get("z");
-  UD_ERROR_IF(pItem->IsArray(), udR_InvalidConfiguration);
-
-  if (!pItem->IsVoid())
-  {
-    has_z = true;
-    z = pItem->AsDouble();
-  }
+  //pItem = &pNode->Get("z");
+  //UD_ERROR_IF(pItem->IsArray(), udR_InvalidConfiguration);
+  //
+  //if (!pItem->IsVoid())
+  //{
+  //  has_z = true;
+  //  z = pItem->AsDouble();
+  //}
 
   SetGlobals(pNode, globals);
 
--- a/src/parsers/vc12DXMLLoad.cpp
+++ b/src/parsers/vc12DXMLLoad.cpp
@@ -20,7 +20,6 @@
 #define LOG_INFO(...) {}
 
 // TODO support 'light' and 'dark' variants of all colours
-// TODO support capitalisation
 static uint32_t vc12DXML_ToColour(const char *pStr)
 {
   struct ColourPair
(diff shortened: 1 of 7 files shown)


src/parsers/vc12dxmlCommon.h
```diff
@@ -0,0 +1,85 @@
...
+  udResult Build(udJSON const *pNode, vc12DXML_ProjectGlobals &globals) override;
+  void Print(FILE *pOut, int indent) const override;
+  void AddToProject(vcState *pProgramState, udProjectNode *pParent) override;
+private:
+
+  char *m_pName;
+  bool m_isClosed;
+  double m_weight;
+  uint32_t m_colour;
+  std::vector<udDouble3> m_points;
```
samsinsane, line 46: Why `std::vector` over `udChunkedArray`?

src/parsers/vc12dxmlCommon.h
```diff
@@ -0,0 +1,85 @@
+#ifndef vc12dxmlCommon_h__
+#define vc12dxmlCommon_h__
+
+#include <stdint.h>
+#include <vector>
+#include <cstdio>
```
samsinsane, line 6: Do you specifically need the C++ version of `stdio.h`?

src/parsers/vc12dxmlCommon.h
```diff
@@ -0,0 +1,85 @@
...
+  ~vc12DXML_Model();
+  udResult BuildChildren(udJSON const *pNode, vc12DXML_ProjectGlobals &globals);
+  udResult Build(udJSON const *pNode, vc12DXML_ProjectGlobals &globals) override;
+  void Print(FILE *pOut, int indent) const override;
+  void AddToProject(vcState *pProgramState, udProjectNode *pParent) override;
+
+private:
+
+  char *m_pName;
+  std::vector<vc12DXML_Item *> m_elements;
```
samsinsane, line 63: Same as above?

src/parsers/vc12dxmlCommon.h
```diff
@@ -0,0 +1,85 @@
...
+  udResult BeginBuild(udJSON const *pNode);
+  udResult Build(udJSON const *pNode, vc12DXML_ProjectGlobals &globals) override;
+  void Print(FILE *pOut, int indent = 0) const override;
+  void AddToProject(vcState *pProgramState, udProjectNode *pParent) override;
+
+private:
+
+  char *m_pName;
+  vc12DXML_ProjectGlobals m_globals; // Every 12dxml file has a set of globals that can change while parsing
+  std::vector<vc12DXML_Item *> m_models;
```
samsinsane, line 82: Same as above?

samsinsane (changes requested):

src/parsers/vc12DXMLLoad.cpp
```diff
@@ -0,0 +1,423 @@
...
+  for (size_t i = 0; i < udLengthOf(Colours); ++i)
+  {
+    if (udStrcmp(pStr, Colours[i].pStr) == 0)
+      return Colours[i].val;
+  }
+
+  return 0xFFFF00FF;
+}
+
+static size_t vc12DXML_StrSplit(char *pStr, const char *pDelims, char **ppSubstrings, size_t subStrCount)
```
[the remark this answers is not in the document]
int-Frank: Yes I was using `udStrTokenSplit`, using a space char as a delim, but for cases where there were mutiple spaces between items, I was getting multiple/empty tokens. Actually this was only happening with a test file I created which I deliberately put multiple spaces between vector elements. As I write this, I'm not sure that was a fair test; why not other characters as well as extra spaces? I'm thinking use `udStrTokenSplit` and if it fails, chalk it up to a malformed item.

src/parsers/vc12DXMLLoad.cpp
```diff
@@ -0,0 +1,423 @@
...
+  pItem = &pNode->Get("data_3d");
+  if (!pItem->IsVoid())
+    vc12DXML_Ammend_data_3d(&pItem->Get("p"), m_points, globals);
+
+  result = udR_Success;
+epilogue:
+  return result;
+}
+
+void vc12DXML_SuperString::Print(FILE *pOut, int indent) const
```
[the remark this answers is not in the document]
int-Frank: Probably not. I used `FILE` as I was passing in `stdout`; not sure I could do that with a `udFile`. It is debug code. As far as keeping this code, it may be useful as more features are added, but am happy to remove.

src/parsers/vc12dxmlCommon.h
int-Frank: I need a contiguous array to when I pass the points to `vcProject_UpdateNodeGeometryFromCartesian()`

src/parsers/vc12dxmlCommon.h
int-Frank: Actually this could be a `udChunkedArray` but I figured since I'm already using `std::vector` I might as well use that.

src/parsers/vc12DXMLLoad.cpp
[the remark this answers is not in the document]
samsinsane: Yeah I don't think `udFile` could do that. The main reason to remove it is that it's exposed in the header as it's a member function, I'd prefer that `FILE` wasn't something that was just available, even if it's an internal header.

src/parsers/vc12dxmlCommon.h
samsinsane: Unless you know the number of points up front (in which case, just use `udAllocType`), the `std::vector` could grow a number of times which could be very slow if it needs to copy the data when it grows - which could happen every time it grows. A `udChunkedArray` may be faster to add in to (and use less memory), and then output to `udDouble3*` using the `ToArray` member function (and then free that pointer).

You don't need to change this, just letting you know how you can do it with a chunked array.

int-Frank: Actually I might at least test this. It might speed up loading, which for some files already is taking a couple of seconds.

int-Frank: Both perform about the same. It seems using `std::vector` performs slightly faster then `udChunkedArray` for the larger of the test files.

samsinsane (approved):

PR #1063, opened 2020-09-16
Awu/shader files

asunawu:
To support .shp files. Non-geolocated shape files work well, however geo-located shape files need to be fix, the udGeoZone zone was not created correctly. Tested files include string shp file, tin shp file, mine shp file and some files from australia-latest-free.shp. After a .shp file is imported or dragged into our client, a new scene will be created the same as the way how vcUDP works. 

image
image


src/vcSHP.cpp
```diff
@@ -0,0 +1,741 @@
...
+  int32_t leftBytes = 0;
+
+  vcSHP_Record record;
+  UD_ERROR_CHECK(udFile_Read(pFile, id, 4 * sizeof(uint8_t)));
+  record.id = (id[0] << 24) | (id[1] << 16) | (id[2] << 8) | id[3];
+  offset += 4;
+
+  UD_ERROR_CHECK(udFile_Read(pFile, length, 4 * sizeof(uint8_t)));
+  record.length = (length[0] << 24) | (length[1] << 16) | (length[2] << 8) | length[3];
+  record.length *= 2;
```
samsinsane, line 289: Might be worth just putting a comment here to indicate that the file stores the length in shorts rather than bytes. Without any comment, this appears to be wrong.

src/vcSHP.cpp
```diff
@@ -0,0 +1,741 @@
...
+static uint32_t UnpackDouble(bool localBigEndian, uint32_t readPos, double *dest, uint8_t *src)
+{
+  if (localBigEndian)
+    SwapByteOrder(8, src + readPos);
+
+  memcpy(dest, src + readPos, 8);
+  return readPos + 8;
+}
+
+static uint32_t UnpackInt32(bool localBigEndian, uint32_t readPos, int32_t *dest, uint8_t *src)
```
samsinsane, line 263: Same as above, but just shift directly into `*dest`.

src/vcSHP.cpp
```diff
@@ -0,0 +1,741 @@
...
+  break;
+  default:
+  break;
+  }
+
+  pSHP->shpRecords.PushBack(record);
+  offset += readPosition;
+
+epilogue:
+  if (buffer)
```
samsinsane, line 538: `udFree` already checks this.

src/vcSHP.cpp
```diff
@@ -0,0 +1,741 @@
...
+  UD_ERROR_CHECK(udFile_Read(pFile, id, 4 * sizeof(uint8_t)));
+  record.id = (id[0] << 24) | (id[1] << 16) | (id[2] << 8) | id[3];
+  offset += 4;
+
+  UD_ERROR_CHECK(udFile_Read(pFile, length, 4 * sizeof(uint8_t)));
+  record.length = (length[0] << 24) | (length[1] << 16) | (length[2] << 8) | length[3];
+  record.length *= 2;
+  offset += 4;
+
+  buffer = udAllocType(uint8_t, record.length, udAF_Zero);
```
samsinsane, line 292: You should that this succeeds before using it. `UD_ERROR_NULL(buffer, udR_MemoryAllocationFailure);`

src/vcSHP.cpp
```diff
@@ -0,0 +1,741 @@
...
+  udResult result;
+
+  udFile *pFile = nullptr;
+  uint8_t *headerBuffer = nullptr;
+  uint32_t offset = 0;
+
+  vcSHP_Header &header = pSHP->shpHeader;
+  UD_ERROR_CHECK(udFile_Open(&pFile, pFilename, udFOF_Read));
+
+  headerBuffer = udAllocType(uint8_t, 100, udAF_Zero);
```
samsinsane, line 558: Check this succeeds.

src/vcSHP.cpp
```diff
@@ -0,0 +1,741 @@
...
+  UD_ERROR_CHECK(udFile_Read(pFile, headerBuffer, 100 * sizeof(uint8_t)));
+  // 00 00 27 0a
+  UD_ERROR_IF((headerBuffer[0] != 0 || headerBuffer[1] != 0 || headerBuffer[2] != 0x27 || headerBuffer[3] != 0x0a), udR_ParseError);
+
+  header.fileLength = (headerBuffer[24] << 24) | (headerBuffer[25] << 16) | (headerBuffer[26] << 8) | headerBuffer[27];
+  if (header.fileLength < UINT_MAX / 2)
```
samsinsane, line 564: You should simplify this to `header.fileLength = 2 * udMin(header.fileLength, UINT32_MAX / 2);`. You should also use `UINT32_MAX` because it can differ to `UINT_MAX`, since `int` only needs to be at least 16bit.

src/vcSHP.cpp
```diff
@@ -0,0 +1,741 @@
...
+epilogue:
+  udFile_Close(&pFile);
+
+  if (headerBuffer)
+    udFree(headerBuffer);
+
+  return result;
+}
+
+void vcSHP_LoadDbfFile(vcSHP *pSHP, const char *pFilename)
```
samsinsane, line 601: This function isn't necessary, you only need to call `vcDBF_Load` as it will allocate the `vcDBF` object. `vcDBF_Create` is intended for when you want to make your own DBF file, rather than use an existing one.

src/vcSHP.cpp
```diff
@@ -0,0 +1,741 @@
...
+  vcDBF_Load(&pSHP->pDBF, pFilename);
+}
+
+udResult vcSHP_LoadFileGroup(vcSHP *pSHP, const char *pFilename)
+{
+  if (pSHP == nullptr || pFilename == nullptr)
+    return udR_InvalidParameter_;
+
+  int i = 1;
+  if (*((uint8_t *)&i) == 1)
```
samsinsane, line 613: As mentioned above, this is unnecessary as you can just shift the bytes into their correct spots, as you've done in a number of places already.

src/vcSHP.cpp
```diff
@@ -0,0 +1,741 @@
...
+  else
+    pSHP->localBigEndian = true;
+
+  // load shp file
+  udResult result;
+  result = vcSHP_LoadShpFile(pSHP, pFilename);
+  if (result != udR_Success)
+  {
+    vcSHP_Release(pSHP);
+    return result;
```
samsinsane, line 624: This function should use the `UD_ERROR_*` macros instead of returning twice.

src/vcSHP.cpp
```diff
@@ -0,0 +1,741 @@
...
+  }
+
+  printf("shape file load successfully %s\n", pFilename);
```
samsinsane, line 627: This isn't necessary and should be removed.

src/vcSHP.cpp
```diff
@@ -0,0 +1,741 @@
...
+
+  char extraFileName[256] = "";
+  int32_t nameLen = (int32_t)strlen(pFilename) - 3;
+  memcpy(extraFileName, pFilename, nameLen);
+  memcpy(extraFileName + nameLen, "dbf", 3);
```
samsinsane, line 632: Consider using `udFilename` and using the `SetExtension` member function instead of this.

src/vcSHP.cpp
```diff
@@ -0,0 +1,741 @@
...
+  vcSHP_LoadDbfFile(pSHP, extraFileName);
+
+  memcpy(extraFileName + nameLen, "prj", 3);
```
samsinsane, line 635: `SetExtension` here again.

src/vcSHP.cpp
```diff
@@ -0,0 +1,741 @@
...
+}
+
+void vcUDP_AddModel(vcState *pProgramState, udProjectNode *pParentNode, vcSHP_Record &record, vcDBF_Record *pDBFRecord, int32_t stringIndex)
+{
+  char buffer[256] = {};
+  const char *nodeName = "";
+  if (pDBFRecord != nullptr && !pDBFRecord->deleted && stringIndex >= 0)
+  {
+    size_t len = strlen(pDBFRecord->pFields[stringIndex].pString);
+    memcpy(buffer, pDBFRecord->pFields[stringIndex].pString, len);
```
samsinsane, line 648: I'm pretty sure you can just use `udStrcpy` or `udStrncpy` instead of this, additionally, you should use `udStrlen` instead of `strlen`.

src/vcSHP.cpp
```diff
@@ -0,0 +1,741 @@
...
+    udStrStripWhiteSpace(buffer);
+    nodeName = udTempStr("%s", buffer);
```
samsinsane, line 650: This isn't necessary, just do `nodeName = buffer`.

src/vcSHP.cpp
```diff
@@ -0,0 +1,741 @@
...
+  }
+  else
+  {
+    nodeName = udTempStr("POI %d", record.id);
```
samsinsane, line 654: You could also use `udSprintf(buffer, "POI %d", record.id)` and get rid of `nodeName`.

src/vcSHP.cpp
```diff
@@ -0,0 +1,741 @@
...
+  memcpy(extraFileName + nameLen, "prj", 3);
+  pSHP->projectionLoad = udFile_Load(extraFileName, &pSHP->WKTString);
+
+  return result;
+}
+
+void vcUDP_AddModel(vcState *pProgramState, udProjectNode *pParentNode, vcSHP_Record &record, vcDBF_Record *pDBFRecord, int32_t stringIndex)
+{
+  char buffer[256] = {};
+  const char *nodeName = "";
```
samsinsane, line 644: This is a pointer and should be named correctly, `pNodeName`.

src/vcSHP.cpp
```diff
@@ -0,0 +1,741 @@
...
+  memcpy(extraFileName + nameLen, "dbf", 3);
+  vcSHP_LoadDbfFile(pSHP, extraFileName);
+
+  memcpy(extraFileName + nameLen, "prj", 3);
+  pSHP->projectionLoad = udFile_Load(extraFileName, &pSHP->WKTString);
+
+  return result;
+}
+
+void vcUDP_AddModel(vcState *pProgramState, udProjectNode *pParentNode, vcSHP_Record &record, vcDBF_Record *pDBFRecord, int32_t stringIndex)
```
samsinsane, line 641: `record` should either be `const vcSHP_Record &record` or `vcSHP_Record *pRecord`.

src/vcSHP.cpp
```diff
@@ -0,0 +1,741 @@
...
+    vcProject_UpdateNodeGeometryFromCartesian(&pProgramState->activeProject, pNode, pProgramState->geozone, udPGT_LineString, record.data.polyM.points, record.data.polyM.pointsCount);
+  break;
+  case vcSHP_Type::PolygonM:
+    vcProject_UpdateNodeGeometryFromCartesian(&pProgramState->activeProject, pNode, pProgramState->geozone, udPGT_Polygon, record.data.polyM.points, record.data.polyM.pointsCount);
+  break;
+  case vcSHP_Type::MultiPointM:
+    vcProject_UpdateNodeGeometryFromCartesian(&pProgramState->activeProject, pNode, pProgramState->geozone, udPGT_MultiPoint, record.data.multiPointM.points, record.data.multiPointM.pointsCount);
+  break;
+  default:
+    printf("invalid shape type %d\n", (int)record.shapeType);
```
samsinsane, line 700: This isn't necessary and should be removed.

src/vcSHP.cpp
```diff
@@ -0,0 +1,741 @@
...
+}
+
+udResult vcSHP_Load(vcState *pProgramState, const char *pFilename)
+{
+  udResult result;
+
+  vcSHP shp;
+  result = vcSHP_LoadFileGroup(&shp, pFilename);
+
+  printf("load file done. name: %s result: %d\n", pFilename, result);
```
samsinsane, line 713: This isn't necessary and should be removed.

src/vcSHP.cpp
```diff
@@ -0,0 +1,741 @@
...
+  vcSHP shp;
+  result = vcSHP_LoadFileGroup(&shp, pFilename);
+
+  printf("load file done. name: %s result: %d\n", pFilename, result);  
+
+  // TODO: create scene from .prj file, can't work correctly, need to be fixed.
+  vcProject_CreateBlankScene(pProgramState, "SHP Import", vcPSZ_NotGeolocated);
+  if (shp.projectionLoad == udR_Success)
+  {
+    printf("%s \n", shp.WKTString);
```
samsinsane, line 719: This isn't necessary and should be removed.

src/vcSHP.cpp
```diff
@@ -0,0 +1,741 @@
...
+  break;
+  }
+  
+}
+
+udResult vcSHP_Load(vcState *pProgramState, const char *pFilename)
+{
+  udResult result;
+
+  vcSHP shp;
```
samsinsane, line 710: You should zero this to avoid any issues if `vcSHP_LoadFileGroup` fails.

src/vcSHP.cpp
```diff
@@ -0,0 +1,741 @@
...
+  result = vcSHP_LoadFileGroup(&shp, pFilename);
```
samsinsane, line 711: This function should use the `UD_ERROR_*` macros.

src/vcSHP.cpp
```diff
@@ -0,0 +1,741 @@
...
+    printf("%s \n", shp.WKTString);
+    udGeoZone zone = {};
+    udGeoZone_SetFromWKT(&zone, shp.WKTString);
+    zone.latLongBoundMin = shp.shpHeader.MBRmin;
+    zone.latLongBoundMax = shp.shpHeader.MBRmax;
+
+    vcGIS_ChangeSpace(&pProgramState->geozone, zone);
+  }
+  
+  udProjectNode *pParentNode = pProgramState->sceneExplorer.clickedItem.pItem != nullptr ? pProgramState->sceneExplorer.clickedItem.pItem : pProgramState->activeProject.pRoot;
```
samsinsane, line 728: I'm not quite sure what this is attempting to do? If a user had a UDS selected, what would this do?

samsinsane (changes requested):

src/vcDBF.cpp
```diff
@@ -719,7 +730,7 @@ udResult vcDBF_RemoveField(vcDBF *pDBF, uint16_t fieldIndex)
   return result;
 }
 
-udResult vcDBF_GetRecord(vcDBF *pDBF, vcDBF_Record **ppRecord, uint32_t recordIndex)
+udResult vcDBF_GetRecord(vcDBF *pDBF, vcDBF_Record **ppRecord, uint16_t recordIndex)
```
samsinsane, line 733: You changed this from `uint32_t`, you should have reverted it back to `uint32_t`.

src/vcDBF.h
```diff
@@ -55,12 +55,14 @@ udResult vcDBF_GetFieldIndex(vcDBF *pDBF, const char *pFieldName, uint16_t *pInd
 uint16_t vcDBF_GetFieldCount(vcDBF *pDBF);
 uint32_t vcDBF_GetRecordCount(vcDBF *pDBF);
 
+uint16_t vcDBF_GetStringFieldIndex(vcDBF *pDBF);
```
samsinsane, line 58: I still think this should be replaced with proper functions to get field information. We know the field count via `vcDBF_GetFieldCount`, so we should be able to query the field name, type and length based on the index.

samsinsane (changes requested):
The `imgui_markdown` and `udcore` submodules have gone backwards.

src/vcSHP.cpp
```diff
@@ -0,0 +1,738 @@
...
+udResult vcSHP_Load(vcState *pProgramState, const char *pFilename)
+{
+  vcSHP shp = {};
+  udProjectNode *pParentNode = nullptr;
+  uint16_t stringIndex = 0;
+  vcDBF_Record *pDBFRecord = nullptr;
+  udGeoZone zone = {};
+
+  udResult result = vcSHP_LoadFileGroup(&shp, pFilename);
+  UD_ERROR_CHECK(result);
```
samsinsane, line 715: `UD_ERROR_HANDLE()` will do the equivalent of this, however, you should call `vcSHP_LoadFileGroup` in this macro instead.

src/vcSHP.cpp
samsinsane: You changed this to call `UD_ERROR_CHECK(result)`, but you should have changed it to `UD_ERROR_CHECK(vcSHP_LoadFileGroup(&shp, pFilename))`, alternatively but less consistent, use `UD_ERROR_HANDLE()` instead of `UD_ERROR_CHECK(result)`.

src/vcDBF.cpp
asunawu: fixed.

src/vcDBF.h
asunawu: fixed.

src/vcSHP.cpp
asunawu: fixed.

src/vcSHP.cpp
asunawu: fixed.

src/vcDBF.cpp
```diff
@@ -719,7 +730,7 @@ udResult vcDBF_RemoveField(vcDBF *pDBF, uint16_t fieldIndex)
   return result;
 }
 
-udResult vcDBF_GetRecord(vcDBF *pDBF, vcDBF_Record **ppRecord, uint32_t recordIndex)
+udResult vcDBF_GetRecord(vcDBF *pDBF, vcDBF_Record **ppRecord, size_t recordIndex)
```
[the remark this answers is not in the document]
asunawu: fixed.

src/vcSHP.cpp
```diff
@@ -0,0 +1,741 @@
...
+  break;
+  }
+}
+
+
+void vcSHP_Release(vcSHP *pSHP)
+{
+  // release .shp file
+  for (uint32_t i = 0; i < (uint32_t)pSHP->shpRecords.length; ++i)
+    vcSHP_ReleaseRecord(pSHP->shpRecords[i]);
```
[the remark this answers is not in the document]
asunawu: fixed.

src/vcSHP.cpp
```diff
@@ -0,0 +1,741 @@
...
+
+  vcSHP_Header shpHeader;
+  udChunkedArray<vcSHP_Record> shpRecords;
+  vcDBF *pDBF;
+
+  udResult projectionLoad;
+  char *WKTString = nullptr;
+};
+
+void vcSHP_ReleaseRecord(vcSHP_Record &record)
```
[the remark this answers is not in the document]
asunawu: fixed.

src/vcSHP.cpp
asunawu: fixed.

src/vcSHP.cpp
```diff
@@ -0,0 +1,741 @@
...
+{
+  for (int i = 0; i < length / 2; i++)
+  {
+    uint8_t temp = src[i];
+    src[i] = src[length - i - 1];
+    src[length - i - 1] = temp;
+  }
+}
+
+static uint32_t UnpackDouble(bool localBigEndian, uint32_t readPos, double *dest, uint8_t *src)
```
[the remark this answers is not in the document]
asunawu: fixed.

src/vcSHP.cpp
asunawu: fixed.

src/vcSHP.cpp
asunawu: fixed.

src/vcSHP.cpp
asunawu: fixed.

src/vcDBF.cpp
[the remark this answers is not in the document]
asunawu: fixed

src/vcDBF.cpp
```diff
@@ -673,6 +673,17 @@ uint32_t vcDBF_GetRecordCount(vcDBF *pDBF)
   return (uint32_t)pDBF->records.length;
 }
 
+int32_t vcDBF_GetStringFieldIndex(vcDBF *pDBF)
```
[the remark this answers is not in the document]
asunawu: Moved this function to vcSHP.cpp

src/vcSHP.cpp
```diff
@@ -0,0 +1,741 @@
+#include "vcSHP.h"
+
+#include "udMath.h"
+#include "udChunkedArray.h"
+#include "udFile.h"
+#include "vcDBF.h"
+#include "udStringUtil.h"
+
+enum class vcSHP_Type
```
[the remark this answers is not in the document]
asunawu: fixed.

src/vcSHP.cpp
asunawu: fixed.

src/vcSHP.cpp
asunawu: fixed.

src/vcSHP.cpp
asunawu: fixed.

src/vcSHP.cpp
asunawu: fixed.

src/vcSHP.cpp
asunawu: fixed.

src/vcSHP.cpp
asunawu: fixed.

src/vcSHP.cpp
asunawu: fixed.

src/vcSHP.cpp
asunawu: fixed.

src/vcSHP.cpp
asunawu: fixed.

src/vcSHP.cpp
asunawu: fixed.

src/vcSHP.cpp
asunawu: fixed.

src/vcSHP.cpp
asunawu: fixed.

src/vcSHP.cpp
asunawu: fixed.

src/vcSHP.cpp
asunawu: fixed.

src/vcSHP.cpp
asunawu: fixed.

src/vcSHP.cpp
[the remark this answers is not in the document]
samsinsane: You didn't change this to the snippet?

src/vcSHP.cpp
samsinsane: There doesn't seem to be a comment for this.

src/vcSHP.cpp
```diff
@@ -0,0 +1,740 @@
...
+
+  return result;
+}
+
+udResult vcSHP_LoadFileGroup(vcSHP *pSHP, const udFilename &fileName)
+{
+  if (pSHP == nullptr || !fileName.HasFilename())
+    return udR_InvalidParameter_;
+
+  udFilename dbfName = fileName;
```
samsinsane, line 623: I had intended for you to only change the internals of this function rather than passing a `udFilename` in, just using one internally. You don't need multiple `udFilename` objects, you can just modify the one file. Create the `udFilename`, load the shapefile, call `SetExtension`, load the DBF, call `SetExtension`, load the PRJ. Does `udFilename` work with this assignment operator, there isn't an explicit one but maybe it's converting to `const char *` and using that constructor.

samsinsane (changes requested):

src/vcSHP.cpp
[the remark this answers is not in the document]
asunawu: it was changed to 

  for (uint32_t i = 0; i < (uint32_t)pSHP->shpRecords.length; ++i)
    vcSHP_ReleaseRecord(&pSHP->shpRecords[i]);

as the next review requires. I have no idea to get the address by "for (vcSHPRecord &record : pSHP->shpRecords)" ?

src/vcSHP.cpp
asunawu: sorry, it was lost after a series of rollback operations, add it again.

src/vcSHP.cpp
asunawu: fixed.

src/vcSHP.cpp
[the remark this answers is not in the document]
samsinsane: If the next PR requires it then that's fine! I would have expected the above to just work? With the `vcSHP_ReleaseRecord` parameter changing to `&record`?

samsinsane (approved):

PR #1054, opened 2020-09-15
Flythrough Missing Translations

Jossos:
- Added 2 Missing Translation Strings
- Fixes [AB#2143](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/2143)

PR #1064, opened 2020-09-16
Fly through FPS options

Jossos:
- Added options for Export FPS for Flythroughs
- Fixes [AB#2147](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/2147)

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1005,14 +1005,6 @@ int main(int argc, char **args)
   programState.activeViewportIndex = 0;
   programState.pActiveViewport = &programState.pViewports[programState.activeViewportIndex];
 
-  programState.supportedVideoFPSs.Init(16);
-  programState.supportedVideoFPSs.PushBack(12);
-  programState.supportedVideoFPSs.PushBack(24);
-  programState.supportedVideoFPSs.PushBack(30);
-  programState.supportedVideoFPSs.PushBack(60);
-  programState.supportedVideoFPSs.PushBack(120);
-  programState.videoFPSIndex = 2;
-
   for (int i = 1; i < argc; ++i)
   {
 #if UDPLATFORM_OSX
@@ -1182,8 +1174,6 @@ epilogue:
 
   vcProject_Deinit(&programState, &programState.activeProject);
   vcTexture_Destroy(&programState.image.pImage);
-
-  programState.supportedVideoFPSs.Deinit();
 
   for (ImWchar *pGlyphs : programState.requiredGlyphs)
     udFree(pGlyphs);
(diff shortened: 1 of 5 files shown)


src/main.cpp
```diff
@@ -1005,6 +1005,14 @@ int main(int argc, char **args)
   programState.activeViewportIndex = 0;
   programState.pActiveViewport = &programState.pViewports[programState.activeViewportIndex];
 
+  programState.supportedVideoFPSs.Init(16);
```
samsinsane, line 1008: Use a static array at the top of the `vcFlythrough.cpp` file, it's what we do with similar UI elements. What you have done here is incredibly unnecessary.

src/scene/vcFlythrough.cpp
```diff
@@ -254,7 +254,21 @@ void vcFlythrough::HandleSceneEmbeddedUI(vcState *pProgramState)
 
   if (m_state != vcFTS_Exporting)
   {
-    if (ImGui::Button(vcString::Get("flythroughExport")))
+    if (ImGui::BeginCombo(vcString::Get("flythroughFPS"), udTempStr_CommaInt(pProgramState->supportedVideoFPSs[pProgramState->videoFPSIndex])))
```
samsinsane, line 257: You should just use `udTempStr` instead as `udTempStr_CommaInt` does more work for no value - it is incredibly unlikely that a video will have more than 1000 frames per second.

src/scene/vcFlythrough.cpp
```diff
@@ -254,7 +254,21 @@ void vcFlythrough::HandleSceneEmbeddedUI(vcState *pProgramState)
...
+    {
+      for (size_t i = 0; i < pProgramState->supportedVideoFPSs.length; ++i)
+      {
+        if (ImGui::Selectable(udTempStr_CommaInt(pProgramState->supportedVideoFPSs[i])))
```
samsinsane, line 261: Same as above, just use `udTempStr`.

src/scene/vcFlythrough.cpp
```diff
@@ -254,7 +254,21 @@ void vcFlythrough::HandleSceneEmbeddedUI(vcState *pProgramState)
...
+        {
+          pProgramState->videoFPSIndex = (int)i;
+          m_exportInfo.frameDelta = (1.0 / (double)pProgramState->supportedVideoFPSs[pProgramState->videoFPSIndex]);
+        }
+      }
+
+      ImGui::EndCombo();
+    }
+
+    if (ImGui::Button("export"))
```
samsinsane, line 271: This reverts a translation fix you made two days ago, #1054.

samsinsane (changes requested):

samsinsane (approved):

PR #1065, opened 2020-09-17
Added some functionality to deal with endianness

int-Frank:
While working on unicode conversion I realised I will probably need some endian stuff, so I gave it its own PR

### DIFF
--- a/src/vcEndian.cpp
+++ b/src/vcEndian.cpp
@@ -1,0 +1,43 @@
+
+#include "vcEndian.h"
+
+vcEndian vcEndian_Endianness()
+{
+  union
+  {
+    uint32_t value;
+    uint8_t data[sizeof(uint32_t)];
+  } number;
+
+  number.data[0] = 0x00;
+  number.data[1] = 0x01;
+  number.data[2] = 0x02;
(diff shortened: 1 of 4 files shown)


samsinsane (changes requested):
It's not necessary to know what Endian the current machines is. Most of the systems that read binary data will fail on a BE machine, it's safe to assume the machine is LE since it will be a huge effort to get to BE. Additionally, the only time you want to convert between Endian is when you're reading/writing binary data to disk or network, and you should already know which one by then. An Endian solution would be to offer `template<typename T, bool LE = true> udResult vcBinaryHelper_ReadData(udFile *pFile, T *pData);` and then you would call it like `vcBinaryHelper_ReadData(pFile, &item);` when you know you're reading LE (pretty much all the time) and `vcBinaryHelper_ReadData<false>(pFile, &item);` when you know you're reading BE (pretty much never). In this helper, you use bit shifting to put the bytes in the correct order as bit shifting is Endian agnostic.

Overall, I'm not sure something like Endian support should be done in isolation like this and should be discussed in detail before putting up a generic solution. Where you intended to use this, just handle LE machine with LE/BE data.

PR #1067, opened 2020-09-17
Added support for more colours and different xml structures

int-Frank:
Needed to support all the test files I have

### DIFF
--- a/src/parsers/vc12DXMLLoad.cpp
+++ b/src/parsers/vc12DXMLLoad.cpp
@@ -19,8 +19,11 @@
 #define LOG_WARNING(...) {}
 #define LOG_INFO(...) {}
 
-static bool vc12DXML_IsGreyColour(const char *pStr, uint32_t &out)
+static bool vc12DXML_IsGreyColour(const char *pStr, uint32_t *pOut)
 {
+  if (pStr == nullptr || pOut == nullptr)
+    return false;
+
   bool success = false;
 
   if (udStrBeginsWithi(pStr, "grey "))
@@ -28,10 +31,10 @@ static bool vc12DXML_IsGreyColour(const char *pStr, uint32_t &out)
     uint32_t scale = udStrAtou(pStr + 5); // Skip "grey "
     if (scale > 255)
       scale = 255;
-    out = 0xFF000000;
-    out |= (scale << 16);
-    out |= (scale << 8);
-    out |= (scale << 0);
+    *pOut = 0xFF000000;
+    *pOut |= (scale << 16);
+    *pOut |= (scale << 8);
+    *pOut |= (scale << 0);
     success = true;
   }
   return success;
@@ -80,7 +83,7 @@ static uint32_t vc12DXML_ToColour(const char *pStr)
   }
 
   uint32_t grey = 0;
-  if (vc12DXML_IsGreyColour(pStr, grey))
+  if (vc12DXML_IsGreyColour(pStr, &grey))
     return grey;
 
   return 0xFFFF9D00;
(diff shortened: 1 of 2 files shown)


src/parsers/vc12DXMLLoad.cpp
```diff
@@ -19,7 +19,29 @@
...
+static bool vc12DXML_IsGreyColour(const char *pStr, uint32_t &out)
+{
+  char *pTemp = udStrdup(pStr);
+  char *tokens[2] = {};
+  int nTokens = udStrTokenSplit(pTemp, " ", tokens, 2);
+  bool success = false;
+
+  if (nTokens == 2 && udStrcmpi(tokens[0], "grey") == 0)
+  {
+    uint32_t scale = udStrAtou(tokens[1]);
```
samsinsane, line 31: You can avoid the `udStrdup` and `udStrTokenSplit` above by using `udStrBeginsWithi(pStr, "grey ")` and then here `udStrAtou(pStr + 5); // Skip "grey "`. This probably doesn't need to change, but I think it can be simplified.

src/parsers/vc12DXMLLoad.cpp
```diff
@@ -269,31 +311,47 @@ void vc12DXML_Project::SetName(const char *pName)
   m_pName = udStrdup(pName);
 }
 
-udResult vc12DXML_Project::BeginBuild(udJSON const *pNode)
-{
-  return Build(pNode, m_globals);
-}
-
-udResult vc12DXML_Project::Build(udJSON const *pNode, vc12DXML_ProjectGlobals &globals)
+udResult vc12DXML_Project::AddProject(udJSON const *pNode, vc12DXML_ProjectGlobals &globals)
```
samsinsane, line 314: `globals` should be `const &` or `*pGlobals`, you either don't modify it, or you make it clear in the calling code that the value is being modified. `AddProject(..., globals);` vs `AddProject(..., &globals);`, in the second example it's clear that we expect `globals` to be modified, but we don't know in the first example if it's being modified or not - if we avoid passing by reference (unless it's const) then we do know.

src/parsers/vc12DXMLLoad.cpp
```diff
@@ -316,6 +374,63 @@ udResult vc12DXML_Project::Build(udJSON const *pNode, vc12DXML_ProjectGlobals &g
   return result;
 }
 
+bool vc12DXML_Project::TryLoad(udJSON const *pNode, vc12DXML_ProjectGlobals &globals)
```
samsinsane, line 377: Same as above.

src/parsers/vc12DXMLLoad.cpp
```diff
@@ -316,6 +374,63 @@ udResult vc12DXML_Project::Build(udJSON const *pNode, vc12DXML_ProjectGlobals &g
...
+    {
+      if (AddProject(pNode, globals) != udR_Success)
+        LOG_WARNING("Failed to load project.");
+    }
+  }
+
+  return result;
+}
+
+udResult vc12DXML_Project::Build(udJSON const *pNode, vc12DXML_ProjectGlobals &globals)
```
samsinsane, line 405: Same as above.

src/parsers/vc12dxmlCommon.h
```diff
@@ -72,6 +72,10 @@ class vc12DXML_Project : public vc12DXML_Item
 
 private:
 
+  udResult AddProject(udJSON const *pNode, vc12DXML_ProjectGlobals &globals);
+  bool TryLoad(udJSON const *pNode, vc12DXML_ProjectGlobals &globals);
+
+private:
```
samsinsane, line 78: It's already `private`, only use the visibility qualifiers once per class.

samsinsane (changes requested):

samsinsane (dismissed):

samsinsane (dismissed):

samsinsane (approved):

PR #1068, opened 2020-09-17
Added Version 1.2 release notes

peterAdams1014:


builds/releasenotes.md
```diff
@@ -1,5 +1,18 @@
 udStream Version History
 
+Version 1.2
+  - Added helper when using Dropbox links
+  - Added markdown support
+  - Added scene explorer icons
+  - Added basic fly through recording
+  - Added 31700 and 19920 EPSG codes
+  - (Mac) Fixed an issue with fonts not loading
```
samsinsane, line 9: This should be `macOS` not `Mac`, we should aim for consistency with the current name that Apple uses.

samsinsane (changes requested):

samsinsane (approved):

PR #1071, opened 2020-09-18
Checks if deleted node is still in use

int-Frank:
Fixes [AB#2178](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/2178)

### DIFF
--- a/src/scene/vcFolder.cpp
+++ b/src/scene/vcFolder.cpp
@@ -356,6 +356,9 @@ void vcFolder::HandleSceneExplorerUI(vcState *pProgramState, size_t *pItemID)
 
         if (ImGui::Selectable(vcString::Get("sceneExplorerRemoveItem")))
         {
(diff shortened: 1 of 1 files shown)


src/scene/vcFolder.cpp
```diff
@@ -356,6 +356,9 @@ void vcFolder::HandleSceneExplorerUI(vcState *pProgramState, size_t *pItemID)
 
         if (ImGui::Selectable(vcString::Get("sceneExplorerRemoveItem")))
         {
+          if ((vcSceneItem*)(pNode->pUserData) == pProgramState->pActiveViewport->cameraInput.pAttachedToSceneItem)
```
samsinsane, line 359: You don't need to cast this to compare the pointers?

samsinsane (approved):

PR #1082, opened 2020-09-21
Added R32F texture format

peterAdams1014:


samsinsane (approved):
We'll need to ensure that the Metal build works for whatever this is for.

PR #1089, opened 2020-09-21
Fixed Label's text missing

jproust:
Use of AddRectFilled(...) instead of AddQuadFilled(...) to simplify a bit and calculate only 2 points instead of 4.
Fixes [AB#2191](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/2191)


asunawu (approved):
Seems ok to me.

PR #1096, opened 2020-09-22
Added a prompt if folder contains same Flythrough images

jproust:
Expanded vcModals_OverwriteExistingFile to specify a string for the message displayed
Also fixed the saved texture to use the extension specified.
Fixes [AB#2225](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/2225)
Fixes [AB#2213](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/2213)

### DIFF
--- a/src/scene/vcFlythrough.cpp
+++ b/src/scene/vcFlythrough.cpp
@@ -12,6 +12,7 @@
 #include "imgui.h"
 #include "imgui_ex/vcImGuiSimpleWidgets.h"
 #include "imgui_internal.h"
+#include "vcModals.h"
 
 static const int vcFlythroughExportFPS[] = { 12, 24, 30, 60, 120 };
 const char *vcFlythroughExportFormats[] =
@@ -61,7 +62,7 @@ void vcFlythrough::AddToScene(vcState *pProgramState, vcRenderData *pRenderData)
     {
       if (m_exportInfo.currentFrame >= 0)
       {
-        if (vcTexture_SaveImage(pProgramState->screenshot.pImage, vcRender_GetSceneFramebuffer(pProgramState->pActiveViewport->pRenderContext), udTempStr("%s/%05d.png", m_exportPath, m_exportInfo.currentFrame)) != udR_Success)
+        if (vcTexture_SaveImage(pProgramState->screenshot.pImage, vcRender_GetSceneFramebuffer(pProgramState->pActiveViewport->pRenderContext), udTempStr("%s/%05d.%s", m_exportPath, m_exportInfo.currentFrame, vcFlythroughExportFormats[m_selectedExportFormatIndex])) != udR_Success)
         {
           m_state = vcFTS_None;
           pProgramState->exportVideo = false;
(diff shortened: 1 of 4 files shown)


src/vcModals.cpp
```diff
@@ -54,30 +54,29 @@ void vcModals_DrawLoggedOut(vcState *pProgramState)
 }
 
 // Presents user with a message if the specified file exists, then returns false if user declines to overwrite the file
-bool vcModals_OverwriteExistingFile(vcState *pProgramState, const char *pFilename)
+bool vcModals_OverwriteExistingFile(vcState* pProgramState, const char* pFilename, const char* pFileExistingMsg)
```
samsinsane, line 57: `*` go with the variable, I assume your VS is misconfigured?

src/vcModals.cpp
```diff
@@ -54,30 +54,29 @@ void vcModals_DrawLoggedOut(vcState *pProgramState)
...
-  const char *pFileExistsMsg = nullptr;
   if (udFileExists(pFilename) == udR_Success)
   {
     const SDL_MessageBoxButtonData buttons[] = {
       { SDL_MESSAGEBOX_BUTTON_ESCAPEKEY_DEFAULT, 0, vcString::Get("popupConfirmNo") },
       { SDL_MESSAGEBOX_BUTTON_RETURNKEY_DEFAULT, 1, vcString::Get("popupConfirmYes") },
     };
-    pFileExistsMsg = vcStringFormat(vcString::Get("convertFileExistsMessage"), pFilename);
+    if (pFileExistingMsg == nullptr)
+      pFileExistingMsg = vcStringFormat(vcString::Get("convertFileExistsMessage"), pFilename);
```
samsinsane, line 67: Isn't this a generic message?

src/scene/vcFlythrough.cpp
```diff
@@ -321,8 +322,13 @@ void vcFlythrough::HandleSceneEmbeddedUI(vcState *pProgramState)
       ImGui::EndCombo();
     }
 
-    vcIGSW_FilePicker(pProgramState, vcString::Get("flythroughExportPath"), m_exportPath, udLengthOf(m_exportPath), nullptr, 0, vcFDT_SelectDirectory, nullptr);
-    
+    // Only testing if a first frame has already been exported in this folder
+    auto testIfExistLambda = [pProgramState, this]
+        {if (!vcModals_OverwriteExistingFile(pProgramState, udTempStr("%s/%05d.%s", m_exportPath, 0, vcFlythroughExportFormats[m_selectedExportFormatIndex]), vcString::Get("flythroughExportAlreadyExists")))
```
samsinsane, line 327: This should be formatted correctly, also this should be performed on export not on file pick. Exporting a second time won't prompt this.

samsinsane (changes requested):

src/vcModals.cpp
jproust: it is, if there is no specific string given.

samsinsane: Sorry, I was not clear at all here, must have been distracted when I wrote it.

What I should have written:
> If this is a generic message, is there any need to have an optional message passed in?

samsinsane: As discussed on Teams, this is necessary since vcFlythrough is using a file to indicate that a folder is going to be overwritten rather than a file. Resolving discussion.

src/scene/vcFlythrough.cpp
jproust: Moved to the Export button

src/vcModals.cpp
jproust: I guess so, FIxed

samsinsane (dismissed):

samsinsane (approved):

PR #1098, opened 2020-09-22
Checking for inf input

int-Frank:
Fixes [AB#2221](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/2221)

### DIFF
--- a/src/scene/vcFlythrough.cpp
+++ b/src/scene/vcFlythrough.cpp
@@ -174,30 +174,38 @@ void vcFlythrough::HandleSceneExplorerUI(vcState *pProgramState, size_t *pItemID
     udDouble3 positionTemp = m_flightPoints[i].m_CameraPosition;
     udDouble2 rotationTemp = m_flightPoints[i].m_CameraHeadingPitch;
 
-    ImGui::InputDouble("PX", &positionTemp.x);
-
-    ImGui::NextColumn();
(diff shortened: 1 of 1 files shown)


src/scene/vcFlythrough.cpp
```diff
@@ -171,19 +171,34 @@ void vcFlythrough::HandleSceneExplorerUI(vcState *pProgramState, size_t *pItemID
...
-    ImGui::InputDouble("RH", &m_flightPoints[i].m_CameraHeadingPitch.x);
+    ImGui::InputDouble("RH", &rotationTemp.x);
 
     ImGui::NextColumn();
-    ImGui::InputDouble("RP", &m_flightPoints[i].m_CameraHeadingPitch.y);
+    ImGui::InputDouble("RP", &rotationTemp.y);
+
+    for (int j = 0; j < 3; ++j)
+    {
+      if (!isinf(positionTemp[j]))
```
int-Frank, line 193: This should be `if (isfinite(...` as recommended by @samsinsane

samsinsane (changes requested):
Can you change it to only perform the loops at the end when the values have changed. There could be hundreds or thousands of points unnecessarily checking the validity of the float values every frame. I'm not sure how expensive the validation functions are, but I've heard they're not cheap, whatever that means.

samsinsane (approved):

PR #1103, opened 2020-09-22
Flythrough First Frame Constrain Max Time

Jossos:
- Fixed bug in calculating First frame max time
- Fixes [AB#2223](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/2223)

### DIFF
--- a/src/scene/vcFlythrough.cpp
+++ b/src/scene/vcFlythrough.cpp
@@ -166,7 +166,7 @@ void vcFlythrough::HandleSceneExplorerUI(vcState *pProgramState, size_t *pItemID
     ImGui::PushID(udTempStr("ftpt_%zu", (*pItemID)));
 
     double minTime = (i == 0 ? 0 : m_flightPoints[i-1].time);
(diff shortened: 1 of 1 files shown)


samsinsane (changes requested):
I don't doubt that this fixes the reported issue, however, the start time of the first frame should always be 0.

peterAdams1014 (dismissed):

Jossos:
> I don't doubt that this fixes the reported issue, however, the start time of the first frame should always be 0.

I'm not sure if this is necessarily true. I'll discuss with Paul, but it's out of scope of this PR

samsinsane:
The original code was setting the min and max of the first frame time to 0, this would indicate to me that the intention is for the first frame time to always be 0. How exactly is that out of scope for this PR?

Jossos:
> The original code was setting the min and max of the first frame time to 0, this would indicate to me that the intention is for the first frame time to always be 0. How exactly is that out of scope for this PR?

so the min/max are for the drag scaler UI. The min is always the previous frames time, and the max is the next frames time. since 0th frame has no previous frame, it's just set to 0. This doesn't show any intention for the time of the first frame to always be 0, otherwise it would just set it to 0.

samsinsane (approved):

PR #1104, opened 2020-09-22
Adding deltaTime to time length when inserting new flight points

int-Frank:
Fixes [AB#2227](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/2227)

### DIFF
--- a/src/scene/vcFlythrough.cpp
+++ b/src/scene/vcFlythrough.cpp
@@ -114,7 +114,7 @@ void vcFlythrough::AddToScene(vcState *pProgramState, vcRenderData *pRenderData)
       for (size_t i = offset + 1; i < m_flightPoints.length; ++i)
         m_flightPoints[i].time += pProgramState->deltaTime;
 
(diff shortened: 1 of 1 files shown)


src/scene/vcFlythrough.cpp
```diff
@@ -113,6 +113,8 @@ void vcFlythrough::AddToScene(vcState *pProgramState, vcRenderData *pRenderData)
 
       for (size_t i = offset + 1; i < m_flightPoints.length; ++i)
         m_flightPoints[i].time += pProgramState->deltaTime;
+
+      m_timeLength += pProgramState->deltaTime;
```
samsinsane, line 117: Probably best to set this to `m_flightPoints[m_flightPoints.length - 1].time` instead of relying on the FPU to produce the same result.

samsinsane (changes requested):

samsinsane (approved):

PR #1127, opened 2020-09-24
Added a 'Cancel' Member to SceneTools

int-Frank:
Fixes [AB#2239](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/2239)

### DIFF
--- a/src/scene/tools/vcSceneTool.h
+++ b/src/scene/tools/vcSceneTool.h
@@ -12,6 +12,7 @@ public:
   virtual void SceneUI(vcState *pProgramState);
   virtual void HandlePicking(vcState *pProgramState, vcRenderData &renderData, const vcRenderPickResult &pickResult);
   virtual void PreviewPicking(vcState *pProgramState, vcRenderData &renderData, const vcRenderPickResult &pickResult);
+  virtual void Cancel(vcState *pProgramState); // Essentially a 'cancel' command has been issued (esc key was pressed)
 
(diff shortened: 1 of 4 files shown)


src/scene/tools/vcSceneTool.h
```diff
@@ -12,12 +12,14 @@ class vcSceneTool
   virtual void SceneUI(vcState *pProgramState);
   virtual void HandlePicking(vcState *pProgramState, vcRenderData &renderData, const vcRenderPickResult &pickResult);
   virtual void PreviewPicking(vcState *pProgramState, vcRenderData &renderData, const vcRenderPickResult &pickResult);
+  virtual void Cancel(vcState *pProgramState); // Essentially a 'cancel' command has been issued (esc key was pressed)
```
samsinsane, line 15: This comment seems like it's trying to avoid confusion since "Cancel" could be interpreted as either an action or a reaction, and in this case we're wanting a reaction. Assuming this is correct, perhaps calling this function `OnCancel` would be better? This is generally how UI naming works, for example, a button has a `Click` and `OnClick` function.

samsinsane (changes requested):

src/scene/tools/vcSceneTool.h
OsiJr: `OnCancel` is more consistent- `OnNodeUpdate` etc

samsinsane (approved):

PR #1131, opened 2020-09-24
Map mode now available for both viewport

jproust:
Map mode can now be set to both viewport at the same time or left or right.
Fixes [AB#2264](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/2264)

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1534,8 +1534,10 @@ void vcRenderSceneUI(vcState *pProgramState, const ImVec2 &windowPos, const ImVe
         for (int v = 0; v < vcMaxViewportCount; v++)
         {
           if (v < pProgramState->settings.activeViewportCount)
-            if (ImGui::SliderFloat(udTempStr("%s %d",vcString::Get("sceneCameraMoveSpeed"), v), &(pProgramState->settings.camera.moveSpeed[v]), vcSL_CameraMinMoveSpeed, vcSL_CameraMaxMoveSpeed, "%.3f m/s", ImGuiSliderFlags_Logarithmic))
+          {
+            if (ImGui::SliderFloat(udTempStr("%s %d", vcString::Get("sceneCameraMoveSpeed"), v), &(pProgramState->settings.camera.moveSpeed[v]), vcSL_CameraMinMoveSpeed, vcSL_CameraMaxMoveSpeed, "%.3f m/s", ImGuiSliderFlags_Logarithmic))
               pProgramState->settings.camera.moveSpeed[v] = udClamp(pProgramState->settings.camera.moveSpeed[v], vcSL_CameraMinMoveSpeed, vcSL_CameraMaxMoveSpeed);
+          }
         }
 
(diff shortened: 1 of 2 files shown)


src/main.cpp
```diff
@@ -1531,8 +1531,12 @@ void vcRenderSceneUI(vcState *pProgramState, const ImVec2 &windowPos, const ImVe
...
           pProgramState->pActiveViewport->camera.headingPitch = UD_DEG2RAD(headingPitch);
 
-        if (ImGui::SliderFloat(vcString::Get("sceneCameraMoveSpeed"), &(pProgramState->settings.camera.moveSpeed[pProgramState->activeViewportIndex]), vcSL_CameraMinMoveSpeed, vcSL_CameraMaxMoveSpeed, "%.3f m/s", ImGuiSliderFlags_Logarithmic))
-          pProgramState->settings.camera.moveSpeed[pProgramState->activeViewportIndex] = udClamp(pProgramState->settings.camera.moveSpeed[pProgramState->activeViewportIndex], vcSL_CameraMinMoveSpeed, vcSL_CameraMaxMoveSpeed);
+        for (int v = 0; v < vcMaxViewportCount; v++)
+        {
+          if (v < pProgramState->settings.activeViewportCount)
+            if (ImGui::SliderFloat(udTempStr("%s %d",vcString::Get("sceneCameraMoveSpeed"), v), &(pProgramState->settings.camera.moveSpeed[v]), vcSL_CameraMinMoveSpeed, vcSL_CameraMaxMoveSpeed, "%.3f m/s", ImGuiSliderFlags_Logarithmic))
+              pProgramState->settings.camera.moveSpeed[v] = udClamp(pProgramState->settings.camera.moveSpeed[v], vcSL_CameraMinMoveSpeed, vcSL_CameraMaxMoveSpeed);
```
OsiJr, line 1538: This needs to be in braces

src/main.cpp
```diff
@@ -1640,7 +1644,11 @@ void vcRenderSceneUI(vcState *pProgramState, const ImVec2 &windowPos, const ImVe
 
       // TODO: Putting this here for now
       if (pProgramState->settings.activeViewportCount > 1)
+      {
+        ImGui::Checkbox(udTempStr("%s##firstViewportMapMode", vcString::Get("orthographicCameraViewport")), &pProgramState->settings.camera.mapMode[0]);
+        ImGui::SameLine();
         ImGui::Checkbox(udTempStr("%s##secondViewportMapMode", vcString::Get("orthographicCameraViewport")), &pProgramState->settings.camera.mapMode[1]);
```
OsiJr, line 1650: This is ugly UX; I don't have an alternative at the moment but this will be confusing

src/main.cpp
jproust: Fixed

src/main.cpp
jproust: I have added 1 & 2 at the end of the checkbox label.
This is just temporary, currently working on adding visualisations effect per viewport. Once it is done, will move the map mode to the tab of each viewport in the visualisation menu.

PR #1135, opened 2020-09-24
Fixed some minor markdown issues

OsiJr:
And viewpoint descriptions use Markdown to display

### DIFF
--- a/src/imgui_ex/vcImGuiSimpleWidgets.cpp
+++ b/src/imgui_ex/vcImGuiSimpleWidgets.cpp
@@ -20,6 +20,11 @@ struct vcIGSWResizeContainer
   size_t *pBufferSize;
 };
 
+struct vcIGSWMarkdownContext
+{
+  ImVec2 areaSize;
+};
+
 int vcIGSW_ResizeString(ImGuiInputTextCallbackData *pData)
 {
   if (pData->EventFlag == ImGuiInputTextFlags_CallbackResize)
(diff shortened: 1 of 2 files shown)


src/imgui_ex/vcImGuiSimpleWidgets.cpp
```diff
@@ -318,13 +329,30 @@ inline ImGui::MarkdownImageData ImageCallback(ImGui::MarkdownLinkCallbackData da
...
 
+void vcIGSW_MarkDownToolTip(ImGui::MarkdownTooltipCallbackData data)
+{
+  if (data.linkData.isImage)
+  {
+    ImGui::SetTooltip("%.*s", data.linkData.textLength, data.linkData.text);
+  }
+  else
+  {
+    ImGui::SetTooltip("%s Open in browser\n%.*s", data.linkIcon, data.linkData.linkLength, data.linkData.link);
```
samsinsane, line 340: This isn't translated?

samsinsane (changes requested):

src/imgui_ex/vcImGuiSimpleWidgets.cpp
OsiJr: Sure isn't. Nice catch.

samsinsane (approved):
String is already in master untranslated (comes from the markdown submodule), merging for now for testing, translation will come later.

PR #1137, opened 2020-09-24
fix on camera map movements

jproust:
Fixes some issue with the map mode while moving the camera on the other viewport (orbiting & move to)
Fixes [AB#2268](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/2268)

### DIFF
--- a/src/vcCamera.cpp
+++ b/src/vcCamera.cpp
@@ -89,6 +89,7 @@ void vcCamera_UpdateSmoothing(vcState * pProgramState, vcCamera *pCamera, vcCame
       udDouble3 step = pCamInput->smoothTranslation * udMin(1.0, stepAmount * sCameraTranslationSmoothingSpeed);
       pCamera->position += step;
       pCamInput->smoothTranslation -= step;
+
       vcCamera_UpdateCameraMap(pProgramState, pCamera);
     }
     else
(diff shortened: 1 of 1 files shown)


src/vcCamera.cpp
```diff
@@ -45,46 +89,8 @@ void vcCamera_UpdateSmoothing(vcState * pProgramState, vcCamera *pCamera, vcCame
...
-              udDouble3 camMapPosLatLong = udGeoZone_CartesianToLatLong(pProgramState->geozone, pProgramState->pViewports[v].camera.position);
-              camPosLatLong.z = camMapPosLatLong.z;
-              pProgramState->pViewports[v].camera.position = udGeoZone_LatLongToCartesian(pProgramState->geozone, camPosLatLong);
-            }
-            pProgramState->pViewports[v].camera.headingPitch = udDouble2::create(0.0f, -UD_HALF_PI);
-          }
-        }
-      }
+
+      vcCamera_UpdateCameraMap(pProgramState, pCamera);
```
OsiJr, line 93: It seems like all of the changes in this file could have been resolved just be changing the if above so this block gets called whenever there is a mapMode regardless of `if (udMagSq3(pCamInput->smoothTranslation) > minSmoothingThreshold)`?

jproust: Fixed, now just call vcCamera_UpdateCameraMap for every movement of the camera.

OsiJr (approved):

PR #1143, opened 2020-09-25
Flythrough Export Modal

Jossos:
- Implemented Modal to disable program while flythrough export is occurring
- Fixes [AB#2203](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/2203)


src/main.cpp
```diff
@@ -1046,6 +1046,8 @@ int main(int argc, char **args)
   programState.activeViewportIndex = 0;
   programState.pActiveViewport = &programState.pViewports[programState.activeViewportIndex];
 
+  programState.flythroughExportCancel = false;
```
samsinsane, line 1049: This should already be the case?

builds/assets/lang/enAU.json
```diff
@@ -329,7 +329,8 @@
   "flythroughFPS": "FPS",
   "flythroughDefaultName": "Camera Path",
   "flythroughExport": "Export",
-  "flythroughExportStop": "Stop",
+  "flythroughExporting": "Exporting Flythrough...",
+  "flythroughExportCancel": "Cancel",
```
OsiJr, line 333: Instead of adding new keys (which will trigger requiring new translations at the last second) use these:
  - `sceneExplorerCancelButton` and;
  - `sceneExplorerExportRunning`

src/vcModals.cpp
```diff
@@ -1815,6 +1815,30 @@ void vcModals_DrawImportShapeFile(vcState *pProgramState)
...
+  {
+    if (pProgramState->closeModals & (1 << vcMT_FlythroughExport))
+      ImGui::CloseCurrentPopup();
+    else
+      pProgramState->modalOpen = true;
+
+    ImGui::TextUnformatted(vcString::Get("flythroughExporting"));
+
+    if (ImGui::Button(vcString::Get("flythroughExportCancel"), ImVec2(200, 30)))
+      pProgramState->flythroughExportCancel = true;
```
samsinsane, line 1836: Alternatively, you could add a function to the `vcFlythrough` scene item, and call that instead? The selected item is always going to be a flythrough since selecting another scene item will cancel the flythrough.
```c++
vcFlythrough *pFlythrough = (vcFlythrough*)pProgramState->sceneExplorer.clickedItem.pItem->pUserData;
```

samsinsane (changes requested):

src/vcModals.cpp
Jossos: i like avoiding clogging up state, but there only ever needs to be 1 instance of this variable. doesn't make sense for each flythrough to have their own cancel variable.

src/main.cpp
Jossos: no this is the initialization part of main, so it hasn't been assigned yet

samsinsane: As discussed on teams, the `programState` is initialised with `{}` so this is already `false`.

src/vcModals.cpp
samsinsane: As discussed on teams, I meant a cancel function rather than a new variable, this would remove the additional state created here in favour of just performing the cancel action.

samsinsane (approved):

PR #1162, opened 2020-10-01
Added domain licensing support

OsiJr:


src/main.cpp
```diff
@@ -1146,7 +1157,10 @@ int main(int argc, char **args)
 
   vcProject_CreateBlankScene(&programState, "Empty Project", vcPSZ_StandardGeoJSON);
 
-  udWorkerPool_AddTask(programState.pWorkerPool, vcMain_AsyncResumeSession, &programState, false);
+  if (!tryDomain)
```
samsinsane, line 1160: Given that this has an else, this should just be `if (tryDomain)`.

samsinsane (approved):

PR #1163, opened 2020-10-02
Gizmo Camera bug Fix

Jossos:
- Fixed a bug where deleting a model while hovering the gizmo disabled the camera
- Fixes [AB#2309](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/2309)


samsinsane (changes requested):
This is just a bandaid - the gizmo needs to be deactivated.

PR #1164, opened 2020-10-02
Added libtiff submodule and a few tests

int-Frank:


vcTesting/src/vcTiffTests.cpp
```diff
@@ -0,0 +1,88 @@
+#include "vcTesting.h"
+#include "tiffio.h"
+#include "udPlatformUtil.h"
+#include "udStringUtil.h"
+
+#define EXPECT_EQ_BREAK(a, b) EXPECT_EQ(a, b); if (a != b) break
+#define EXPECT_NE_BREAK(a, b) EXPECT_NE(a, b); if (a == b) break
```
samsinsane, line 7: You can just use `ASSERT_*` instead if you want the test to stop.

samsinsane (approved):

PR #1170, opened 2020-10-02
Added tiff memory streamer

int-Frank:


vcTesting/src/vcTiffTests.cpp
```diff
@@ -47,38 +46,149 @@ static TIFF *vcTesting_LoadTiffTestFile(char const *pFileName)
...
+    EXPECT_EQ(directoryCount, 1);
+    if (directoryCount != 1)
+      break;
 
     // Colour data
-    EXPECT_EQ_BREAK(TIFFGetField(pTif, TIFFTAG_IMAGEWIDTH, &w), 1);
-    EXPECT_EQ_BREAK(w, 3ul);
-    EXPECT_EQ_BREAK(TIFFGetField(pTif, TIFFTAG_IMAGELENGTH, &h), 1);
-    EXPECT_EQ_BREAK(h, 3ul);
+    intRes = TIFFGetField(pTif, TIFFTAG_IMAGEWIDTH, &w);
```
int-Frank, line 100: Maybe these can be Macrod

vcTesting/src/vcTiffTests.cpp
```diff
@@ -47,38 +46,149 @@ static TIFF *vcTesting_LoadTiffTestFile(char const *pFileName)
...
+    EXPECT_EQ(memcmp(gTEST_FILE_1_PIXELS, raster, npixels * sizeof(uint32)), 0);
+  } while (false);
+
+  _TIFFfree(raster);
+  TIFFClose(pTif);
+  vcTiffHandler_DestroyMemoryStream(&pStream);
+  udFree(pData);
+}
+
+TEST(vcTiff, LoadTiffFile)
```
int-Frank, line 139: Turns out we don't actually use libtiff to load tiff from file. But this test exists nonetheless.

PR #1174, opened 2020-10-06
Added tiff support throughout udStream

int-Frank:
Fixes [AB#2281](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/2281)

### DIFF
--- a/src/scene/vcMedia.cpp
+++ b/src/scene/vcMedia.cpp
@@ -148,9 +148,9 @@ void vcMedia::AddToScene(vcState *pProgramState, vcRenderData *pRenderData)
     udInt2 imageSize = {};
     if (m_pImageData != nullptr)
     {
-      uint32_t width, height, channelCount;
+      vcImageData tempData = {};
+      uint8_t *pData = vcTexture_DecodePixels((uint8_t *)m_pImageData, m_imageDataSize, &tempData);
 
-      uint8_t *pData = stbi_load_from_memory((stbi_uc *)m_pImageData, (int)m_imageDataSize, (int *)& width, (int *)& height, (int *)& channelCount, 4);
       udFree(m_pImageData);
       m_imageDataSize = 0;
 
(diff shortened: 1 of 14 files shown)


vcGL/src/vcTextureHelper.cpp
```diff
@@ -414,3 +415,62 @@ udResult vcTexture_SaveImage(vcTexture *pTexture, vcFramebuffer *pFramebuffer, c
...
+    if (pPixelData != nullptr)
+      break;
+
+    pPixelData = vcTexture_DecodePixels_tiff(pFileContents, fileLength, pImageData);
+    if (pPixelData != nullptr)
+      break;
+
+  } while (false);
+
+  return pPixelData;
```
OsiJr, line 475:
```suggestion
  uint8_t *pPixelData = nullptr;

  // try stbi
  if (pPixelData == nullptr)
    pPixelData = vcTexture_DecodePixels_stbi(pFileContents, fileLength, pImageData);

  // if nothing loaded, try libtiff
  if (pPixelData == nullptr)
    pPixelData = vcTexture_DecodePixels_tiff(pFileContents, fileLength, pImageData);
    
  return pPixelData;
```

Is a bit cleaner and

vcGL/src/vcTiff.cpp
```diff
@@ -0,0 +1,189 @@
...
+  vcMemoryInStream *pStream = nullptr;
+  TIFF *pTif = nullptr;
+  size_t npixels = 0;
+  uint32 w, h;
+  uint32 *pRaster = nullptr;
+
+  UD_ERROR_NULL(pBuffer, udR_InvalidParameter_);
+  UD_ERROR_NULL(pW, udR_InvalidParameter_);
+  UD_ERROR_NULL(pH, udR_InvalidParameter_);
+  UD_ERROR_NULL(ppPixelData, udR_InvalidParameter_);
```
OsiJr, line 161: The convention for `vc*` is to just have this as an if check above everything else.

src/scene/vcMedia.cpp
```diff
@@ -148,9 +148,9 @@ void vcMedia::AddToScene(vcState *pProgramState, vcRenderData *pRenderData)
     udInt2 imageSize = {};
     if (m_pImageData != nullptr)
```
OsiJr, line 149: In general, this file looks like it should have use the texture cache instead (outside the scope of this PR)

int-Frank: Ok. Is adapting `vcMedia` to use the texture cache within the scope of this PR? As I write this I would imagine it should probably be it's own PR.

int-Frank: Should I roll this back?

OsiJr (approved):

PR #1177, opened 2020-10-09
Fhart/tiff convert support

int-Frank:
This is the first of a few PR's for Tiff conversion, task [AB#2334](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/2334)

Supports converting image tiff as flat images. Have yet to optimise for tile/strip/scanline tiffs, compressed (haven't tested), geotiff, 3d images and image data broken up on multiple planes.

### DIFF
--- a/project.lua
+++ b/project.lua
@@ -26,6 +26,7 @@ project "udStream"
 	includedirs { "src", "src/scene", "src/rendering" }
 	includedirs { "3rdParty/udcore/Include" }
 	includedirs { "3rdParty/imgui" }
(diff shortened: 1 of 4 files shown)


OsiJr:
As mentioned in DM's; I have reservations about using virtual functions in the convert pipe (particularly per point). Will merge as this is only partial support anyway and it can continue to evolve.

OsiJr (approved):

PR #1179, opened 2020-10-09
Out-of-Geozone-bounds position error

Jossos:
- Added an error for when the user sets the position too far away
- Fixes [AB#2338](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/2338)

### DIFF
--- a/src/vcProject.cpp
+++ b/src/vcProject.cpp
@@ -10,6 +10,8 @@
 #include "udStringUtil.h"
 #include "vcError.h"
 
+#include <cmath>
+
 const char *vcProject_ErrorToString(udError error)
 {
   switch (error)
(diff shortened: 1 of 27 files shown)


src/vcProject.cpp
```diff
@@ -788,24 +790,54 @@ bool vcProject_UpdateNodeGeometryFromCartesian(vcProject *pProject, udProjectNod
...
+    {
+      if (!std::isfinite(((double*)pGeom)[i]))
+      {
+        result = udE_InvalidParameter;
+        break;
+      }
+    }
+
+    if (result != udE_InvalidParameter)
+      result = udProjectNode_SetGeometry(pProject->pProject, pNode, newType, numPoints, (double*)pGeom);
```
OsiJr, line 803: This will prevent the save happening (so bad data isn't written out) but bad data already in the file will still reproduce the crash. Not calling `udProjectNode_SetGeometry` will also not trigger `OnNodeUpdate` so the state of the node might be broken.

OsiJr (changes requested):
I don't have an immediate (complete) solution to the problem but this isn't a complete solution yet either.

Jossos:
> I don't have an immediate (complete) solution to the problem but this isn't a complete solution yet either.

ok well the issue specifically is with position.x being too large for the geozone, and cosh and sinh return exponentially huge values depending on zone.radius i think. This is all in udGeoZone_CartesianToLatLong

samsinsane (changes requested):
Not a huge fan of this change. `udGeoZone` really should be doing the `isfinite` checks, if it spits out garbage and returns `udR_Success`, that's a problem. We can also make this optional for the cases where validation is unnecessary.

PR #1183, opened 2020-10-13
Fixed crash on Login error

jproust:
popStyleColor() crashes because there is no pushStyleColor() before so the .back on the vector crashes.
Fixes [AB#2354](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/2354)

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2991,10 +2991,7 @@ void vcMain_ShowLoginWindow(vcState *pProgramState)
     {
       ImGui::TextUnformatted(vcString::Get(loginStatusKeys[pProgramState->loginStatus]));
 
-      const bool isErrorStatus = udStrBeginsWith(loginStatusKeys[pProgramState->loginStatus], "loginError");
-
-      if (isErrorStatus)
-        ImGui::PopStyleColor();
+      udStrBeginsWith(loginStatusKeys[pProgramState->loginStatus], "loginError");
 
       // Tool for support to get reasons for failures, requires Alt & Ctrl
       if (pProgramState->logoutReason != udE_Success && io.KeyAlt && io.KeyCtrl)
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -2991,10 +2991,7 @@ void vcMain_ShowLoginWindow(vcState *pProgramState)
     {
       ImGui::TextUnformatted(vcString::Get(loginStatusKeys[pProgramState->loginStatus]));
 
-      const bool isErrorStatus = udStrBeginsWith(loginStatusKeys[pProgramState->loginStatus], "loginError");
-
-      if (isErrorStatus)
-        ImGui::PopStyleColor();
+      udStrBeginsWith(loginStatusKeys[pProgramState->loginStatus], "loginError");
```
samsinsane, line 2994: This doesn't do anything, just remove it?

samsinsane (changes requested):

src/main.cpp
jproust: True, I just saw that the error was unnecessary, I did not check the function.

src/main.cpp
```diff
@@ -2991,10 +2991,7 @@ void vcMain_ShowLoginWindow(vcState *pProgramState)
     {
       ImGui::TextUnformatted(vcString::Get(loginStatusKeys[pProgramState->loginStatus]));
```
Jossos, line 2992: Sorry my bad, during a rebase I accidentally removed some code.

It should be this:

      const bool isErrorStatus = udStrBeginsWith(loginStatusKeys[pProgramState->loginStatus], "loginError");

      if (isErrorStatus)
        ImGui::PushStyleColor(ImGuiCol_Text, ImVec4(1.0, 0.0, 0.0, 1.0));

      ImGui::TextUnformatted(vcString::Get(loginStatusKeys[pProgramState->loginStatus]));

      if (isErrorStatus)
        ImGui::PopStyleColor();

src/main.cpp
jproust: Removed

src/main.cpp
jproust: Fixed

Jossos (approved):

samsinsane (approved):

PR #1184, opened 2020-10-13
Area Measurement Mesh for Side Shapes

Jossos:
- Generate meshes correctly on sides of buildings
- Project for Delaney algorithm from first 3 points normal
- Fixes [AB#2335](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/2335)

### DIFF
--- a/src/scene/vcPOI.cpp
+++ b/src/scene/vcPOI.cpp
@@ -1403,17 +1403,6 @@ static bool vcPOI_LineLineIntersection2D(const udDouble3 &line1Point1, const udD
   return true;
 }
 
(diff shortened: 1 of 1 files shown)


Jossos:
Peter asked me to try get rid of the * quat::create(half_pi), so this isn't ready to merge until I've either fixed that, or discovered it's actually correct

Jossos:
> Peter asked me to try get rid of the * quat::create(half_pi), so this isn't ready to merge until I've either fixed that, or discovered it's actually correct

So I checked this, and it looks like the quaternion generated from the plane normal tells us how to get from {1, 0, 0} to the plane normal, but we project to {0, 0, 1}. Adding or subtracting half_pi to the pitch fixes this.

peterAdams1014 (approved):

PR #1189, opened 2020-10-14
Speed up perceived start time

Jossos:
- Speed up perceived start time
- Fixes [AB#2324](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/2324)

Most changes in main.cpp is just pre-existing code moving around


samsinsane (changes requested):
Having run through the start-up using this PR a couple of times and compared to the current start-up. This is a huge step backwards in user experience, and while UI items may display on screen sooner than before, you can't interact with them due to the constant stalls while loading resources synchronously on the main thread.

PR #1195, opened 2020-10-15
POI Rotate Gizmo Pivot

Jossos:
- Added gizmo pivot so that gizmo stays in same position when rotating
- Fixes [AB#2348](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/2348)

### DIFF
--- a/src/scene/vcPOI.cpp
+++ b/src/scene/vcPOI.cpp
@@ -493,6 +493,7 @@ vcPOI::vcPOI(vcProject *pProject, udProjectNode *pNode, vcState *pProgramState) 
   m_showArea = false;
   m_showLength = false;
   m_showAngles = false;
+  m_gizmoPivot = udDouble3::zero();
   m_labels.Init(32);
 
   memset(&m_line, 0, sizeof(m_line));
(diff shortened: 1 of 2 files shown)


OsiJr (changes requested):
This fix isn't correct. The real issue is that the centroid isn't stable under different rotations. Setting it to average (not ideal) or factoring the rotation for the AABB calculations are better solutions.

Jossos:
> This fix isn't correct. The real issue is that the centroid isn't stable under different rotations. Setting it to average (not ideal) or factoring the rotation for the AABB calculations are better solutions.

using average gives pretty bad results. using AABB isn't ideal, but it's pretty good for most situations, just not rotation gizmo.

When you say "factoring the rotation for AABB calculations" , im not sure what you mean. I did do something where i moved the entire shape when rotating so that the centroid is always at the same position as the rotation gizmo, and it looks really bad.

OsiJr:
> When you say "factoring the rotation for AABB calculations" , im not sure what you mean. I did do something where i moved the entire shape when rotating so that the centroid is always at the same position as the rotation gizmo, and it looks really bad.

I meant that it should either;
  1. Be the centre of a tight convex hull (rather than the min/max value on each axis) or;
  2. Generate the AABB where the first 2 points (with either up or north) define the axis.

Under either of those options the centroid will remain stable during rotation.

PR #1196, opened 2020-10-15
Converting tiff now supports a lot more of the tif spec

int-Frank:
Fixes [AB#2334](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/2334) 

Supports a lot more of the tiff spec, but does not yet support geotiff. Currently supports:
 - pulling all images from all internal directories
 - general image loading
 - optimised loading tiled image data
 - 1, 2, 4, 8, 16, 32 and 64-bit uint samples (pixel channels)
 - 2, 4, 8, 16, 32 and 64-bit int samples
 - 32 and 64-bit float samples
 - up to 4 samples per pixel
 - Grey scale images 
 - RGB images
 - Palettized images
 - Samples stored contiguous or separately

### DIFF
--- a/src/vcTiffConvert.cpp
+++ b/src/vcTiffConvert.cpp
@@ -473,7 +473,7 @@ static udError vcTiffImageReaderReadPixel_Tile(vcTiffConvertData *pConvertData, 
   pixelCoords.x = double(pConvertData->tile.currentPixel % pConvertData->tile.tileDimensions.x) * pConvertData->pointResolution;
   pixelCoords.y = double(pConvertData->tile.currentPixel / pConvertData->tile.tileDimensions.x) * pConvertData->pointResolution;
   pixelCoords.z = 0.0;
-  tileOrigin = {(double)pConvertData->tile.currentTileCoords.x * pConvertData->tile.tileDimensions.x, (double)pConvertData->tile.currentTileCoords.y * pConvertData->tile.tileDimensions.y};
+  tileOrigin = {(double)pConvertData->tile.currentTileCoords.x * pConvertData->tile.tileDimensions.x, (double)pConvertData->tile.currentTileCoords.y * pConvertData->tile.tileDimensions.y, 0.0};
   tileOrigin *= pConvertData->pointResolution;
 
   pPixelData->point = pConvertData->origin + tileOrigin + pixelCoords;
@@ -719,7 +719,7 @@ udError TiffConvert_ReadFloat(struct udConvertCustomItem *pConvertInput, struct 
   udError result;
   udError pixReadRes;
   vcTiffPixelData pixelData = {};
-  uint32_t colour;
+  uint32_t colour = 0xFFFF00FF;
   vcTiffConvertData *pData = nullptr;
 
   if (pConvertInput == nullptr || pBuffer == nullptr)
--- a/src/vcTiffConvert.cpp
+++ b/src/vcTiffConvert.cpp
@@ -53,6 +53,22 @@ struct vcTiffConvertData;
 typedef void(*vcTiffImageReaderDestroyFn)(vcTiffConvertData *pData);
 typedef udError(*vcTiffImageReaderReadPixelFn)(vcTiffConvertData *pData, vcTiffPixelData *pPixelData);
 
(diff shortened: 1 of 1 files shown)


src/vcTiffConvert.cpp
```diff
@@ -10,157 +10,592 @@
...
+    if (format.sampleFormat == SAMPLEFORMAT_INT)
+      pSample->t_int64[sampleIndex] = ((int8_t *)(pBuf))[ind];
+    else if (format.sampleFormat == SAMPLEFORMAT_UINT)
+      pSample->t_uint64[sampleIndex] = ((uint8_t *)(pBuf))[ind];
+    else
+      UD_ERROR_SET(udE_InvalidConfiguration);
+  }
+  else if (format.bitsPerSample == 16)
+  {
+    // Currently do not support 16-bit floats
```
samsinsane, line 331: Just an FYI, I believe there's a function that can pack and/or unpack 16bit floats in udStream already. This is used for the 16bit float on the graphics card - I'm confident it won't be immediately usable for you though, so not a required change for this PR.

src/vcTiffConvert.cpp
```diff
@@ -10,157 +10,592 @@
...
+  else
   {
+    if (pData->tile.pRaster == nullptr)
+    {
+      pData->tile.pRaster = (uint8_t *)_TIFFmalloc(TIFFTileSize(pData->pTif) * udMin(pData->format.samplesPerPixel, MAX_SAMPLES));
+      UD_ERROR_NULL(pData->tile.pRaster, udE_MemoryAllocationFailure);
+    }
+
+    size_t pixelCount = (size_t)pData->tile.tileDimensions.x * pData->tile.tileDimensions.y;
+    pBuf = (uint8 *)_TIFFmalloc(TIFFTileSize(pData->pTif)); // TODO could allocated this once per image
```
samsinsane, line 402: Given that this function frees `pBuf`, is there any reason to use `_TIFFmalloc` over `udAllocType` here?

src/vcTiffConvert.cpp
```diff
@@ -10,157 +10,592 @@
...
+epilogue:
+  return result;
+}
+
+static void vcTiffImageReaderDestroy_Tile(vcTiffConvertData *pData)
+{
+  if (pData == nullptr)
+    return;
+
+  _TIFFfree(pData->tile.pRaster);
```
samsinsane, line 444: Is there a case where this isn't allocated by your code above? Seems like using `udAllocType` would be superior in this case.

src/vcTiffConvert.cpp
```diff
@@ -10,157 +10,592 @@
...
+struct ScanLineData
+{
+  uint8_t *pRaster;
+  uint64_t currentPixel;
+  uint32 row;
+};
+
+struct vcTiffConvertData
+{
+  TIFF *pTif;
```
samsinsane, line 74: Doesn't make a huge amount of sense to rename this from `pTiff`, it results in a number of lines changing and is kind of less clear. Is a TIF different to a TIFF?

src/vcTiffConvert.cpp
```diff
@@ -10,157 +10,592 @@
...
+
+  if (nBits < 8)
+    return uint32_t(255ll * value / ((1ll << nBits) - 1));
+  if (nBits > 8)
+    return uint32_t(value >> (nBits - 8));
+  return (uint32_t)value;
+}
 
-  void SetConvertDefaults(udDouble3 const & origin, double pointResolution, enum udConvertCustomItemFlags flags)
+static udError vcTiff_SampleDataToColour(vcTiffImageFormat const & format, vcTiffSampleData const & data, uint32_t *pColour)
```
samsinsane, line 141: Both `*` and `&` go with the variable name, this happens in a few other functions. You may need to adjust your Visual Studio settings if this is automatically happening.

samsinsane (changes requested):
Reconsider all usages of the libtiff allocation system - it's only necessary to use if libtiff is going to provide or free the pointers at some point.

OsiJr (approved):

PR #1198, opened 2020-10-16
Optimization for SHP file loading

asunawu:
To read 100M content or a record larger than 100M at a time, instead of reading records one by one. This change is for loading files from URL.
And the data structure also has been reorganized to make the code more concise.
[AB#2270](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/2270).

### DIFF
--- a/src/vcSHP.cpp
+++ b/src/vcSHP.cpp
@@ -6,6 +6,11 @@
 #include "vcDBF.h"
 #include "udStringUtil.h"
 
+enum vcSHPConstant
+{
+  vcSHP_MemoryCache = 1024 * 1000 * 100, // 100M
+};
+
 enum vcSHPType
 {
   vcSHPType_Empty = 0,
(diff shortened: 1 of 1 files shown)


samsinsane (changes requested):
There's too many changes here which makes this almost impossible to read. You need to break this up into the two changes you made, each with their own PR:
- Reading 100MB blocks at a time
- Reorganising and consolidating data structures and code

asunawu:
The first change requires to load data from memory instead of udFile_Read. This change itself will almost affect the same amount of code. If I don’t simplify the data structure, not only the review, but the code is very hard to read. Fortunately, I’ve fully tested it. With these changes, the entire module is working normally.

asunawu:
Sorry for so many code changed, however this is the most efficient method I thought of.

src/vcSHP.cpp
```diff
@@ -258,318 +129,181 @@ void vcSHP_Release(vcSHP *pSHP)
...
+      memcpy(&record.MBRMax.y, *ppReadPosition, sizeof(double));
+      *ppReadPosition += sizeof(double);
     }
 
-    // z
-    readPosition = UnpackDouble(readPosition, &record.data.polyZ.ZRange.x, buffer);
-    readPosition = UnpackDouble(readPosition, &record.data.polyZ.ZRange.y, buffer);
-    for (int i = 0; i < record.data.polyZ.pointsCount; i++)
+    // read parts count
+    if (record.shapeType == vcSHPType_Polyline || record.shapeType == vcSHPType_Polygon || record.shapeType == vcSHPType_PolylineZ || record.shapeType == vcSHPType_PolygonZ || record.shapeType == vcSHPType_PolylineM || record.shapeType == vcSHPType_PolygonM || record.shapeType == vcSHPType_MultiPatch)
```
OsiJr, line 215: What doesn't this apply to?

OsiJr (approved):

samsinsane (approved):

PR #1200, opened 2020-10-16
Filter Shape Name Change

Jossos:
- Filters change names to the appropriate default shape name if their shape changes and their name is still default
- Fixes [AB#2386](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/2386)

### DIFF
--- a/src/scene/vcQueryNode.cpp
+++ b/src/scene/vcQueryNode.cpp
@@ -149,20 +149,6 @@ void vcQueryNode::HandleSceneExplorerUI(vcState *pProgramState, size_t *pItemID)
 {
   bool changed = false;
 
-  const char *filterShapeNames[] = { vcString::Get("sceneFilterShapeBox"), vcString::Get("sceneFilterShapeCylinder"), vcString::Get("sceneFilterShapeSphere") };
-  int shape = m_shape;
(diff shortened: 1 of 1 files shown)


src/scene/vcQueryNode.cpp
```diff
@@ -153,6 +153,11 @@ void vcQueryNode::HandleSceneExplorerUI(vcState *pProgramState, size_t *pItemID)
   int shape = m_shape;
   if (ImGui::Combo(udTempStr("%s##FilterShape%zu", vcString::Get("sceneFilterShape"), *pItemID), &shape, filterShapeNames, (int)udLengthOf(filterShapeNames)))
   {
+    // Change name if is still default
+    const char *defaultShapeNames[] = { vcString::Get("sceneExplorerFilterBoxDefaultName"), vcString::Get("sceneExplorerFilterCylinderDefaultName"), vcString::Get("sceneExplorerFilterSphereDefaultName") };
```
samsinsane, line 157: You need to assert the size of this is equal to the number of possible shapes. A new shape being added won't trigger this code to be magically updated, and will eventually lead to a crash when someone inevitably changes to or from the new shape type.

samsinsane (changes requested):
Just remove the ability to change types, in the long run this won't make any sense. Right now, you can't change to the cross section filter, but in the future as we add support for unions this likely won't be possible to change anyway. It was a feature that never really made any sense to begin with that should have been removed long ago.

samsinsane (approved):

PR #1212, opened 2020-10-22
[User Guide] Added user guide to the project

OsiJr:


src/vcModals.cpp
```diff
@@ -2174,6 +2174,51 @@ void vcModals_DrawImportAnnotations(vcState *pProgramState)
...
+  if (ImGui::BeginPopupModal("###modalUserGuide", nullptr, ImGuiWindowFlags_NoTitleBar))
+  {
+    if (pProgramState->closeModals & (1 << vcMT_UserGuide))
+      ImGui::CloseCurrentPopup();
+    else
+      pProgramState->modalOpen = true;
+
+    static const char *pGuideData = nullptr;
+
+    if (ImGui::Button("Reload") || pGuideData == nullptr)
```
samsinsane, line 2194: This needs to be translated.

src/vcModals.cpp
```diff
@@ -2174,6 +2174,51 @@ void vcModals_DrawImportAnnotations(vcState *pProgramState)
...
+    {
+      udFree(pGuideData);
+      udFile_Load("asset://assets/guide/userguide.md", &pGuideData);
+    }
+
+    ImGui::SameLine();
+
+    if (ImGui::Button("Close"))
```
samsinsane, line 2202: This needs to be translated.

samsinsane (changes requested):

src/vcModals.cpp
OsiJr: This needs to be wrapped in GitBuild but same same

samsinsane (approved):

PR #1214, opened 2020-10-22
Added ability to remove an attached model

int-Frank:
Fixes [AB#2412](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/2412)


OsiJr (approved):

OsiJr (changes requested):
This will detach it but it will come back after the project is loaded again

PR #1215, opened 2020-10-22
Added limit to how many bytes are read when loading CSVs

int-Frank:
Fixes [AB#2414](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/2414)

### DIFF
--- a/src/vcModals.cpp
+++ b/src/vcModals.cpp
@@ -1865,6 +1865,7 @@ void vcModals_DrawImportAnnotations(vcState *pProgramState)
 {
   static vcCSVImportSettings importSettings = {};
   static udGeoZone importZone = {};
+  const uint64_t maxFileReadBytes = 1024 * 64;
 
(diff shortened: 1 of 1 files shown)


src/vcModals.cpp
```diff
@@ -1978,7 +1979,7 @@ void vcModals_DrawImportAnnotations(vcState *pProgramState)
         if (ImGui::Button(vcString::Get("annotationsAddToScene"), ImVec2(100.f, 0)))
         {
           // read remaining bytes
-          vcCSV_Read(pProgramState->pImportAnnotationsContext);
+          vcCSV_Read(pProgramState->pImportAnnotationsContext, maxFileReadBytes);
```
samsinsane, line 1982: How is the rest of the data in the file read?

samsinsane (changes requested):

PR #1217, opened 2020-10-26
remove existing keyframes

jproust:
Fixes [AB#2440](https://dev.azure.com/euclideon/57c3bd2a-8f94-4578-b195-2e4fa9d8f295/_workitems/edit/2440)


OsiJr (changes requested):
The existing functionality is also there on request. Its to "correct" poor pathing for a small section. Theres 3 different ways people seem to want this to work.

1. Override the existing timeline
2. Extend the existing timeline so that existing content is after the current recording
3. Delete everything after the current recording.

Should this be a setting or a drop down on the "record" or something when the timeline its not at the end of the recording?

@CarlCFX?

CarlCFX:
@OsiJr When I spoke to Adam about it, I was of the mindset that, if you've scrubbed through the time to a particular point and start recording again, it should remove all the previous nodes after that point and record new nodes/frames from there.

